PR #184, opened 2021-09-25
Fix past repeat epoch and cleanup user admin

exrhizo:
Require: https://github.com/coordinape/coordinape-backend/pull/46

After seeing Moonshot use the admin user epoch, I reordered the fields and switched the polarity of non_giver.

Also, `const { deleteUser, deleteEpoch } = useAdminApi()` had been kept in the action renderer and would be called with stale values.

### DIFF
--- a/src/hooks/useAdminApi.ts
+++ b/src/hooks/useAdminApi.ts
@@ -1,4 +1,8 @@
-import { useRecoilValue, useSetRecoilState } from 'recoil';
+import {
+  useRecoilCallback,
+  useSetRecoilState,
+  CallbackInterface,
+} from 'recoil';
 
 import {
   rSelectedCircleId,
@@ -30,12 +34,21 @@ export const useAdminApi = () => {
   const updateEpochsMap = useSetRecoilState(rEpochsRaw);
   const updateUsersMap = useSetRecoilState(rUsersMapRaw);
 
-  // A fake circleId will just return nothing
-  const selectedCircleId = useRecoilValue(rSelectedCircleId) ?? -1;
-  const myAddress = useRecoilValue(rMyAddress);
+  const getState = useRecoilCallback(
+    ({ snapshot }: CallbackInterface) => async () => {
+      const selectedCircleId =
+        (await snapshot.getPromise(rSelectedCircleId)) ?? -1;
+      const myAddress = await snapshot.getPromise(rMyAddress);
+      return {
+        selectedCircleId,
+        myAddress,
+      };
+    }
+  );
 
   const updateCircle = (params: PutCirclesParam) =>
     callWithLoadCatch(async () => {
+      const { myAddress, selectedCircleId } = await getState();
       if (myAddress === undefined) throw 'myAddress required';
       const newCircle = await api.putCircles(
         selectedCircleId,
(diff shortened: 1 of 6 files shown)


src/pages/AdminPage/AdminPage.tsx
```diff
@@ -268,7 +262,7 @@ const AdminPage = () => {
         },
         {
           label: 'Are they admin?',
-          render: (u: IUser) => (u.role === 0 ? 'No' : 'Yes'),
```
levity, line 271: maybe also add a column for Opt Out status, not just Force Opt Out?

src/hooks/useAdminApi.ts
```diff
@@ -30,12 +34,21 @@ export const useAdminApi = () => {
   const updateEpochsMap = useSetRecoilState(rEpochsRaw);
   const updateUsersMap = useSetRecoilState(rUsersMapRaw);
 
-  // A fake circleId will just return nothing
-  const selectedCircleId = useRecoilValue(rSelectedCircleId) ?? -1;
-  const myAddress = useRecoilValue(rMyAddress);
+  const getState = useRecoilCallback(
```
levity, line 37: interesting. so `useAdminApi()` for some reason isn't refreshing these two values when you change your address or the circle id? that seems weird... i wonder if there's a fix that gets closer to the root issue than this?

src/pages/AdminPage/AdminUserModal.tsx
```diff
@@ -76,14 +88,24 @@ export const AdminUserModal = ({
...
               type="number"
               label="Starting Tokens"
             />
+            <ApeToggle
+              {...non_giver}
+              onChange={(v) => nonGiverOnChange(!v)}
+              value={!nonGiverValue}
+              label="Can Give?"
+            />
+            <ApeToggle {...fields.fixed_non_receiver} label="Force Opt'd Out" />
```
levity, line 103:
```suggestion
            <ApeToggle {...fields.fixed_non_receiver} label="Force Opted Out" />
```

src/pages/AdminPage/AdminUserModal.tsx
```diff
@@ -76,14 +88,24 @@ export const AdminUserModal = ({
...
+            <ApeToggle
+              {...fields.non_receiver}
+              disabled={fields.fixed_non_receiver.value}
+              label="Opt'd Out"
```
levity, line 107:
```suggestion
              label="Opted Out"
```

levity (changes requested):

src/hooks/useAdminApi.ts
levity: not a blocker, just thinking out loud

exrhizo: The problem was actually that it was being set on the column's useMemo and those didn't update. The underlying dependency isn't necessarily available either. I think it is best for these functions be be stateless, but perhaps what I am doing here is still odd...

src/pages/AdminPage/AdminPage.tsx
exrhizo: This inspired me to go on a whole rewording spree :0

exrhizo:
@levity You put "changes requested" but none of your comments indicated what was needed

levity (approved):

PR #186, opened 2021-09-27
Initial Circle Creation Page

exrhizo:
https://linear.app/yearn/issue/APE-300/create-info-help-component
https://linear.app/yearn/issue/APE-294/add-captcha-to-circle-creation-form
https://linear.app/yearn/issue/APE-284/page-with-circle-creation-form

### TODO: Testing with server
 - Does the app wait to navigate until after the profile is fetched?
 - [Backend PR](/coordinape/coordinape-backend#47)

### CreateCirclePage
 - Add to `routes`
 - Update `LandingPage`, `DefaultPage` and `AdminPage` to have buttons to this

### New Components
 - ApeAutocomplete - Encapsulated what was in the map drawer
 - ApeInfoTooltip
 - DialogNotice
   - Refactored from PopUpModal
   - `PopUpModal` was no longer needed, legacy from before tabbed give, same message in `AllocateGive`
 - InfoIcon
 - FormCaptcha
 - FormAutocomplete simple wrapper around ApeAutocomplete

### Updates
 - ApeTextField
   - With info tooltip prop
   - Fix height so error doesn't reflow
 - StaticTable
    - Now with placeholders when empty

### Fixes 
 - `RecoilAppController` was fetching circles before wallet connect
   - Actually there is still a bug to do with disconnecting from one wallet to another, using old membership
   - This could use a refactoring
   - useAdminApi should have been using new ids when updating circles

### Cleanup
 - Remove unused variables 

 ## Future Work
 - Only error fields after they have been touched or submit clicked
 - Implement


src/forms/CreateCircleForm.tsx
```diff
@@ -0,0 +1,50 @@
...
+  .object({
+    user_name: z
+      .string()
+      .min(3, 'User name must be at least 3 characters long.'),
+    circle_name: z
+      .string()
+      .min(3, 'Circle name must be at least 3 characters long.'),
+    protocol_name: z
+      .string()
+      .min(3, 'User name must be at least 3 characters long.'),
```
crabsinger, line 15: spelling: "Protocol" not user here.

crabsinger (commented):
I looked over this - I'm not super familiar to give a meaningful review of this code - but no obvious things stands out beyond a typo.

I tested this on vercel staging and it seems like after clikcing "Launch this circle" it doesn't redirect anywhere - I see your TODO on this `Add the success modal` so maybe that also redirects to the Add folks to circle, or whatever the next part of the flow should be.

Another question: it seems the `Organization Name` field is required, but for a new user this is blank, but it does seem to create a circle even with a text input organization that doesn't match the drop down. So this UI is a bit confusing.

src/forms/CreateCircleForm.tsx
levity: Organization, in fact 😅  We want to avoid using the name "protocol" in user-facing contexts, it's a legacy naming choice

levity:
> Another question: do we have a switch or something to gate access to this as we start to roll out deploys? Or one big branch of all the FE changes that we merge and ship when we're ready?

Generally speaking a branch should be focused on a specific feature so it can be shipped all at once. If something's large enough that it spans multiple PRs, we should use a feature flag or something to hide it till it's all ready, to keep the production UX behaving seamlessly

src/routes/paths.ts
```diff
@@ -21,6 +21,7 @@ export const getMapPath = () => '/map';
 export const getVouchingPath = () => '/vouching';
 export const getHistoryPath = () => '/history';
 export const getAdminPath = () => '/admin';
+export const getCreateCirclePath = () => '/createCircle';
```
levity, line 24: why not just have these be consts when they're not parameterized?

levity: case-sensitive paths are a bit less conventional in URLs -- maybe `create-circle` or `new-circle` is better

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+      uniqBy(
+        myAdminCircles.map(({ protocol }) => protocol),
+        'id'
+      ),
+    [myAdminCircles]
+  );
+
+  const { createCircle } = useApi();
+
+  if (myAddress === undefined) {
```
levity, line 115:
```suggestion
  if (!myAddress) {
```

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+    return <></>;
```
levity, line 116: it would be good to say e.g. "Connect your wallet to begin" here rather than have the page be completely blank

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+            </div>
+            <div className={classes.twoColumnGrid}>
+              <FormTextField {...fields.user_name} label="Username" fullWidth />
+              <ApeTextField
+                label="ETH Address"
+                value={myAddress}
+                disabled
+                fullWidth
+              />
+              {myAdminCircles ? (
```
levity, line 166: you need to check emptiness, not existence, here. currently "Select from list to add to existing org" shows up even when a wallet has no circles

src/components/FormHCaptcha/FormHCaptcha.tsx
```diff
@@ -0,0 +1,70 @@
...
+}) => {
+  const classes = useStyles();
+
+  useEffect(() => {
+    if (DOMAIN_IS_LOCALHOST) {
+      console.error(
+        'hCaptcha prohibits localhost and 127.0.0.1 as supplied hostnames. See FormHCaptcha.'
+      );
+      // See: https://docs.hcaptcha.com/#localdev
+      // Put `127.0.0.1 localhost.ape` in hosts
```
levity, line 48:
```suggestion
      // Put `127.0.0.1 local-ape.host` in hosts
```

src/components/FormHCaptcha/FormHCaptcha.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }, []);
+
+  return (
+    <div className={clsx(className, classes.root)}>
+      <HCaptcha
+        sitekey={
+          TEST_MODE
+            ? '10000000-ffff-ffff-ffff-000000000001'
+            : '66997284-f648-46d2-8900-10ace3be9697'
```
levity, line 62: pull this out into an environment variable, please... i hope this isn't sensitive info? it's public now

src/pages/AssetMapPage/AMDrawer.tsx
```diff
@@ -6,9 +6,8 @@ import { makeStyles, Button } from '@material-ui/core';
 import DnsIcon from '@material-ui/icons/Dns';
```
levity, line 6: to @crabsinger's point, this sort of unrelated refactoring might be better done in a separate PR, to reduce the amount of regression risk per deployment. in the future, you could create a branch for this first, then create the main branch for the feature off of that one.

you can even do this after the fact, as long as you have a commit with these changes that comes first. you would checkout that commit, create a new branch off of that commit, push the branch, then go back to your latest commit.

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+                label="How we can contact you?"
+              />
+              <FormTextField
+                {...fields.research_org_link}
+                label="Web link"
+                fullWidth
+              />
+              <FormTextField
+                {...fields.research_org_ens}
+                label="Snapshot ENS"
```
levity, line 235: this is going to confuse people... was it already in the Typeform? if not, i recommend omitting it

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+            <div className={classes.twoColumnGrid}>
+              <FormTextField
+                {...fields.research_contact}
+                fullWidth
+                label="How we can contact you?"
+              />
+              <FormTextField
+                {...fields.research_org_link}
+                label="Web link"
```
levity, line 230:
```suggestion
                label="Your organization's website"
```

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+                fullWidth
+              />
+              <FormTextField
+                {...fields.research_org_ens}
+                label="Snapshot ENS"
+                fullWidth
+              />
+            </div>
+            <FormHCaptcha {...fields.h_captcha_token} />
```
levity, line 239: style suggestion: don't put `h` in all the names... unless you're expecting to have multiple captcha implementations in this codebase, it's sufficient to just refer to `captcha` and treat the fact that it's specifically hCaptcha as an implementation detail
(not a merge blocker)

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+              address: myAddress,
+              research_what_needs,
+              research_how_use,
+              research_org_structure,
+              research_org_link,
+              research_org_ens,
+              research_contact,
+              ...params,
+            })
+          )
```
levity, line 148: i see you have a todo item "Add the success modal"--my suggestion would be that there's no need for a modal; if there are no errors to show, redirect to the circle admin page

src/pages/CreateCirclePage/index.tsx
```diff
@@ -0,0 +1,2 @@
+import CreateCirclePage from './CreateCirclePage';
```
levity, line 1: i'm just bikeshedding now, but this boilerplate could be skipped if you just had `pages/CreateCirclePage.tsx` (or less redundantly, `pages/CreateCircle.tsx`)

src/pages/LandingPage/LandingPage.tsx
```diff
@@ -412,24 +423,22 @@ const LandingPage = () => {
                 variant="contained"
                 color="secondary"
                 size="small"
-                disableElevation={isDownXs}
+                disableElevation
+                startIcon={<DiscordIcon />}
               >
-                Get Started
+                Join our Discord
```
levity, line 429: this button still links to the typeform

src/pages/LandingPage/LandingPage.tsx
```diff
@@ -412,24 +423,22 @@ const LandingPage = () => {
...
+              </Button>
+            </a>
+            <a href={APP_URL}>
+              <Button
+                variant="contained"
+                color="secondary"
+                size="small"
+                disableElevation
+              >
+                Launch Coordinape
```
levity, line 439:
```suggestion
                Launch App
```

levity: i'm aware this was like this already

src/recoilState/appState.ts
```diff
@@ -136,15 +136,24 @@ export const rMyCircles = selector<ICircle[]>({
   key: 'rMyCircles',
   get: ({ get }: IRecoilGetParams) => {
     const myProfile = get(rMyProfile);
-    return (
-      myProfile?.users
-        ?.map((u) => createCircleWithDefaults(u.circle))
-        .sort((a, b) => a.protocol_id - b.protocol_id) ?? []
-    );
+    return (myProfile?.users ?? [])
```
levity, line 139: style suggestion: `if (!myProfile?.users) return [];` and then `return myProfile?.users.map...`

levity (changes requested):

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,263 @@
...
+                  fullWidth
+                />
+              )}
+              <FormTextField
+                {...fields.circle_name}
+                label="Circle Name"
+                fullWidth
+              />
+            </div>
+            <h3 className={classes.sectionTitle}>Optional Fields</h3>
```
levity, line 193: if the user's already a circle admin, should we skip all the research fields?

src/routes/paths.ts
exrhizo: I got that pattern from a library I was using in another project, some paths are parameterized, it could all be constants

exrhizo: I changed it to use kabab case, I'd want to look up that other lib before changing to all constants. I think the point is that you can use them in the same way if they are the same type

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: I'd rather just roll those into my own commit, I don't like creating a ton of commits on a branch

Also, I don't like javascripts magic conversion of everything to truthy vs non-truthy. I prefer to use operators that make it explicit, but I did this

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: I meant for the DefaultPage to show instead. In the future I want to refactor the routes.

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: Good catch

exrhizo: My aversion to implicit truthy conversion wouldn't have saved me here, using `!!`

src/components/FormHCaptcha/FormHCaptcha.tsx
exrhizo: yes

src/pages/AssetMapPage/AMDrawer.tsx
exrhizo: This isn't unrelated refactoring, this is collecting usages into a component

If you want me to code more responsibly, I will need less aggressive deadlines

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: talked to Alx and removed these

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: removed

src/components/FormHCaptcha/FormHCaptcha.tsx
exrhizo: How could it be a secret, it gets sent to the client

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: Did that

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: eh, I did it before reading this

src/pages/CreateCirclePage/index.tsx
exrhizo: I think that having folders is good to be consistent when there are more large pages

src/pages/LandingPage/LandingPage.tsx
exrhizo: Yes, good catch

src/pages/LandingPage/LandingPage.tsx
exrhizo: Okay

src/recoilState/appState.ts
exrhizo: thats more lines

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: It's not so intense now..
I am not opposed to this, just don't feel strongly to do it

src/pages/CreateCirclePage/index.tsx
exrhizo: It could be done like the components

src/components/ApeTextField/ApeTextField.tsx
```diff
@@ -37,7 +39,7 @@ const useStyles = makeStyles((theme) => ({
     },
   },
   inputRootError: {
-    border: `1px solid ${theme.colors.red}`,
+    border: `1px solid ${theme.colors.red}dd`,
```
crabsinger, line 42: `dd` is that intentional?

crabsinger (commented):
looked over this again. don't have much to add on the code side. tested the flow on Vercel and looks pretty good.

One bug: it doesn't seem to redirect me to the new circle admin page after creation. It redirects me to a circle admin page, but not the new one I created.

src/components/ApeTextField/ApeTextField.tsx
exrhizo: i know this is hacky, I don't like import transparentize to add a little transparency

exrhizo: Ideally I'd like the theme to have `redFade80` or something -it'd be good to cleanup styles everywhere, and alx wants to change fonts even

levity: this was new to me too -- https://caniuse.com/?search=rrggbbaa

crabsinger: Wow cool. @exrhizo I just didn't know about alpha RGBA CSS.

src/components/RecoilAppController/RecoilAppController.tsx
```diff
@@ -132,10 +132,10 @@ export const RecoilAppController = () => {
   }, [web3Context.library]);
 
   useEffect(() => {
-    if (selectedCircleId !== undefined) {
+    if (selectedCircleId !== undefined && myProfile) {
```
levity, line 135: is there any risk of recreating the race condition that you fixed in [35d5079]?(https://github.com/coordinape/coordinape/commit/35d5079555bcfec2eeec3f921d5f69d7a71da8bb#diff-e98035a9e3e99aed9d5cdae22c5662c49215d334c22be8439c240720dc93fb3cL123)

src/pages/AdminPage/AdminPage.tsx
```diff
@@ -122,6 +126,12 @@ const useStyles = makeStyles((theme) => ({
       color: '#4e7577',
     },
   },
+  tablePlaceholderTitle: {
+    fontSize: 40,
```
levity, line 130:
```suggestion
    fontSize: 20,
```

levity: ^ so it doesn't compete with the page title

src/pages/AdminPage/AdminPage.tsx
```diff
@@ -400,6 +440,22 @@ const AdminPage = () => {
         perPage={15}
         filter={filterUser}
         sortable
+        placeholder={
```
levity, line 443: this will never appear, because the admin themselves is in the circle already

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,223 @@
...
+                disabled
+                fullWidth
+                infoTooltip="Uses the address of your connected wallet"
+              />
+              <FormTextField
+                {...fields.circle_name}
+                label="Circle Name"
+                fullWidth
+              />
+              {myAdminCircles.length ? (
```
levity, line 163: > I don't like javascripts magic conversion of everything to truthy vs non-truthy

🤔
😆
i'm just joshing you, this is fine with me

src/utils/domain.ts
```diff
@@ -22,11 +30,13 @@ export const APP_URL =
 export const STORAGE_URL = (process.env
   .REACT_APP_S3_BASE_URL as string).replace(/\/$/, '');
 export const API_URL = process.env.REACT_APP_API_BASE_URL as string;
+export const API_IS_PRODUCTION = API_URL === 'https://coordinape.me/api';
```
levity, line 33: this should not be hardcoded. you can use process.env.NODE_ENV for this ([docs](https://create-react-app.dev/docs/adding-custom-environment-variables/))

src/pages/CreateCirclePage/CreateCirclePage.tsx
levity: the dark background looks good

src/components/FormHCaptcha/FormHCaptcha.tsx
levity: 🤦‍♂️  lol

levity: i'm going to add this to the Vercel production environment variables

src/pages/AssetMapPage/AMDrawer.tsx
levity: > This isn't unrelated refactoring, this is collecting usages into a component

ah yes, you're right, sorry, i was going a bit too quickly

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -0,0 +1,232 @@
...
+                />
+              ) : (
+                <FormTextField
+                  {...fields.protocol_name}
+                  label="Organization Name"
+                  fullWidth
+                  infoTooltip="A circle admin can add to an existing organization."
+                />
+              )}
+              <FormTextField
```
levity, line 190: gotta hand it to you btw @exrhizo, adding these two fields was stupid-easy. nice work on the form controller!

src/components/RecoilAppController/RecoilAppController.tsx
exrhizo: This whole flow needs a rethink. There is the bug Regan found, which I think this fixed, but then I found an entirely different one..

I'd like to rewrite all of this logic so it is less prone to race conditions, and it intersects with the access control ticket w/ zashton, and exploring using never ending promises as initial value of some atoms. We could also store all the wallets people have used and make it easier to manage multiple wallets. And! people should be able to view other circles in their org!

src/pages/AdminPage/AdminPage.tsx
exrhizo: I noticed :/

src/pages/CreateCirclePage/CreateCirclePage.tsx
exrhizo: yay!

levity (dismissed):

levity (approved):

PR #187, opened 2021-09-27
Add plausbile analytics to all pages

crabsinger:
#1 Add plausbile analytics to all pages

Add https://plausible.io analytics

Reviewers: @exrhizo @levity

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -31,6 +31,7 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Coordinape</title>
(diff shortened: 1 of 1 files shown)


levity (approved):

levity:
what will happen in development? is it smart enough to ignore those requests?

exrhizo:
That was easy

crabsinger:
@levity 

> Running on localhost? Our script automatically disables itself when running on localhost as the majority of people don't want those stats to be counted. If you do want to track stats on localhost, you can use our extension.

PR #189, opened 2021-10-04
log API gets

levity:
so we can see that there are some API requests happening too early, before a wallet is connected


levity:
This was just for my understanding, not meant to go to prod

PR #190, opened 2021-10-05
Remove Coinbase Wallet option

crabsinger:
https://linear.app/coordinape/issue/DEV-135/coinbase-wallet-signature-not-authorized

Remove Coinbase Wallet connection option - because it doesn't work.

Maybe if we use scaffold-eth we can use all their wallet options that hopefully are fully supported.

Reviewers: @exrhizo @levity


exrhizo:
This worked in the past :|

PR #193, opened 2021-10-06
DEV-292

levity:
Reorder fields in new user modal

### DIFF
--- a/src/components/ApeToggle/ApeToggle.tsx
+++ b/src/components/ApeToggle/ApeToggle.tsx
@@ -5,6 +5,8 @@ import uniqueId from 'lodash/uniqueId';
 
 import { makeStyles, ButtonGroup, Button } from '@material-ui/core';
 
+import { ApeInfoTooltip } from 'components';
+
 const useStyles = makeStyles((theme) => ({
   root: {
     display: 'flex',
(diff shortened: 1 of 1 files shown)


thenerdcat (approved):
Changes pushed. I also had to remove height:18 in the helperBox class for fixing an issue with the margin of the modal input fields

levity:
Going forward, you'll want to create different branches for different tasks so we don't accidentally merge work in progress

src/components/ApeTextField/ApeTextField.tsx
```diff
@@ -71,7 +71,6 @@ const useStyles = makeStyles((theme) => ({
     padding: theme.spacing(0, 1),
   },
   helperBox: {
-    height: 18,
```
exrhizo, line 74: I had done this to get the create circle page to have it so when errors appeared, they didn't reflow the Ui, so if this is wierd in other places, I'd rather use the fixed height everywhere and adjust the spacing on the grid. We don't really have a unified standard for these types of things, something needing to work with our designer on.

exrhizo: Also these can all switch to the inline errors, which should work except for epoch has really long errors..

PR #195, opened 2021-10-07
Clean up useAllocation to fix allocation not working

exrhizo:
https://linear.app/coordinape/issue/DEV-286/allocation-buttons-not-working-buttons-are-disabled

The `useAllocation` hooks were not easily readable,
so I rewrote them to be more simple and organized.

Now the give page shows cards for teammates that are in the local state's `pendingGive` that way there can't be a discrepancy between what there are controls for and the underlying state.

when the teammate selector is disabled, there is still a teammates data structure, it is just set to be all available teammates and the hook throws an error if you try to change them.

### DIFF
--- a/src/hooks/useDeepChangeEffect.ts
+++ b/src/hooks/useDeepChangeEffect.ts
@@ -1,0 +1,18 @@
+import { useEffect, DependencyList } from 'react';
+
+import isEqual from 'lodash/isEqual';
+
+import { usePrevious } from 'hooks';
+
+/**
+ * It's like useEffect but runs when the deps list has a deep change.
+ * See: {@link isEqual}.
+ */
+export function useDeepChangeEffect(func: () => void, deps: DependencyList) {
+  const prev = usePrevious(deps);
+  useEffect(() => {
+    if (!prev || !isEqual(deps, prev)) {
+      func();
+    }
+  }, deps);
+}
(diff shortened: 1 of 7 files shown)


src/hooks/useDeepChangeEffect.ts
```diff
@@ -0,0 +1,18 @@
+import { useEffect, DependencyList } from 'react';
+
+import isEqual from 'lodash/isEqual';
+
+import { usePrevious } from 'hooks';
+
+/**
+ * It's like useEffect but runs when the deps list has a deep change.
```
levity, line 8:
```suggestion
 * It's like useEffect but runs only when the deps list has a deep change.
```

src/hooks/useDeepChangeEffect.ts
```diff
@@ -0,0 +1,18 @@
...
+ * See: {@link isEqual}.
+ */
+export function useDeepChangeEffect(func: () => void, deps: DependencyList) {
+  const prev = usePrevious(deps);
+  useEffect(() => {
```
levity, line 13: should this support hook cleanup? https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1

exrhizo: might as well

levity:
btw, does this still play nice with the changes around allocation you made in #197? just checking

exrhizo:
> btw, does this still play nice with the changes around allocation you made in #197? just checking

I'd push this first and it probably needs a manual merge, but it would be minor in complexity

PR #196, opened 2021-10-07
Add the sentry integration that will report errors

exrhizo:
Try this out and see what we learn.


crabsinger:
@exrhizo what will sentry/integrations add in terms of error reporting vs what we have now?

levity (approved):

PR #197, opened 2021-10-08
Improve error handling and move api hooks

exrhizo:
### Moved functions that call API into useApi
 - Move towards separating state from mutations
 - `useMe`: `updateProfile`, `updateAvatar`, `updateBackground`

### Error Handling cleanups
 - Provider types included
 - When signature fails, fail immediately with the error message
 - If allocating give, mention the smaller transaction size there
 - wrap all `useAsyncLoadCatch.ts` usages with `try{}catch(e){}` or `.catch()`
 - just show a info alert if the user rejects the signature

### DIFF
--- a/src/pages/AdminPage/AdminCircleModal.tsx
+++ b/src/pages/AdminPage/AdminCircleModal.tsx
@@ -173,9 +173,13 @@ export const AdminCircleModal = ({
   };
 
   const editDiscordWebhook = async () => {
-    const _webhook = await getDiscordWebhook();
-    setWebhook(_webhook);
-    setAllowEdit(true);
+    try {
+      const _webhook = await getDiscordWebhook();
+      setWebhook(_webhook);
+      setAllowEdit(true);
+    } catch (e) {
+      console.warn(e);
+    }
   };
 
   const onChangeWith = (set: (v: string) => void) => (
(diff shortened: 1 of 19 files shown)


src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -75,9 +75,11 @@ export const NewNominationModal = ({
       name,
       address,
       description,
-    }).then(() => {
-      onClose();
-    });
+    })
+      .then(() => {
```
levity, line 79: this could just be `then(onClose)` (non-merge-blocking style feedback)

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -75,9 +75,11 @@ export const NewNominationModal = ({
...
+        onClose();
+      })
+      .catch((e) => console.warn(e));
```
levity, line 82: likewise `catch(console.warn)` (non-merge-blocking style feedback)

src/services/api.ts
```diff
@@ -25,13 +26,13 @@ import {
 axios.defaults.baseURL = API_URL;
 
 export class APIService {
-  provider = undefined;
+  provider = undefined as Web3Provider | undefined;
```
levity, line 29: is there a way to type this just as `Web3Provider`? we'd never want it to be `undefined`, right?

src/utils/provider.ts
```diff
@@ -1,39 +1,46 @@
-import { Wallet, ethers } from 'ethers';
+import { Web3Provider } from '@ethersproject/providers';
+import { ethers } from 'ethers';
 
-export const getSignature = async (data: string, provider?: any) => {
+export const getSignature = async (data: string, provider?: Web3Provider) => {
```
levity, line 4: likewise here if we make `provider` non-optional then we can drop the following null check

levity (changes requested):
the question about `await` is the only thing that might be blocking

src/services/api.ts
exrhizo: I think you need to be connected with a wallet to have a provider

src/utils/provider.ts
exrhizo: Same, this is going to be out of scope for me here

src/pages/AdminPage/AdminCircleModal.tsx
```diff
@@ -189,46 +193,50 @@ export const AdminCircleModal = ({
...
+        allocText !== circle.allocText ||
+        allowEdit ||
+        defaultOptIn !== circle.default_opt_in ||
+        nominationDaysLimit !== circle.nomination_days_limit ||
+        allocText !== circle.allocText ||
+        vouchingText !== circle.vouchingText ||
+        onlyGiverVouch !== circle.only_giver_vouch ||
+        teamSelection !== circle.team_selection
+      ) {
+        updateCircle({
```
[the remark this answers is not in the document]
exrhizo: Good call

exrhizo:
@levity made the changes you requested before merging

PR #200, opened 2021-10-08
Update react-script to v5 alpha for Webpack 5 support

hammadj:
[craco](https://github.com/gsoft-inc/craco) used to polyfill node APIs that webpack 5 doesn't polyfill out of the box anymore. Webpack 5 needed to address bundler bugs with Ceramic libraries.

[Relevant discussion from CRA](https://github.com/facebook/create-react-app/discussions/11278)

Test Plan: ensure project still builds and runs without errors

PR #203, opened 2021-10-12
Cleanup and Componentize Allocation Cards

exrhizo:
Based on https://github.com/coordinape/coordinape/pull/200

## Cleanup and Componentize Allocation Cards

In the past the user saw their own card, add this back in with one component: `ProfileCard`

### Split off components:
 - CardInfoText
 - ProfileSkills
 - ProfileSocialIcons
 - ThreeDotMenu

#### Use Icons for Social Icons
 - Github
 - Telegram
 - weblink

### DIFF
--- a/src/components/ThreeDotMenu/ThreeDotMenu.tsx
+++ b/src/components/ThreeDotMenu/ThreeDotMenu.tsx
@@ -4,14 +4,6 @@ import { IconButton, Button, Popover, makeStyles } from '@material-ui/core';
 import MoreHorizIcon from '@material-ui/icons/MoreHoriz';
 
 const useStyles = makeStyles(theme => ({
-  moreButton: {
-    // margin: 0,
-    // padding: theme.spacing(0, 1),
-    // minWidth: 20,
-    // fontSize: 17,
-    // fontWeight: 800,
-    // color: theme.colors.text,
-  },
   morePaper: {
     width: 170,
     padding: theme.spacing(1, 0),
@@ -46,7 +38,6 @@ export const ThreeDotMenu = ({ actions }: { actions: IAction[] }) => {
     <>
       <IconButton
         size="small"
-        className={classes.moreButton}
         onClick={({ currentTarget }) => setAnchorEl(currentTarget)}
       >
         <MoreHorizIcon />


levity:
looks like there are some linter errors (lemme know if this comment is redundant)

src/components/ProfileCard/GiftInput.tsx
```diff
@@ -0,0 +1,160 @@
...
+
+  const onChangeNote = ({
+    target: { value },
+  }: React.ChangeEvent<HTMLTextAreaElement>) => updateGift({ note: value });
+
+  return (
+    <div className={classes.root}>
+      {tokens !== undefined ? (
+        <ApeTextField
+          label={`${selectedCircle?.tokenName} Allocated`}
```
levity, line 100: it's weird to me to see strings like this because that implies it's expected behavior to render "undefined Allocated" in some circumstances

levity: Maybe the whole component should short-circuit and render nothing if !selectedCircle.

Or perhaps components like this should depend upon normal props for their circles rather than using a hook. 

This is the sort of thing that I was getting at when I said earlier today that hooks are a cursed abstraction. Functional components are powerful and composable for the same reason that functional programming in general is powerful and composable: with a pure function, you know that its output depends only on its input, not any non-local state. So it's easy to reason about and control, because the scope of things you have to consider in order to understand what a piece of code will do is clearly defined by the code itself. But hooks allow functional components to depend on non-local state, so this benefit goes out the window. \</soapbox>

src/components/ProfileSkills/ProfileSkills.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const ProfileSkills = ({
+  user,
+  max = 3,
+}: {
+  user: IUser;
+  max: number;
+}) => {
+  const classes = useStyles();
+  const skills = user?.profile?.skills ?? ([] as string[]);
```
levity, line 31: in this case you're typing `user` as `IUser` so the first `?` is unnecessary, no?

src/components/ProfileSkills/ProfileSkills.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+  return (
+    <>
+      {skills.length > 0 &&
+        skills.slice(0, max).map(skill => (
+          <div key={skill} className={classes.skillItem}>
+            {skill}
+          </div>
+        ))}
+      {user?.role == 1 ? (
```
levity, line 41: this can be more concise as `{user.role == 1 && <div>...</div>}`
also for readability's sake we should use a named const/enum instead of 1

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -0,0 +1,197 @@
...
+  user: IUser;
+  tokens: number;
+  note: string;
+  disabled?: boolean;
+  updateGift?: TUpdateGift;
+  isMe?: boolean;
+}) => {
+  const classes = useStyles();
+  const history = useHistory();
+  const selectedCircle = useSelectedCircle();
```
levity, line 123: in addition to the reasons above, we probably want to start limiting the use of this hook ASAP (i.e. pushing it higher up in the component hierarchy), because the designs for the vault/org UI have some pages that will need to render data for multiple circles at once.

so the more our middle- to low-level components are decoupled from the assumption that there's only one "selected" circle at a time, the more reusable they'll be.

src/components/ProfileCard/GiftInput.tsx
exrhizo: I think normal props make sense for "dumb" components.

I'm imagining you long ago `<soapbox>`
Good point.

src/components/ProfileSkills/ProfileSkills.tsx
exrhizo: I don't like imports.. lol - Okay done!

src/components/ProfileCard/ProfileCard.tsx
exrhizo: Pushing it up the hierarchy makes sense to me. I don't think it is actually a problem with current components that might be reused.

src/hooks/useNavigation.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @returns functions that create onClick handlers.
+ *
+ * Example:
+ * const { getToMap } = useNavigation();
+ * ...
+ * return (<Button onClick={getToMap({hightlight: '0x5453424324bbbecdea'})}>To map</Button>);
+ *
+ * The handler will respond to crtl & meta click for new window.
+ */
+export const useNavigation = () => {
```
levity, line 19: interesting! when do you think this should be used instead of an `<a>` tag?

levity (dismissed):

src/hooks/useNavigation.ts
exrhizo: A single source of truth for navigation (sorta along with paths.ts)

You can use it like `getToProfile({ address: 'me' })()` if wanted.

when you use an `a` tag that changes style.. I could do a global clear style on a tags.

There is also the Link component, but then that is importing both something from paths and something from router..

I think that it would make sense to standardize something with router `state` that allows adding things that don't live in the url but maybe not because I think that highlighting already makes sense to be a search param

levity (approved):

PR #209, opened 2021-10-13
feat: create merkle distributor hook for an epoch

Niraj-Kamdar:
I have created the `useMerkleDistributor` which takes `epoch_id` as an argument and produce merkle distributor info for that epoch.

I have copied most of the `Uniswap Merkle Distributor` code and made some changes because it was using a discontinued library.

Example Merkle Tree for epoch 8 on Test circle on staging env:
```json
{
  "merkleRoot": "0x307864356433363035363564386137663730323061376637633137303765303031323830303337353230373135393130363234643261623830343466643139653337",
  "tokenTotal": "0x01f1",
  "claims": {
    "0x0dC053c330918C3c4Cb556922973ab21F567919B": {
      "index": 0,
      "amount": "0x53",
      "proof": [
        "0x4735d14dac8e9ee13004dfb155e381af4a0dfcd22d6e375aa6033676326ed30d",
        "0x307832613862653436373861623732386534663762333866633232333165353035366164386265303066343035306531666135666665373533386236616433363534",
        "0x307862643265343638363933653438373135366636646636653131646234663662316438323163653062393831316339623539623962626563323632346334353039"
      ]
    },
    "0x2823E46A075Ad7eDB3144097A8DCB48Cd8c00602": {
      "index": 1,
      "amount": "0x0c",
      "proof": [
        "0x11fbecefba204f171baf9e5211ff3f6cbaeff48f61f5e57b1ddda23dceaca9f2",
        "0x307835613237613766323834666437643136643732336536393334383461393163663562653532633765376361356238313532656432633837653931356466376435",
        "0x307862643265343638363933653438373135366636646636653131646234663662316438323163653062393831316339623539623962626563323632346334353039"
      ]
    },
    "0x47cA1711F0F79dD970053Db79646025a2BfCb385": {
      "index": 2,
      "amount": "0x34",
      "proof": [
        "0xdcdfc6365b6b613a289ed5debe2357b718c53306819b219ee76b3325226c0120",
        "0x307836303832343030613661633030623262376431613135366263313637356332616333346363306233363963323737356139383366653136663963613139613664"
      ]
    },
    "0xA3849B6c2242d46211BD26D6cB949c25E07F95AC": {
      "index": 3,
      "amount": "0x0114",
      "proof": [
        "0xcd73f5c763fd93ba64b4fe5f6c79f2b3ac98d9144beaecb1337b26a09e5aa659",
        "0x307836303832343030613661633030623262376431613135366263313637356332616333346363306233363963323737356139383366653136663963613139613664"
      ]
    },
    "0xf521Bb7437bEc77b0B15286dC3f49A87b9946773": {
      "index": 4,
      "amount": "0x47",
      "proof": [
        "0x41b75c049ebcd5fd8e788f868eef96bfad2956834f46e6d45b5ce9b223a322fb",
        "0x307835613237613766323834666437643136643732336536393334383461393163663562653532633765376361356238313532656432633837653931356466376435",
        "0x307862643265343638363933653438373135366636646636653131646234663662316438323163653062393831316339623539623962626563323632346334353039"
      ]
    },
    "0xfA20d0E606E3ED73daE6Ce3B3744Cbe4B25A4002": {
      "index": 5,
      "amount": "0x03",
      "proof": [
        "0x94361ceaa7a1c4f28c1d8bab36bd7e3a33799dd03525e7053983cd512c7c0c38",
        "0x307832613862653436373861623732386534663762333866633232333165353035366164386265303066343035306531666135666665373533386236616433363534",
        "0x307862643265343638363933653438373135366636646636653131646234663662316438323163653062393831316339623539623962626563323632346334353039"
      ]
    }
  }
}

```


levity:
can you create a test that shows how `parseBalanceMap` generates the expected output for a given set of inputs? you can add it right alongside the file as `parse-balance-map.test.ts`; then `yarn test` should notice it and run it automatically

levity:
btw, be sure to pull in the latest changes from master and run `yarn` again so that linting & formatting will run automatically when you commit next

src/recoilState/merkleDistributor.ts
```diff
@@ -0,0 +1,22 @@
...
+  MerkleDistributorInfo,
+  parseBalanceMap,
+} from 'helpers/merkle-distributor/parse-balance-map';
+import { selectorFamily } from 'recoil';
+
+import { rReceivedGiftsByUser } from './receivedGiftsByUser';
+
+import { IRecoilGetParams } from 'types';
+
+export const rMerkleDistributor = selectorFamily<
```
levity, line 11: maybe `rEpochRoot` is a better name for this

src/hooks/useMerkleDistributor.ts
```diff
@@ -0,0 +1,6 @@
+import { useRecoilValue } from 'recoil';
+
+import { rMerkleDistributor } from 'recoilState/merkleDistributor';
+
+export const useMerkleDistributor = (epochId: number) =>
+  useRecoilValue(rMerkleDistributor(epochId));
```
exrhizo, line 6: Just inline this in the recoilState/appState

recoil is global so, let's keep all those there

src/hooks/useReceivedGiftsByUser.ts
```diff
@@ -0,0 +1,6 @@
+import { useRecoilValue } from 'recoil';
+
+import { rReceivedGiftsByUser } from 'recoilState/receivedGiftsByUser';
+
+export const useReceivedGiftsByUser = (epochId: number) =>
+  useRecoilValue(rReceivedGiftsByUser(epochId));
```
exrhizo, line 6: same

src/recoilState/merkleDistributor.ts
exrhizo: `rEpochMerkleRoot`

src/recoilState/merkleDistributor.ts
```diff
@@ -0,0 +1,22 @@
...
+  MerkleDistributorInfo | undefined,
+  number
+>({
+  key: 'rReceivedGiftsByUser',
```
exrhizo, line 15: The key needs to change, and needs to be the same as the variable name

src/recoilState/merkleDistributor.ts
```diff
@@ -0,0 +1,22 @@
...
+  get: (epochId: number) => ({ get }: IRecoilGetParams) => {
+    const receivedGiftsByUser = get(rReceivedGiftsByUser(epochId));
+    if (Object.keys(receivedGiftsByUser).length !== 0) {
```
exrhizo, line 18: is it possible to create an empty tree?
I want to move away from having `| undefined`

exrhizo: maybe that is odd, I think we will preview the distribution before upload

src/recoilState/receivedGiftsByUser.ts
```diff
@@ -0,0 +1,42 @@
...
+    const gifts = get(rGifts).filter((g) => g.epoch_id === epochId);
+    const giftsByUser = iti(gifts).toGroups(
+      (g) => g.recipient_id,
+      (g) => g.tokens
+    );
+    const totalReceivedByUserMap = iti(giftsByUser.keys()).toMap(
+      (userId) =>
+        assertDef(
+          usersMap.get(userId)?.address,
+          'Missing user while generating Merkel Root'
```
exrhizo, line 28: This could be used by something else, call it whatever the selector is called rather than Merkel Root

src/recoilState/merkleDistributor.ts
```diff
@@ -0,0 +1,22 @@
+import {
```
exrhizo, line 1: Don't have separate files for these in `recoilState` just put it in appState for now, later I might reorganize them

src/helpers/bufferToHex.ts
```diff
@@ -0,0 +1,3 @@
+export const bufferToHex = function (buf: Buffer): string {
```
exrhizo, line 1: This is too small for it's own file, would rather it be in `tools.ts` 

Similar this could be reorganized

exrhizo (commented):
I still need to read into the `merkel-distributor/*` files.

Are these all directly copied? if so, can we have them as a library?

Niraj-Kamdar:
> I still need to read into the `merkel-distributor/*` files.
> Are these all directly copied? if so, can we have them as a library?

Mostly but since they were using unmaintained dependencies, I have made some changes. 
No, we can't use it as library because it's a standalone command line program and they haven't released it as a library

src/recoilState/merkleDistributor.ts
Niraj-Kamdar: No it's not possible. If we don't supply any input, we can't generate any merkle tree.

src/recoilState/appState.ts
```diff
@@ -500,3 +505,70 @@ export const rUserGifts = selectorFamily<
 });
 export const useUserGifts = (userId: number) =>
   useRecoilValue(rUserGifts(userId));
+
+export type IReceivedGiftsByUser = { [account: string]: number | string };
+
+// Todo: Fetch this data from API when available
+export const rUnclaimedTokensByUser = atom<Record<string, number>>({
```
levity, line 512: the data we need to add to the current merkle root is not just the amount of unclaimed tokens -- it's actually the entire merkle root for the previous epoch. so the code would be something like this:

```js
const previousRoot = fetchPreviousEpochRoot(); // TODO from API or subgraph
const newRoot = parseBalanceMap(receivedGiftsByUser); // only the ones from the current epoch

// this is what you need to implement, and what gets sent to the contract
const combinedRoot = combineRoots(previousRoot, newRoot); 
```

levity (dismissed):

src/recoilState/appState.ts
```diff
@@ -500,3 +505,70 @@ export const rUserGifts = selectorFamily<
...
+  get:
+    (epochId: number) =>
+    ({ get }: IRecoilGetParams) => {
+      const usersMap = get(rUsersMap);
+      const gifts = get(rGifts).filter(g => g.epoch_id === epochId);
+      const giftsByUser = iti(gifts).toGroups(
+        g => g.recipient_id,
+        g => g.tokens
+      );
+      const unclaimedTokensByUser = get(rPreviousEpochGiftsByUser);
```
levity, line 533:
```suggestion
      const previousTokensByUser = get(rPreviousEpochGiftsByUser);
```

levity:
there's another part of the spec that was left out, which is the need to specify a total token amount when generating the merkle root. i'm going to work on this

src/lib/README
```diff
@@ -0,0 +1,2 @@
+This folder is for storing relatively stand-alone pieces of code, pieces that
```
exrhizo, line 1: TODO: move `canvas-color-tracker` here

exrhizo (commented):
Honestly, a lot for me to understand in depth, I see the tests and I appreciate that this really benefits from them.

Is it possible to mark which files are copied? Are they copied verbatim? if so, a comment at the top

exrhizo (approved):
LGTM

PR #215, opened 2021-10-16
Fetch profile data from SelfID in addition to backend

META-DREAMER:
Fallback to using profile data from SelfID when the user hasn't set their own profile data.


exrhizo (commented):
Like I said on discord, to get all the user profiles, you would need to do something at the point where `rUsersMapRaw` is loaded into `rUsersMap`

It `rUsersMap` already depends on `rMyProfile`, so you could set up something in `useCircle` when `selectCircle` is called.

Although I am going to be rewriting that soon for our switch to new access controlled endpoints. and this is good as a proof of concept at this point in time.

levity (approved):
I might be missing some context, but in isolation, this LGTM

META-DREAMER:
> Like I said on discord, to get all the user profiles, you would need to do something at the point where `rUsersMapRaw` is loaded into `rUsersMap`

@exrhizo updated PR to fetch profile data for those users as well. the only thing that really affects is the profile image since there isnt any way (AFAICT) to actually click into another contributors profile

PR #225, opened 2021-10-21
SelfID Fetch selfidprofiles with select circle

exrhizo:
# Refactor for including selfId profiles

Forked from #215

How users were being fetched in the PR I forked, would mean that when you switched circles, it would cause all the users from the last circle to be refetched from ceramic. By putting the fetch at the end of `selectAndFetchCircle` the ceramic fetches can happen asynchronously and fill in profile data when they come in.

### Cleanup useCircle
 - Have the fetchers with type inference
 - after pulling in users data, fetch the selfids

### `rUsersMap` to include any existing selfIdProfiles

Create a batch `getSelfIdProfiles`

### Poor Developer experience

@META-DREAMER  - I see why you were talking about wanting to use react query

Looking at this, it's crazy that the profile is loaded in different ways and merged arbitrarily.

Historically, the app was built without profiles, only users, and the same person would have different unconnected users, so profiles was built ontop of that.


src/utils/selfIdHelpers.ts
```diff
@@ -40,35 +41,35 @@ export const mergeSelfIdProfileInfo = (
...
     };
   } catch (e) {
-    return null;
+    return;
   }
 };
+
+export const getSelfIdProfiles = async (addresses: string[]) => {
+  return (await Promise.all(addresses.map(getSelfIdProfile))).filter(
+    p => !!p
```
levity, line 73: `p => p` works fine here too

levity:
@exrhizo looks like there's a type error blocking the build

exrhizo:
@levity I feel good merging this

PR #240, opened 2021-10-26
Fix admin reset issue

zashton:


src/forms/AdminUserForm.tsx
```diff
@@ -19,7 +19,7 @@ const schema = z
     non_giver: zBooleanToNumber,
     fixed_non_receiver: zBooleanToNumber,
     non_receiver: zBooleanToNumber,
-    admin: zBooleanToNumber,
+    role: zBooleanToNumber,
```
exrhizo, line 22: This is wrong still, it should be a boolean to a number..

PR #248, opened 2021-10-29
feat: add tests for apeVault

Niraj-Kamdar:


hardhat/test/ApeProtocol/ApeVault.ts
```diff
@@ -0,0 +1,43 @@
...
+  const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000';
+
+  let accounts: Signer[];
+
+  before(async () => {
+    accounts = await ethers.getSigners();
+    await deployments.fixture();
+  });
+
+  it('Test vault creation', async () => {
```
levity, line 21: please use `test` here instead of `it`

`it` is only meant to be used when, in combination with the first argument, it makes an English sentence. Like
```
it('can be created', async () => {
```

It refers to the subject of the `describe` block, in this case 'Ape Vault'.

levity:
What command should be run to run all the tests?

hardhat/test/ApeProtocol/ApeVault.ts
Niraj-Kamdar: This is just the draft PR, I was going to improve docs afterwards.

levity:
nice progress here. when i set up `.env` with the RPC URL from my personal Infura account, i get 2/3 tests to pass. The 3rd one fails with 
```Error: VM Exception while processing transaction: reverted with reason string 'ERC20: transfer amount exceeds balance'```
but maybe you're aware of that already.

Niraj-Kamdar:
> when i set up `.env` with the RPC URL from my personal Infura account, i get 2/3 tests to pass. The 3rd one fails with
> ```
> Error: VM Exception while processing transaction: reverted with reason string 'ERC20: transfer amount exceeds balance'
> ```
> but maybe you're aware of that already.

Yes, this is weird error due to network timeout. Unfortunately I can't configure custom timeout on hardhat network. So, I am digging into it. I am not sure this is due to mainnet forking or inherent problem in hardhat. 

If you try running test one by one by commenting others, it will work fine 🥲

levity (approved):

PR #259, opened 2021-11-03
WIP Starting Web3 integrations for vault contracts

Too-Far:


levity:
hey @Too-Far, check out Niraj's branch at #248, it's already got hardhat setup to generate types; can we work with the way that's set up rather than having types in a different place in the filesystem?

levity:
#248 is merged now, so you can just work off of master. check out the README to see how to generate types; ping me in TG chat if anything's unclear.

things slightly changed in Niraj's newest PR #261, but the typechain files will still be in the same plae.

PR #261, opened 2021-11-04
fix: hardhat tests

Niraj-Kamdar:


.gitmodules
```diff
@@ -1,3 +1,3 @@
-[submodule "hardhat/coordinape-protocol"]
-	path = hardhat/coordinape-protocol
+[submodule "hardhat/contracts/coordinape-protocol"]
```
Niraj-Kamdar, line 1: I have done this to make it easy for us to add more contracts in future. We will need to add some yearn contracts for testing migration and it would be nice to add it as it's own submodule instead of including it into coordinape-protocol. Let me know if you think otherwise

hardhat/test/00-ape-protocol/00-ape-vault-factory.ts
```diff
@@ -22,4 +27,17 @@ describe('ApeVault', () => {
...
     expect(await vault.vault()).to.equal(USDC_YVAULT_ADDRESS);
   });
+
+  after(async () => {
+    await network.provider.request({
+      method: 'hardhat_reset',
+      params: [
+        {
+          forking: {
+            jsonRpcUrl: process.env.ETHEREUM_RPC_URL ?? 'http://127.0.0.1:7545',
```
levity, line 37: we should probably set blockNumber here at least to whatever the block number was at the start of the test, for the sake of caching

hardhat/test/00-ape-protocol/00-ape-vault-factory.ts
```diff
@@ -1,10 +1,15 @@
-import { expect } from 'chai';
+import chai from 'chai';
```
levity, line 1: you overwrote some of my changes here--please look at https://github.com/coordinape/coordinape/pull/248/commits/2539a5bb2cff9b219abb73090e9e2962ef613266

.gitmodules
levity: sounds good!

hardhat/test/00-ape-protocol/00-ape-vault-factory.ts
Niraj-Kamdar: sure

hardhat/test/00-ape-protocol/00-ape-vault-factory.ts
Niraj-Kamdar: If I don't use middleware, the linter will give linting error and also won't give appropriate suggestions
Screenshot 2021-11-06 at 1 01 48 PM

Niraj-Kamdar: NVM, I see hardhat-waffle will take care of this. I was just using ethereum-waffle before.

Niraj-Kamdar: When we use hardhat-waffle, hardhat inject this implicitly and linter and IDE can't recognise it that's why it's showing error

hardhat/test/00-ape-protocol/00-ape-vault-factory.ts
Niraj-Kamdar: I am researching on ways to cache forked chain but meanwhile we can set it to london hard fork.

levity (approved):

PR #263, opened 2021-11-06
fix: console log for setup

Niraj-Kamdar:


levity:
Rather than tying this to an environment, and having it be the same for all devs, it's better to use the [debug](https://www.npmjs.com/package/debug) library. it's already in the project. i'll make the change in this branch to demonstrate

levity (approved):
now if you want to show logs you can run this:
```
DEBUG="coordinape:tests" yarn hardhat:test
```

or if you want to show all debug logs starting with 'coordinape':
```
DEBUG="coordinape:*" yarn hardhat:test
```

PR #264, opened 2021-11-07
Dont imply that the loader is getting a modal

exrhizo:
I'm working my way to the login process

You'd mentioned that by having the white box, the loader looked like it was going to load something in it, no longer.


levity (approved):
Looks great 👌

PR #257, opened 2021-11-01
Fix responsive header

levity:
See #245 for context

PR #267, opened 2021-11-08
Fix responsive header fork

exrhizo:
Forked https://github.com/coordinape/coordinape/pull/257/files

Removed `styles/common.ts` because it is only used in one place.

Unimplemented the hamburger button because the header is going to be restructured significantly and we have a design now.

Did a minimal update so that we can use this, basically just increased header height on small screens.


levity (approved):
looks good on my phone

PR #278, opened 2021-11-09
Change ApeAvatar component to default to user's initials

### LINKED ISSUE #272: Change ApeAvatar component to default to user's initials
**Current Behavior:**
The ApeAvatar component currently defaults to a [placeholder image](https://github.com/coordinape/coordinape/blob/master/src/components/ApeAvatar/ApeAvatar.tsx#L22), which isn't very informative.

**Desired Behavior:**
Update the [ApeAvatar component](https://github.com/coordinape/coordinape/blob/master/src/components/ApeAvatar/ApeAvatar.tsx) to display the user's initials instead of the placeholder image.

zencephalon:
Uses https://ui-avatars.com/ to generate initial icons. It's a free API that does >800mm reqs/mo and I've used it for a few projects. It'll handle splitting a name into initials when given names with a segment, otherwise it uses the first two letters of a single name.

We can also customize the API output as desired.

Before:
Screen Shot 2021-11-09 at 3 22 35 PM

After:
Screen Shot 2021-11-09 at 3 21 05 PM

Closes #272

### DIFF
--- a/src/components/ApeAvatar/ApeAvatar.tsx
+++ b/src/components/ApeAvatar/ApeAvatar.tsx
@@ -19,18 +19,16 @@ export const ApeAvatar = ({
 }) => {
   // TODO: simplify so all: <ApeAvatar path={getAvatarPath(p?.avatar)} />
   const p = profile ?? user?.profile;
-  const src = p?.avatar ? getAvatarPath(p?.avatar) : path ?? AVATAR_PLACEHOLDER;
+  const placeholder = user?.name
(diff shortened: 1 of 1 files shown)


levity:
Looks great on the Vercel preview app! 

Btw, don't worry about the "Typecheck, Lint and Test" check failing--the smart contract tests are failing because they require an RPC URL, and I have set that up as a Github secret environment variable, which pull requests from forks can't read. I'll have to find some workaround for this situation.

levity (approved):

PR #283, opened 2021-11-10
simplify title on admin page

levity:
Scratching my own itch here... I think this looks better

Before
Screen Shot 2021-11-10 at 11 31 37 AM

After
Screen Shot 2021-11-10 at 11 32 05 AM


exrhizo (approved):
I guess with the new admin page it will have org above this

I believe org needs a treatment at some point

PR #284, opened 2021-11-10
Extract getAvatarPathWithInitialsFallback, use in ApeAvatar and AMForceGraph

### LINKED ISSUE #272: Change ApeAvatar component to default to user's initials

zencephalon:
Closes #272 

This extracts placeholder logic from `ApeAvatar` in `util/domains` and uses the new logic in both `ApeAvatar` and `AMForceGraph`.

Builds on work from https://github.com/coordinape/coordinape/pull/278

### DIFF
--- a/src/utils/domain.ts
+++ b/src/utils/domain.ts
@@ -62,3 +62,14 @@ export const getAvatarPath = (avatar?: string, placeholder?: string) => {
 
   return `${STORAGE_URL}/${avatar}`;
 };
+
+export const getAvatarPathWithInitialsFallback = (
+  avatar?: string,
+  name?: string
+) => {
+  const placeholder = name
+    ? `https://ui-avatars.com/api/?name=${encodeURIComponent(name ?? '')}`
+    : AVATAR_PLACEHOLDER;
+
+  return avatar ? getAvatarPath(avatar) : placeholder;
+};
--- a/src/utils/domain.ts
+++ b/src/utils/domain.ts
@@ -63,10 +63,7 @@ export const getAvatarPath = (avatar?: string, placeholder?: string) => {
   return `${STORAGE_URL}/${avatar}`;
 };
 
-export const getAvatarPathWithInitialsFallback = (
-  avatar?: string,
-  name?: string
-) => {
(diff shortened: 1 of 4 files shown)


src/utils/domain.ts
```diff
@@ -62,3 +62,14 @@ export const getAvatarPath = (avatar?: string, placeholder?: string) => {
 
   return `${STORAGE_URL}/${avatar}`;
 };
+
+export const getAvatarPathWithInitialsFallback = (
```
levity, line 66: how about `getAvatarPathWithFallback` for brevity's sake? also a bit more future-proof if we change the implementation

zencephalon: Yeah, sounds good to me.

levity:
hmm, I'm seeing this on the preview app with Chrome/macOS:

Screen Shot 2021-11-10 at 3 07 18 PM

zencephalon:
@levity Mm, yeah, that looks wrong. I'm finding it a bit hard to setup a circle where I can test this feature on my own. Could I get added to a circle in staging so I can try it out there?

zencephalon:
Screen Shot 2021-11-11 at 3 58 06 PM

Hey @levity thanks for taking a look yesterday. I believe I have it working today.

Also cool that U also enjoy martial arts and transpersonal psychology! I feel curious about whether we've read some of the same books because that strikes me as a particular combination of interests.

levity (approved):
nice!

levity:
> Also cool that U also enjoy martial arts and transpersonal psychology! I feel curious about whether we've read some of the same books because that strikes me as a particular combination of interests.

I bet we have! :)

PR #285, opened 2021-11-11
[WIP] Feature/de duplicates components

f1lander:
This PR removes the duplicated ApeFieldTextNew to use custom styles defined as prop types i the component.

Usage Example:
Screen Shot 2021-11-10 at 19 00 57
Screen Shot 2021-11-10 at 18 46 07


src/components/ApeTextField/ApeTextField.tsx
```diff
@@ -3,88 +3,25 @@ import React, { useState } from 'react';
...
-    width: '100%',
-    textAlign: 'center',
-  },
-}));
-
 // ApeTextField
 //
 // Using the same interface as MaterialUI's TextField to make it compatible
 // with the the calendar.
+
```
exrhizo, line 16: NIT: no new line

src/components/ApeTextField/ApeTextField.tsx
```diff
@@ -203,3 +140,113 @@ export const ApeTextField = ({
     </div>
   );
 };
+
+//#region Styles
```
exrhizo, line 144: NIT: delete comment

src/types/ape-text-field.d.ts
```diff
@@ -0,0 +1,16 @@
+import { TextFieldProps } from '@material-ui/core';
```
exrhizo, line 1: I'd have this in the same ApeTextField.tsx

When i've needed to share these types of things, I put them in `global.d.ts`

Can we keep this in `ApeTextField.tsx` and the decide how to structure this in the future. I think if we had a seperate file we'd want it in the same component folder with storybook, ect.

Like I said, I'd like us to have that type of conversation in another PR that isn't blocking, and come to a decision with a few engineers about what we want for our standards.

exrhizo (commented):
Looks good after addressing these few small things

src/components/ApeTextField/ApeTextField.tsx
f1lander: This is a region is not a comment is just a way to keep region in a file, you can colorized with some extensions in your launch

f1lander: https://github.com/microsoft/vscode/issues/12146

src/types/ape-text-field.d.ts
f1lander: I just removed from the ApeTextField.tsx since aI using one combined type

exrhizo: This is the only thing blocking me approving.

exrhizo (approved):
LGTM

levity (approved):

PR #298, opened 2021-11-16
Auth webhook for Hasura

levity:
A [Vercel serverless function](https://vercel.com/docs/concepts/functions/serverless-functions) 
- that takes an access token generated by [the PHP backend](https://github.com/coordinape/coordinape-backend/blob/main/app/Http/Controllers/ProfileController.php#L84), 
- connects to the db 
  - at `process.env.DATABASE_URL`
  - using [Prisma](https://www.prisma.io/docs/getting-started/setup-prisma/add-to-existing-project/relational-databases-typescript-postgres),
- looks up the associated profile
  - (thank goodness the token-hashing scheme is [simple](https://github.com/laravel/sanctum/blob/2.x/src/PersonalAccessToken.php#L56))
- and returns the matching profile's data.

For demonstration purposes, not to be merged in this form. 

You can try this out locally with [`vercel dev`](https://vercel.com/docs/cli#commands/dev). Note that you need to run `npx prisma generate` after making changes to the schema file and possibly the first time after switching to this branch.

Preview apps will not work until I add `DATABASE_URL` to the Vercel environment.


levity:
I added a readonly user to the staging DB and set up DATABASE_URL in Vercel to use that user, so I'll be able to test this route in the next preview deployment.

levity:
Note that this route will not work in production until DATABASE_URL is set in Vercel for that environment, which I'm probably not going to do until we've switched over to Postgres.

PR #299, opened 2021-11-16
feat: add command for minting tokens on testnet

Niraj-Kamdar:


Niraj-Kamdar:
Merging this since needed to test vault functionality

PR #304, opened 2021-11-17
Support ENS resolution in zEthAddress form helper

### LINKED ISSUE #279: ENS Name Resolution
## Improve `zEthAddress` in `formHelpers.ts`

The simplest way to do this is to have the Zod parser use asynParse and get the address for an ens if it is ens or if it is a regular address validate that. If the ens resolution fails, then that should fail the validation.

See: https://github.com/colinhacks/zod#parseasync

For this to work, I think that `createForm` will need to use `useRecoilValueLoadable` otherwise it will cause the form to suspend.

---

### Alternatively 
(This way has the drawback that validation would still need to be spread between this and the above zod parser)

Create an `FormEnsField` that wraps `ApeTextField` with the same interface as `FormTextField` and intercepts onChange and resolves ens names to an address.

The field needs to recognize the difference between an ENS name and address, and only resolve for the ens, it can display an indication if the ens name doesn't resolve.

zencephalon:
Allows inputting ENS names anywhere we take an ETH address in a form. Closes #279 

Some screenshots of it in action:

1. Error states still work

Screen Shot 2021-11-17 at 1 48 28 PM

2. ENS names now work
Screen Shot 2021-11-17 at 1 50 44 PM

3. ENS gets translated to bare address before it ends up in the database

Screen Shot 2021-11-17 at 1 50 59 PM

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -8,8 +8,8 @@ jobs:
     env:
       CI: true
       REACT_APP_S3_BASE_URL: https://coordinape.s3.amazonaws.com
-      REACT_APP_API_BASE_URL: https://myvault.live/api
-      ETHEREUM_RPC_URL: ${{ secrets.ETHEREUM_RPC_URL }}
+      REACT_APP_API_BASE_URL: https://staging-api.coordinape.com/api
+      ETHEREUM_RPC_URL: https://mainnet.infura.io/v3/275c8fc2ec294d6aac6bfc881d9a9dff
 
     steps:
       - uses: actions/checkout@v2
--- a/src/forms/createForm.tsx
+++ b/src/forms/createForm.tsx
@@ -7,7 +7,7 @@ import {
   useRecoilCallback,
   CallbackInterface,
   selectorFamily,
-  useRecoilValue,
+  useRecoilValueLoadable,
   useRecoilState,
   useSetRecoilState,
 } from 'recoil';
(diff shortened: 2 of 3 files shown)


crabsinger:
Nice work on this @zencephalon. I let Vercel run this PR, but it seems the CI is failing... kinda looks like unrelated changes though.

zencephalon:
@crabsinger I think that CI failure is one that we always get when receiving a PR from a fork, because the forked repos can’t read the environment secret for the RPC address. I think @levity has context on that.

crabsinger:
@zencephalon yes, the CI issue is a regression we introduced not related to your change. We're trying to find a good solution on-going.

crabsinger:
Can you rebase off master @zencephalon - we removed the secret requirement
https://github.com/coordinape/coordinape/commit/56b30282297c4fc14b72ec4513d0edb8afe3c685
so you should be able to get this passing CI.

Will this ens resolution work on the vouching form? cc @Zemm111 @AlexDistill

zencephalon:
> Can you rebase off master @zencephalon - we removed the secret requirement [56b3028](https://github.com/coordinape/coordinape/commit/56b30282297c4fc14b72ec4513d0edb8afe3c685) so you should be able to get this passing CI.

Ok!

> 
> Will this ens resolution work on the vouching form? cc @Zemm111 @AlexDistill

It looks like `NewNominationModal` doesn't use `createForm` at the moment, so it would take extra work to get ENS resolution working in that particular form. Seems like we could refactor that modal to use `createForm`. Happy to open an issue for it if that sounds sensible.

crabsinger (approved):

crabsinger:
@zencephalon let's do the follow up in a separate PR. I'll merge this one. Awesome contribution!

PR #306, opened 2021-11-18
refactors ReadMore component to show tooltip

f1lander:
Screen Shot 2021-11-18 at 16 02 56
Screen Shot 2021-11-18 at 16 07 18

### DIFF
--- a/src/components/ReadMore/ReadMore.tsx
+++ b/src/components/ReadMore/ReadMore.tsx
@@ -64,6 +64,7 @@ export const ReadMore: React.FC<ReadMoreProps> = (
                   See more
                 </button>
               }
+              placement="top"
             >
               {props.children}
             </ApeInfoTooltip>
--- a/src/components/ReadMore/ReadMore.tsx
+++ b/src/components/ReadMore/ReadMore.tsx
@@ -1,9 +1,9 @@
-import React, { useState } from 'react';
+import React from 'react';
 
-import { makeStyles, ClickAwayListener } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core';
 
 import { ApeInfoTooltip } from 'components/ApeInfoTooltip/ApeInfoTooltip';
-import useMobileDetect from 'hooks/useMobileDetect';
+
 interface ReadMoreProps {
   /** this property decide to hide the text and show the options see more or see less*/
   isHidden?: boolean;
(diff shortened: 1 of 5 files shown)


src/hooks/useMobileDetect.ts
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+export default function useMobileDetect() {
```
crabsinger, line 3: @f1lander Are we introducing this useMobileDetect for the first time here?

Long term, and in relation to our plan for the mobile redesign, do you think this is approach we should ideally be taking?

src/components/ReadMore/ReadMore.tsx
```diff
@@ -31,16 +33,54 @@ export const ReadMore: React.FC<ReadMoreProps> = (
...
-    <div>
-      <div className={isHidden ? textHidden : undefined}>{props.children}</div>
+    <span>
+      <div className={_isHidden ? textHidden : undefined}>{props.children}</div>
       {_isHidden && (
-        <button className={linkButton} onClick={() => setIsHidden(!isHidden)}>
-          {!isHidden ? 'see less' : 'see more'}
-        </button>
+        <>
+          {isMobile ? (
```
crabsinger, line 52: I'm not sure if this is the best way to do this - it might be - but it does feel a little brittle to me... what are the best practices for mobile design around this?

src/hooks/useMobileDetect.ts
f1lander: @crabsinger yes this is for the first time, just a simple hooks thats allow to us to detect if is a mobile device for this feature is important because I can decide to implement the clickaway listener so the "see more" option can be clickable because on touch device there is note onhover event to show the tooltip

src/components/ReadMore/ReadMore.tsx
f1lander: I'm pretty sure that this is a good practice for this particular case, I don't see another way because I'm switching to use 2 different way to implement the tooltip as explain above, in mobile is not posible onhover event to show the tooltip at least we can just go for the click event to show it.

Noticed in this is not just detect the sizes of the screen is to detect if is mobile because I need to know if the device does has touch feature, for just size we can use the actual hooks from material-ui

example:

`import { useTheme } from '@mui/material/styles';
import useMediaQuery from '@mui/material/useMediaQuery';

function MyComponent() {
  const theme = useTheme();
  const matches = useMediaQuery(theme.breakpoints.down('sm'));

  return <span>{`theme.breakpoints.down('sm') matches: ${matches}`}</span>;
}`

I'm using this for something that I started to work about the top navbar menu on mobile

levity:
@AlexDistill pls check out the new Read More implementation here

src/components/ReadMore/ReadMore.tsx
levity: @f1lander since this is basically a pattern for a tooltip that you hover on desktop and tap on mobile, do you want to move this behavior into ApeInfoTooltip itself? We would want all tooltips to have this behavior, no?

f1lander: @levity you right, that would be better! I'll do it.

levity (approved):

levity:
Uh oh 😅 
https://www.dropbox.com/s/0470bz14q42kzja/Screen%20Shot%202021-11-23%20at%204.50.27%20PM.png?dl=0

f1lander:
@levity what are the steps to reproduce this? https://www.dropbox.com/s/0470bz14q42kzja/Screen%20Shot%202021-11-23%20at%204.50.27%20PM.png?dl=0 in which browser?

levity:
Chrome on MacOS. The full text was:

> Had a nice little dev spike: wrote a Vercel serverless function to do token authorization using the Prisma ORM, for use with Hasura. Only worked Mon & Tues. https://www.notion.so/coordinape/99185d7d93c84826a31766a5b24c2c85?v=2e3f6621ee554aea8654c8c43ae1a258&p=dda852417e1f4d17993875aad8cceaf7

> @levity what are the steps to reproduce this? https://www.dropbox.com/s/0470bz14q42kzja/Screen%20Shot%202021-11-23%20at%204.50.27%20PM.png?dl=0 in which browser?

levity:
Maybe it only happens when it's on the user's own profile card?

f1lander:
@levity hmm maybe, I will try later when I get home

AlexDistill:
I see the same on my end. Also I am noticing that "See More" link is shown
even on profiles with very short text. I think it should only be show in
the full text would not fit

AlexDistill:
Also a small visual detail (sorry to be sending this late, it's just the first moment I've had to review the fix)

Can we change the color of the "Info" icon to #939EA1 and align it with the center of the "Coordinape" text 
image

PR #311, opened 2021-11-22
add tooltips to AdminCircleModal

jsur:
- Added hover styles to `ApeInfoTooltip`
- Added `YesNoToolTip` and content to `AdminCircleModal`


jsur:
Ah, missed a thing, will create a new one.

PR #315, opened 2021-11-23
Test implementation

zashton:
@levity
added test implementation that can be ran with node command
`npx ts-node -O '{"module":"commonjs"}' api/auth.ts`


levity:
hey @zashton, i'm a little unclear about what you're trying to accomplish here. the spike PR i wrote was already correctly parsing a token. I should have documented that you can test it by running `vercel dev` and 
```
http POST :3000/api/auth token="4|GrZjBcKo31BcnrarFzXa1zVOSGwSceQuxZ8bN1z9"
```
(`http` is https://httpie.io/cli -- easier to use than `curl`)

all you need to do is tweak how the token is read-- get it from "Authorization: Bearer..." instead of "token"-- and change the response to add the X-Hasura values.

other things about this PR:
- not sure why there is a migration file -- you want to be testing against a copy of the staging db, not a blank db
- make sure you use `yarn` instead of `npm`
- it looks like the auto-linters aren't working for the `api` folder, because your formatting is way off from  the rest of the repo, e.g. 4 spaces instead of 2. i'll try to fix that

levity:
I just pushed a fix to my original branch so the formatting checks run in the `api` folder as well, and switched to checking the Authorization header.

PR #318, opened 2021-11-25
[WIP] adds mobile menu navbar

f1lander:

### DIFF
--- a/public/svgs/hamburger.svg
+++ b/public/svgs/hamburger.svg
@@ -1,3 +1,0 @@
-<svg width="28" height="26" viewBox="0 0 28 26" fill="none" xmlns="http://www.w3.org/2000/svg">
-<path d="M2 2H26M2 13H26M2 24H26" stroke="white" stroke-width="3" stroke-linecap="round"/>
-</svg>
--- a/public/svgs/hamburger.svg
+++ b/public/svgs/hamburger.svg
@@ -1,0 +1,3 @@
+<svg width="28" height="26" viewBox="0 0 28 26" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M2 2H26M2 13H26M2 24H26" stroke="white" stroke-width="3" stroke-linecap="round"/>
+</svg>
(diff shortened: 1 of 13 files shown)


public/svgs/hamburger.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="28" height="26" viewBox="0 0 28 26" fill="none" xmlns="http://www.w3.org/2000/svg">
```
exrhizo, line 1: Add these as Icons, see `icons/index.ts`

src/components/MainLayout/MainHeader.tsx
```diff
@@ -206,17 +258,65 @@ export const MainHeader = () => {
       {suspendedButtons}
     </div>
   ) : (
-    <div className={classes.root}>
+    <Box className={classes.root}>
```
exrhizo, line 261: What is the improvement of having Box here?

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -4,14 +4,24 @@ import clsx from 'clsx';
...
+  Hidden,
+} from '@material-ui/core';
 
 import { useMe, useCircle, useGlobalUi } from 'hooks';
 import { useCircleEpochsStatus } from 'recoilState';
 import * as paths from 'routes/paths';
 
 import { ICircle } from 'types';
 
+interface MenuNavigationLinksProps {
```
exrhizo, line 21: I'd inline this - as especially it is only one prop

exrhizo (changes requested):
I really like the gif previews you create for these

src/components/MainLayout/MainHeader.tsx
f1lander: Because we're using MUI allows to us to use that useful props so it's much shorter than create class when I will use different styles for each boxes, instead of create a class that would be adds padding, flex props etc and maybe will be unique for that part its better to use it, maybe in this particular Box I can take away some of the css props to get a better implementation as in the other implementations.

exrhizo: As none of those props are used here - it's more performant and simpler to use HTML.

public/svgs/hamburger.svg
f1lander: Ok, just a question, why we don't use the actual icons from material-icons, for hamburger and close icon? @AlexDistill

exrhizo:
Merged in here: https://github.com/coordinape/coordinape/pull/320

PR #320, opened 2021-12-01
Merge dev 398 dev 821

f1lander:


levity:
I can't scroll down to the last link on the menu (Docs) when it's offscreen -- I can drag the screen down so it becomes visible, but once I let go, it snaps back out of view

image

levity:
Also, please make sure to always have meaningful names for PRs, so they can be found, so feedback can be left in the right place

PR #340, opened 2021-12-04
add fallback initials avatar to Circle in ProfilePage

jsur:
- Added fallback initials avatar image to `ProfilePage` and `AdminCircleModal`.

I didn't find other pages that use `circle.logo` or `.logo` for this purpose, but let me know if there are others.

Screenshot 2021-12-04 at 13 31 51

Screenshot 2021-12-04 at 13 40 50

Screenshot 2021-12-04 at 13 41 25

### DIFF
--- a/src/pages/ProfilePage/ProfilePage.tsx
+++ b/src/pages/ProfilePage/ProfilePage.tsx
@@ -21,7 +21,7 @@ import {
 } from 'recoilState/app';
 import { useSetEditProfileOpen } from 'recoilState/ui';
 import { EXTERNAL_URL_FEEDBACK } from 'routes/paths';
-import { getAvatarPath } from 'utils/domain';
+import { getAvatarPath, getCircleAvatar } from 'utils/domain';
 
 import { IMyProfile, IProfile } from 'types';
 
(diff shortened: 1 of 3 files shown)


src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -317,36 +317,39 @@ const ProfilePageContent = ({
...
-                    alt={u.circle.name}
-                    src={
-                      u.circle?.logo ? getAvatarPath(u.circle?.logo) : undefined
-                    }
-                  >
-                    {u.circle.name}
-                  </Avatar>
+            {profile?.users?.map(
+              u =>
+                u.circle && (
```
jsur, line 322: I replaced a ternary with `&&` to make this simpler.

src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -317,36 +317,39 @@ const ProfilePageContent = ({
...
           </Section>
           <Section title="Recent Epoch Activity" asColumn>
-            {recentEpochs?.map(({ bio, circle }, i) =>
-              circle ? (
-                <div className={classes.recentEpoch} key={i}>
-                  <div className={classes.recentEpochTitle}>
-                    {circle.protocol.name} {circle.name}
+            {recentEpochs?.map(
+              ({ bio, circle }, i) =>
+                circle && (
```
jsur, line 345: Ternary removed here as well.

src/utils/domain.ts
```diff
@@ -61,9 +65,17 @@ export const getAvatarPath = (avatar?: string, placeholder?: string) => {
...
-  const placeholder = name
-    ? `https://ui-avatars.com/api/?name=${encodeURIComponent(name ?? '')}`
-    : AVATAR_PLACEHOLDER;
+  const placeholder = name ? getInitialsUrl(name) : AVATAR_PLACEHOLDER;
 
   return avatar ? getAvatarPath(avatar) : placeholder;
 };
+
+export const getCircleAvatar = ({
+  avatar,
```
jsur, line 74: Which one is better: `avatar` or `logo`?

jsur:
Whoops, forgot this in draft status, changed it.

src/pages/ProfilePage/ProfilePage.tsx
exrhizo: Thank you

src/utils/domain.ts
exrhizo: I could go either way, in the future we can have Hasura do things like this. I like avatar because it is consistent and the user and circle are handled the same way generally.

exrhizo: Maybe we could refactor and use a different name (although I can't think of one that fits both cases well)

exrhizo:
Sweet, Nice and simple improvement - Thank you!

exrhizo (approved):

PR #341, opened 2021-12-04
Change download CSV to use API instead of direct link (#335)

### LINKED ISSUE #335: Use v2 CSV Export endpoint
The v2 CSV endpoint needs to be in `api.ts` instead of a regular link. That way it can use the auth credentials.

image

CryptoGraffe:
Closes #335

### DIFF
--- a/src/hooks/useApiAdminCircle.ts
+++ b/src/hooks/useApiAdminCircle.ts
@@ -86,6 +86,13 @@ export const useApiAdminCircle = (circleId: number) => {
     [circleId]
   );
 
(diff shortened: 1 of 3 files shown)


exrhizo (approved):
Thank you - appreciating you jumping in.

There are a few nit picks that I'd have done differently. 
I'd like to get the project to a place where we have a simple guide for conventions, maybe not as strict as a ruleset but more like guidance.

Anyway, big thank you, this was actually blocking us from improving our access controls, it's the last old endpoint!

PR #344, opened 2021-12-04
Fixes header and sort text that used 'Give' instead of custom token name #342

### LINKED ISSUE #342: /give page doesn't use custom token name consistently
See attached image. At least two instances of "Give" should be "GOVE"
give_vs_gove

CryptoGraffe:
Fixes #342


exrhizo (approved):

exrhizo:
Awesome!

exrhizo (approved):
Got the pinned version - looks good.

PR #346, opened 2021-12-07
Cleanup package.json

exrhizo:
- Move packages to `devDependencies`
 - Remove packages that are not being used.
 - Pin packages to version numbers


exrhizo:
This got borked, so systematicly:
https://github.com/coordinape/coordinape/pull/347

PR #359, opened 2021-12-10
[WIP] poc stitches.dev for styled components

f1lander:


src/components/StitchesButton/Button.tsx
```diff
@@ -0,0 +1,21 @@
+import { styled } from '@stitches/react';
+
+export const Button = styled('button', {
+  // base styles
+  variants: {
+    size: {
+      small: {
+        fontSize: '13px',
+        height: '25px',
```
META-DREAMER, line 9: IMO the best way to use this would be to not hard close any values, but instead define everything in a global theme file and use the aliases here. That way we can enforce typography and spacing scale / consistency and makes it much easier to maintain and refactor in the future

f1lander: Right now this is just a test I already start to config the stitches.config.TS with all the theme tokens as the docs defined

PR #361, opened 2021-12-14
Removed redundant deposit modal

Too-Far:
This PR just changes the fund modal display to the deposit modal to remove redundancy


levity:
If FundModal is no longer used, please remove it entirely

src/pages/VaultsPage/HasVaults.tsx
```diff
@@ -237,11 +235,11 @@ export default function HasVaults({
               variant="contained"
               color="primary"
               size="small"
-              onClick={handleClickfn}
+              onClick={handleClick}
             >
               Fund This Vault
             </Button>
-            <FundModal openfn={openfn} onClose={setOpenfn} />
+            <DepositModal vault={vault} open={open} onClose={setOpen} />
```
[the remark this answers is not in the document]
levity: same as with the other modals in this file; please fix those too

src/pages/VaultsPage/HasVaults.tsx
```diff
@@ -143,17 +142,16 @@ export default function HasVaults({
   const classes = useStyles();
   const [open, setOpen] = useState<boolean>(false);
   const [openwd, setOpenwd] = useState<boolean>(false);
-  const [openfn, setOpenfn] = useState<boolean>(false);
 
   const handleClick = () => {
     setOpen(!open);
   };
   const handleClickwd = () => {
     setOpenwd(!open);
```
levity, line 150: while you're in this file, can you take a look at this? that's a bug, it's not using `openwd`

levity: although, you know, you could argue that only one of these modals should be open at any given time, so you could use a single `useState` for all of them that is set to either a blank string (all modals closed) or a short string e.g. 'deposit' to open the deposit modal

src/pages/VaultsPage/HasVaults.tsx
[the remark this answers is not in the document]
Too-Far: Im not sure that I am following, the onClose just sends a state mutation that changes the bool that controls whether or not the modal is displayed?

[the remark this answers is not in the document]
levity: Ahh, I didn't look at the modal code itself. This works but it's not idiomatic. Ordinarily what you'd do is have `onClose={() => setOpen(false)` here and then `onClose={onClose}` in the component itself. The general rule here is that when you pass callbacks to child components, they should be treated as opaque, so the child can be as reusable as possible. The way it's done here, the modal's own code is coupled with its parent's implementation. Let me know if you have more questions about that.

I think @thenerdcat wrote this originally, so I'll tag him too.

src/pages/VaultsPage/WithdrawModal.tsx
```diff
@@ -37,18 +37,14 @@ const useStyles = makeStyles(theme => ({
...
   user?: IUser;
 }
 
-export default function WithdrawModal({ openwd, onClose }: WithdrawModalProps) {
+export default function WithdrawModal({ open, onClose }: WithdrawModalProps) {
   const classes = useStyles();
   const history = useHistory();
 
-  const handleClose = () => {
-    onClose(false);
```
levity, line 49: @Too-Far @thenerdcat this is what i was referring to in [this comment](https://github.com/coordinape/coordinape/pull/361#discussion_r769772892). the modal shouldn't have to know anything about its callback, but the way this was implemented, it needs to pass one argument and it has to be `false`. this allows the modal to pass some other argument and produce unintended behavior, but more importantly, it also limits reusability--we wouldn't be able to reuse it in different contexts where the parent components have different implementations for `onClose`.

levity (approved):

src/pages/VaultsPage/WithdrawModal.tsx
Too-Far: ahhh ok, I see what you're saying now. Thank you for clarifying!

PR #363, opened 2021-12-14
add warning on opt out, rename DialogNotice to ActionDialog

jsur:
- add warning on opt out
- Remove TODO: rename DialogNotice to ActionDialog

Looks like this, the 0 GIVE value is there just for demo purpose, it will check for > 0 GIVE received.

Screenshot 2021-12-14 at 9 45 59


levity (approved):
Looks good!

PR #365, opened 2021-12-14
added zero address to allow for eth vault creation

Too-Far:
This PR rounds out the fixes for vault creation. You can now mint a vault with any given token. I used a zero address for eth to allow use of same data structure.


levity:
I'm not sure that this is how the Yearn vaults for ETH work, but maybe you want to use the addresses for WETH here?

Too-Far:
Yea I went back and forth with that, I think youre right, and its a safer option anyhow. Ill change and push

levity (approved):

PR #368, opened 2021-12-14
Pattern for catching errors in zod transform

exrhizo:
### Problem,
**If zod safeParseAsync encounters an error,** 
the async doesn't return and the form validation comes back undefined because it hangs in the recoil getter:
`const parsed = useRecoilValueLoadable(rParsed(instanceKey)).valueMaybe();`
is `undefined` and never reloads and doesn't show an error in the console.
( I didn't check if using try catch with regular parseAsync could be better, would require changing error handling)

**So the form shows up not valid, but without any field errors.**

Solution,
A wrapper that creates a refine and then a transform with the successful value.

### DIFF
--- a/src/forms/zTransformCatch.test.ts
+++ b/src/forms/zTransformCatch.test.ts
@@ -1,0 +1,40 @@
+import { z } from 'zod';
+
+import { waitSeconds } from 'utils';
+
+import { zTransformCatch, ZParseError } from './zTransformCatch';
+
+const NEG_ERROR_MESSAGE = 'Negative number';
+
+const schema = z.object({
+  x: z.number(),
+});
+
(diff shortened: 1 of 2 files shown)


levity:
Can you open a discussion issue in the zod repo about this? It seems strange to me that such an... involved... workaround would be necessary, so I'd be curious to know if the maintainer would have a suggestion for how to do this more idiomatically.

exrhizo:
I found a discussion thread in the repo and it hadn't been touched in a while, that is what motivated me to create this PR now, so that I could show the maintainer what crazy thing I had written to try and get this functionality.

https://github.com/colinhacks/zod/discussions/345

exrhizo:
Closing until this and we can revisit if the situation arises again

PR #370, opened 2021-12-16
Add gqty, & use it to fetch the organization's name

levity:
based on #353


.eslintrc
```diff
@@ -43,7 +51,7 @@
     "@typescript-eslint/explicit-module-boundary-types": "off",
     "@typescript-eslint/no-explicit-any": "off",
     "@typescript-eslint/no-empty-function": "off",
-    "@typescript-eslint/ban-ts-ignore": "off",
+    "@typescript-eslint/ban-ts-comment": "off",
```
exrhizo, line 54: whats this do?

src/components/OrganizationHeader/OrganizationHeader.tsx
```diff
@@ -92,10 +92,15 @@ const useStyles = makeStyles(theme => ({
...
+  name?: string;
   buttonText: string;
   onButtonClick: () => void;
 }
-export const OrganizationHeader = ({ buttonText, onButtonClick }: Props) => {
+export const OrganizationHeader = ({
+  name,
+  buttonText,
+  onButtonClick,
+}: Props) => {
```
exrhizo, line 103: in this case, `useSelectedCircle` ought to be passed in from parent as well (perhaps for another PR)

src/services/api.ts
```diff
@@ -43,11 +45,26 @@ export class APIService {
...
-        }
-      : {};
+
+    const auth: Record<string, unknown> = {};
+    if (token) {
+      const authHeader = 'Bearer ' + token;
+      auth.headers = { Authorization: authHeader };
+
+      // add interceptor so GQty requests have proper authorization
+      this.interceptor = hasuraAxios.interceptors.request.use(
```
exrhizo, line 55: Another thing that interceptors can be used for is triggering re-authorization when a token expires

.eslintrc
levity: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/ban-ts-comment.md

the original doesn't seem to exist (maybe it got renamed)

exrhizo (approved):

PR #371, opened 2021-12-16
Store vaults under organization's ID

levity:


src/services/api/hooks.ts
```diff
@@ -0,0 +1,16 @@
+import { useQuery } from 'lib/gqty';
+
+import { useSelectedCircle } from 'recoilState';
+
+export const useCurrentOrg = () => {
```
exrhizo, line 5: It's wierd to me to have hooks in this folder, it's two levels deep - lets have it be in `hooks/` or have another top level for `gqty`

src/services/api/hooks.ts
```diff
@@ -0,0 +1,16 @@
...
+  const { circleId } = useSelectedCircle();
+  const query = useQuery();
+  const currentOrg = query
+    .organizations({ where: { circles: { id: { _eq: circleId } } } })
+    .map(org => ({
+      id: org.id,
```
exrhizo, line 11: As commented in the other PR, this isn't ideomatic gqty

src/recoilState/vaults.ts
```diff
@@ -16,14 +16,19 @@ export const rVaults = atom({
...
-export const useVaults = () => useRecoilValue(rVaults);
+export const useVaults = (orgId: number) => {
+  return useRecoilValue(rVaults)[orgId] || [];
+};
 
 export const useFakeVaultApi = () => {
   const addVault = useRecoilCallback(
     ({ snapshot, set }) =>
-      async (v: IVault) => {
+      async (org: number, v: IVault) => {
```
exrhizo, line 26: nit, I like variable names to like `orgId` because to me org should be the org object

exrhizo:
I think this would be easier to iterate on if you just merged it into the other branch

src/hooks/gqty.ts
```diff
@@ -0,0 +1,16 @@
+import { useQuery } from 'lib/gqty';
+
+import { useSelectedCircle } from 'recoilState';
+
+export const useCurrentOrg = () => {
+  const { circleId } = useSelectedCircle();
```
exrhizo, line 6: `useSelectedCircleId` would be better here, then it doesn't depend on state being loaded in recoil

exrhizo: perhaps another similar recoil atom is added for organization and it can be independent of the selected circle?

src/hooks/gqty.ts
```diff
@@ -0,0 +1,16 @@
...
+  const query = useQuery();
+  const currentOrg = query
+    .organizations({ where: { circles: { id: { _eq: circleId } } } })
+    .map(org => ({
+      id: org.id,
```
exrhizo, line 11: I don't think this is needed, to separate out each, because the whole thing that gqty is doing is only querying what you use

src/hooks/gqty.ts
levity: > perhaps another similar recoil atom is added for organization and it can be independent of the selected circle?

the way the UI is currently, this is what makes the most sense to me

src/hooks/gqty.ts
levity: the hook here is intended to return an actual query result, not a live query that the hook consumers would add their own fields to. it may not look like the "ideal usage" of gqty but i'm trying to find something that gets us moving in the right direction in the context of what we've built already. definitely open to revisiting all this later

PR #373, opened 2021-12-20
hasura add permissions, relationships, superadmin role 

zashton:
1. added superadmin role for profiles with admin_view to query all data.
2. added missing relationships for nominees/vouches
3. update column read permissions for 7 current tracked tables
3. added update permission for profile to allow only own self to update data

should i continue the same for the rest of the untracked tables ?


api/hasura/auth.ts
```diff
@@ -19,9 +19,15 @@ export default async function handler(req, res) {
       },
     });
     assert(tokenRow, 'The token provided was not recognized');
+    const profile = await prisma.profile.findFirst({
```
levity, line 22: Try using a [relation filter](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#relation-filters) to combine this and the above into a single query

META-DREAMER:
Looks good to me! Yea feel free to do this with the rest of the tables we want the front end / API users to consume.

As we start migrating the front end to make mutations / write to Hasura, we will also want to start defining the insert / update / delete permissions for each table (e.g. allow any user in a circle to insert vouches, only allow admin to update certain fields, etc). These should give the minimal necessary permissions, and in the process of doing these you should map it to any existing business logic we have in Laravel and start marking things as deprecated  so we can know once we have feature parity and can deprecate all of laravel

We could start allowing API read access to users own circles if we made a UI to generate an API token by signing a message with their wallet. Basically same flow as our front end but show them the token to use in their own code talking to our GraphQL endpoint.

META-DREAMER (dismissed):
Looks good to me! Yea feel free to do this with the rest of the tables we want the front end / API users to consume.

As we start migrating the front end to make mutations / write to Hasura, we will also want to start defining the insert / update / delete permissions for each table (e.g. allow any user in a circle to insert vouches, only allow admin to update certain fields, etc). These should give the minimal necessary permissions, and in the process of doing these you should map it to any existing business logic we have in Laravel and start marking things as deprecated so we can know once we have feature parity and can deprecate all of laravel

We could start allowing API read access to users own circles if we made a UI to generate an API token by signing a message with their wallet. Basically same flow as our front end but show them the token to use in their own code talking to our GraphQL endpoint.

levity:
> should i continue the same for the rest of the untracked tables ?

sounds good to me!

levity (approved):

PR #376, opened 2021-12-20
[WIP] ft/stitches UI components

f1lander:
### Storybook is not available for now here.
image


src/stitches.config.ts
```diff
@@ -0,0 +1,301 @@
...
+      third: '#EAEAEB',
+      gray: '#EFF3F4',
+
+      lightBorder: '#DEDEDE',
+      lightGray: '#C3CDCF',
+      mediumGray: '#A7B0B4',
+      lightText: '#859296',
+      border: '#939EA1',
+      placeholder: '#99A2A5',
+      placeholderDark: '#7C898D',
```
exrhizo, line 46: This is the sort of thing that we can simplify, maybe it is something like grey400@40%

However we do this, I think it would be best if we didn't stoop to the level of the poor way it's been done in the past.

Ideally there is something like the design tokens so that we can easily do themes in the future.

src/stitches.config.ts
```diff
@@ -0,0 +1,301 @@
...
+      pacificBlue: '#00ABBF',
+      selected: '#31A5AC',
+      lightBlue: '#84C7CA',
+      darkBlue: '#5D9C9F',
+    },
+    space: {
+      px: '1px',
+      '2xs': '0.125rem', // 2px
+      xs: '0.25rem', // 4px
+      '1sm': '0.375rem', // 6px
```
exrhizo, line 62: idk, this is confusing `1sm` vs `sm1`,

Maybe we need to have a style conversation to review this... with a few people talking together

src/stitches.config.ts
```diff
@@ -0,0 +1,301 @@
...
+      '5xl': '5rem', // 80px
+
+      // non standard
+    },
+    sizes: {
+      max: 'max-content',
+      min: 'min-content',
+      full: '100%',
+      px: '1px',
+      '2xs': '0.125rem', // 2px
```
exrhizo, line 82: these should be the same, like `scale = { px: '1px' }` and then `sizes: { ...scale }` and `space: { ...scale }`

src/stitches.config.ts
```diff
@@ -0,0 +1,301 @@
...
+      widest: '0.1em',
+    },
+    borderWidths: {},
+    borderStyles: {},
+    shadows: {},
+    zIndices: {},
+    transitions: {},
+  },
+  media: {
+    xs: MediaQueryKeys.xs,
```
exrhizo, line 162: `... MediaQueryKeys`

src/stitches.config.ts
```diff
@@ -0,0 +1,301 @@
...
+    sm: MediaQueryKeys.sm,
+    md: MediaQueryKeys.md,
+    lg: MediaQueryKeys.lg,
+    motion: '(prefers-reduced-motion)',
+    hover: '(any-hover: hover)',
+    dark: '(prefers-color-scheme: dark)',
+    light: '(prefers-color-scheme: light)',
+  },
+  utils: {
```
exrhizo, line 171: let's seperate things that we expect to be shared between theme variations and things that would be changed, if for instance we allowed users to set primary color ect

src/ui/Box/Box.tsx
```diff
@@ -0,0 +1,8 @@
+import { styled } from '../../stitches.config';
+
+export const Box = styled('div', {
```
exrhizo, line 3: I'm not a fan of having this here, ideally I think we'd have a style reset included as css for the whole app

src/ui/Button/Button.tsx
```diff
@@ -0,0 +1,78 @@
...
+  justifyContent: 'center',
+  height: '$xl2',
+  fontSize: '$md',
+  fontWeight: '$normal',
+  borderRadius: '$xs',
+
+  color: 'White',
+  variants: {
+    color: {
+      red: {
```
exrhizo, line 33: primary rather than red,like in the demo:
image

src/stitches.config.ts
exrhizo: I found a screenshot, I shared for this point below

exrhizo (commented):
I'd like us to have a conversation about all of this with @levity  and maybe @AlexDistill 

I have a lot of opinions, but really we need the group to cohere, not just one person's opinion or another's.

I wonder about organizing our code like metadreamer has here:
https://github.com/meta-cred/core/tree/main/packages/ui/src

src/stitches.config.ts
levity: agreed

src/pages/AllocationPage/AllocationPage.tsx
```diff
@@ -282,10 +282,7 @@ export const AllocationPage = () => {
             />
             <div className={classes.buttonContainer}>
               {epochDirty ? (
-                <Button
-                  className={classes.saveButton}
```
levity, line 286: remember to remove the `saveButton` styles that are now no longer used

src/stitches.config.ts
f1lander: It was in that way, some space are shared in boths

src/ui/Box/Box.tsx
f1lander: This is how stitches recommended to do, but If want we can move it into a global css

src/stitches.config.ts
f1lander: @exrhizo what do you mean by this? which part are you pointing here?

src/stitches.config.ts
f1lander: Sure, I guess first we need to get the rights colors system, for now just leave as we have defined on material-colors

exrhizo (commented):
Let's schedule a meeting to approve this and merge it, I want to talk about it

f1lander:
Sounds perfect!

src/components/MainLayout/MainHeader.tsx
```diff
@@ -314,20 +232,90 @@ export const HeaderNav = () => {
...
+            position: 'relative',
+            '&::after': {
+              content: `" "`,
+              position: 'absolute',
+              left: '50%',
+              right: '50%',
+              backgroundColor: '$mediumRed',
+              transition: 'all 0.3s',
+              bottom: 0,
+              height: '$2xs',
```
levity, line 267: you can remove this token from the theme and just hardcode this value since it's not a layout property
```suggestion
              height: '2px',
```

src/stitches.config.ts
```diff
@@ -0,0 +1,307 @@
...
+    },
+  },
+  defaultVariants: {
+    size: 'lg',
+  },
+};
+
+const spaces = {
+  px: '1px',
+  '2xs': '2px',
```
levity, line 42:
```suggestion
```

src/hooks/useMediaQuery.ts
```diff
@@ -0,0 +1,23 @@
...
+      setMatches(media.matches);
+    }
+    const listener = () => {
+      setMatches(media.matches);
+    };
+    media.addListener(listener);
+    return () => media.removeListener(listener);
+  }, [matches, query]);
+
+  return matches;
```
exrhizo, line 22: This could have `useMemo`

src/ui/IconButton/IconButton.tsx
```diff
@@ -0,0 +1,98 @@
+import { styled } from '../../stitches.config';
+
+export const IconButton = styled('button', {
+  // Reset
```
exrhizo, line 4: In the future lets have a global reset.

exrhizo (approved):
LGTM

levity (approved):

PR #379, opened 2021-12-22
Fix up deposit ability

Too-Far:
Still need to set up a WETH mint function in hardhat in order to test WETH Vaults

I also changed the ETH vaults to WETH


levity:
as noted in the meeting today, we'll want to allow the user to have more control over the amount of allowance they set. but we're not going to do that just yet, since the priority is getting a working "rough draft" of the whole end-to-end flow first, so we can get it in front of more people for internal testing & UX feedback. so i'm just making a note of it here for the future, it doesn't need to block this merge.

src/hooks/useGetAnyTokenValue.ts
```diff
@@ -0,0 +1,36 @@
...
+      const signer = await web3Context.library.getSigner();
+      const token = new ERC20Service(
+        await web3Context.library,
+        await signer.getAddress(),
+        tokenAddress
+      );
+      setBalance(await token.getBalanceOf(account ? account : ''));
+    }
+  };
+  return { getTokenBalance, bal };
```
levity, line 35: rename `bal` to `balance` here for clarity please

src/pages/VaultsPage/DepositModal.tsx
```diff
@@ -35,10 +36,26 @@ export default function DepositModal({
 }) {
   const classes = useStyles();
   const history = useHistory();
+  const [balance, setBalance] = useState<any>();
```
levity, line 39:
```suggestion
  const [max, setMax] = useState<any>();
```
this indicates what the value is used for here better, IMO

levity (changes requested):
looks good, just suggested a few name changes. the logic around decimal places depending on token type is something we'll probably want to extract into a helper at some point, but it doesn't need to block this merge.

levity:
Note there's also some CI error around the number of arguments. (I haven't run this code locally)

src/pages/VaultsPage/AllocateModal.tsx
```diff
@@ -100,8 +99,8 @@ export default function AllocateModal({ open, onClose }: AllocateModalProps) {
       {({ fields, handleSubmit, changedOutput }) => (
         <FormModal
           onClose={onClose}
-          open={open}
           onSubmit={handleSubmit}
+          open={false}
```
levity, line 103: does this prevent the modal from opening?

src/pages/VaultsPage/CreateVaultModal.tsx
```diff
@@ -9,9 +9,9 @@ import CreateVaultForm from 'forms/CreateVaultForm';
 
 export const CreateVaultModal = ({
   onClose,
-  open,
-}: {
-  open?: boolean;
+}: // open,
+{
+  // open?: boolean;
```
levity, line 14: should be safe to remove all this instead of commenting it out?

levity:
hey @Too-Far, this is looking pretty good; check out my changes above. there's one main thing that would be good to get working before we merge this:

- [ ] check for an existing allowance, and skip the allowance tx if it's large enough

other things I noticed that don't have to block the merge:

- [ ] update the vault balance amount automatically after depositing
- [ ] make the "0" placeholder value go away when you start typing
- [ ] ETH/WETH: we're checking ETH balance, but the tx is depositing WETH -- either just show the WETH balance, or add ETH->WETH conversion

levity:
@Too-Far on second thought, i think i'll just merge this now (once checks pass), & you can put new work in a separate branch & PR

levity (approved):

levity:
Issues above are now being tracked here: https://github.com/coordinape/coordinape/issues/457

PR #380, opened 2021-12-22
WIP: fetch epochs from gqty

levity:
this isn't working for some reason--even though the server response does contain epoch data, they don't appear in the component


src/pages/VaultsPage/HasVaults.tsx
```diff
@@ -241,6 +245,32 @@ export default function HasVaults({ epochs, vault }: HasVaultsProps) {
...
+                {c.name}
+                <ul>
+                  {c
+                    .epochs({
+                      // limit: 2,
+                      // where: { ended: { _eq: false } },
+                      order_by: [{ start_date: order_by.asc }],
+                    })
+                    .map(e => {
+                      <li key={e.id}>
```
META-DREAMER, line 264: since on first pass this will be null, you can do this:
```suggestion
                      <li key={e.id || 0}>
```

levity:
we've turned on suspense mode, so i don't think it would ever be null?
would you not recommend using that?

Issue #290, opened 2021-11-13
Skill tags not being saved when updating profile info


AlexDistill:
I went to update my skills tags because I had random ones on my account. Went to edit profile and the tags selected in edit view did not match tags shown on my profile. I removed just one tag from the edit view so the SAVE button would be usable, then I hit save. The metamask signature transaction showed the incorrect data in my signature (it didn't show the tags selected in my edit screen, it showed the same tags that were displaying on the profile

jsur:
`<ProfileSkills />` is set to show max 3 skill tags currently, but all the tags selected are available in state:

<img width="1022" alt="Screenshot 2021-12-12 at 9 25 01" src="https://user-images.githubusercontent.com/1258424/145704159-a4117e5e-127e-40cf-9df5-f1ffede661db.png">

I also didn't get any Metamask actions when saving profile data, so I'm not sure if there's anything to fix here?

Zemm111:
@levity or @exrhizo is this one still open?

levity:
> `<ProfileSkills />` is set to show max 3 skill tags currently, but all the tags selected are available in state

we should probably change this to show all tags on the profile

jsur:
@levity I'll change that, it's only a prop change.

jsur:
Maybe something like this, skill tags are centered with some padding on the sides?

<img width="1275" alt="Screenshot 2021-12-26 at 13 16 46" src="https://user-images.githubusercontent.com/1258424/147406335-92a22611-5442-46dd-ad90-fe62a86b65f4.png">

jsur:
Mobile:

<img width="452" alt="Screenshot 2021-12-26 at 13 17 50" src="https://user-images.githubusercontent.com/1258424/147406349-e8f6b372-d6ce-4295-9a3a-aaff4b79e8c1.png">

jsur:
https://github.com/coordinape/coordinape/pull/382

PR #382, opened 2021-12-26
show all skill tags in ProfilePage

jsur:
- Change skill tag max count to 50 in ProfilePage, adjust styles


jsur:
As discussed in https://github.com/coordinape/coordinape/issues/290

exrhizo:
Sweet!

exrhizo (approved):

PR #383, opened 2021-12-28
Add read permissions for remaining untracked tables

zashton:
Tables Added to Hasura

- burns
- circle_metadata
- histories
- pending_token_gifts
- token_gifts
- personal_access_tokens
- teammates

Views added

- gift_private
- pending_gift_private
- circle_private

@META-DREAMER 
theres no direct way in hasura to customise permissions for specific columns, so i created views and added permissions on them and attached to the parent table as a relationship. (Restricting notes visibility to only those you have sent/received and discord webhook for circles to circle admin only)


hasura/metadata/databases/staging/tables/public_circles.yaml
```diff
@@ -79,25 +116,3 @@ select_permissions:
...
-    - only_giver_vouch
-    - team_sel_text
-    - team_selection
-    - token_name
-    - vouching
-    - vouching_text
-    filter:
-      id:
-        _in: X-Hasura-Admin-Circle-Ids
-  role: user
```
META-DREAMER, line 103: Any reason this was removed?

zashton: From my understanding, hasura create/update with data validation/business logic are triggered in action handlers using admin secret. So i removed the permission for the role, or is it still needed ?

hasura/metadata/databases/staging/tables/public_pending_gift_private.yaml
```diff
@@ -0,0 +1,23 @@
...
+    - recipient_id
+    - sender_id
+    - note
+    filter:
+      _or:
+      - recipient_id:
+          _in: X-Hasura-Circle-User-Ids
+      - sender_id:
+          _in: X-Hasura-Circle-User-Ids
+  role: user
```
META-DREAMER, line 23: Lets say we want to allow circle admins to see `recipient_id` and `sender_id` but not the `note`. That would be a use case where we want to setup a separate `circle-admin` role (vs `id:  _in: X-Hasura-Admin-Circle-Ids`).

zashton: we want to only allow the recipient / sender of the note to be able to view the note. circle admins are also restricted to this same condition. Is there a different way to accomplish this compared to view ?

hasura/metadata/databases/staging/tables/public_pending_token_gifts.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - sender_address
+    - sender_id
+    - tokens
+    - updated_at
+    filter:
+      circle:
+        users:
+          profile:
+            id:
+              _eq: X-Hasura-User-Id
```
META-DREAMER, line 76: Are pending gifts supposed to be visible by all members of the circle? If not then this might need to be changed, but if so its fine.

hasura/migrations/staging/1640679052192_custom_views/up.sql
```diff
@@ -0,0 +1,13 @@
+
+
+CREATE VIEW public.circle_private AS
+  SELECT id AS circle_id, discord_webhook
+    FROM circles;
+
+CREATE VIEW public.gift_private AS
+ SELECT recipient_id, sender_id, note
+   FROM token_gifts;
```
META-DREAMER, line 9: Do we only need these columns? what about the amount of tokens?

hasura/metadata/databases/staging/tables/public_pending_token_gifts.yaml
zashton: Yes they can see so the map of the graph can be constructed. Pending gifts are all the gifts sent in the current active epoch. Only the notes are restricted hence the private view

api/hasura/auth.ts
```diff
@@ -26,9 +26,26 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
...
     res.status(200).json({
       'X-Hasura-User-Id': tokenRow.tokenable_id.toString(),
       'X-Hasura-Role': profile.admin_view ? 'superadmin' : 'user',
+      'X-Hasura-Circle-User-Ids': `{ ${users
+        .map(u => u.id.toString())
+        .join()} }`,
+      'X-Hasura-Admin-Circle-Ids': `{ ${users
+        .filter(u => u.role === 1)
+        .map(u => u.circle_id.toString())
+        .join()} }`,
```
META-DREAMER, line 48: This is one way to do it, but actually not necessary since all this info is available in the same database. [See here](https://hasura.io/docs/latest/graphql/core/auth/authorization/roles-variables.html#relationship-information-is-available-in-the-same-database)

This is fine for now as it might make it simpler / cleaner to define these permissions in the console instead of having to setup the complicated rule each time

META-DREAMER (approved):
few comments and questions but looks good to me overall, feel free to merge once addressed.

hasura/migrations/staging/1640679052192_custom_views/up.sql
zashton: so in the pending_gifts query you can access the amount of tokens, for the gift_private view i've added custom permissions to check if you are the sender/recipient in order to access this view to privatise the note field

hasura/metadata/databases/staging/tables/public_circles.yaml
exrhizo: As much as possible, we will want to write graphql mutations to update state. Data validation and business logic can be done either in the front end or with constraints in postgres.

zashton: i'm not sure how would it work if data validation and business logic is on the frontend, if someone fork the frontend and remove the business logic and data validation could they bypass it ? Constraints for DB usually is used for ensuring data integrity rather than data validation. Maybe you could try (in this case one of the things you will need to check is for vouching and close all the nominations if its off), idk whether it will work

levity: > we will want to write graphql mutations to update state

@exrhizo yes, and, a hasura graphql mutation can call a webhook. we will want to have the business logic behind the mutation, not in the frontend, so that the same API can be used by multiple clients. e.g. in the case of vouching for a nominee, this could be done through a Discord bot in the future. you want a client to be able to simply call e.g. `addVouch` rather than need to directly manipulate our schema.

> hasura create/update with data validation/business logic are triggered in action handlers using admin secret. So i removed the permission for the role, or is it still needed ?

@zashton in this case, at least some of these are still needed, since circle admins can change these values from the circle admin settings page. when there isn't complex business logic needed, we should just do a simple update directly.

for slightly more complex validations we can use [check constraints](https://www.dropbox.com/s/5q6qx81qrgntgkg/Screen%20Shot%202022-01-06%20at%2010.10.21%20AM.png?dl=0) ([docs](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS)) in the hasura console.

zashton: @levity which endpoints do you think can be a simple update with no business logic ?

levity: > @levity which endpoints do you think can be a simple update with no business logic ?

circle admin settings is the first one that comes to mind... oh, i see what you're saying now, changing `min_vouches` has an effect on existing nominees. hmm, in this case, maybe we can split that out into a separate action.

api/hasura/auth.ts
levity: > it might make it simpler / cleaner to define these permissions in the console instead of having to setup the complicated rule each time

+1 -- more performant too, unless Hasura is really smart about recognizing & caching common subqueries

hasura/metadata/databases/staging/tables/public_circles.yaml
zashton: shouldn't it be together in one action handler, otherwise if 1 action succeeds and the other fails there will be data integrity issues. How does the client parse the constraint errors that the DB throws and display a meaningful error message for the user ?

levity: In this case, IMO, it's not really a data integrity issue if min_vouches is changed after some nominee already exists. This is more of a product question. Someone might want to change the setting for the future but not affect current nominees. So having this in two steps is not a big deal IMO.

That said, if you still want to do it in one step, using a webhook, I'm ok with that too.

PR #384, opened 2021-12-29
Defunct - Proposal for local dev environment

exrhizo:
Requires: Hasura cli version >= 2.1.1

This is a proposed setup for Hasura local development w/ all sub systems

### Goals:
 - Laravel manages the DB migrations
   - Easily create larvel migrations and develop in Hasura concurrently
 - `setup.sh` + `docker-compose up` + `yarn run seed-db 0x00ME00` + `vercel dev`
   -  **Run everything needed** for all our subsystems to run **locally**, including auth connecting & laravel
 - Enable integration testing in CI
 - Enable contributors to easily work on big circles (100+)

### Notice
 - Using hasura database name "default" is preferable
   - w/ env variable `PG_DATABASE_URL` this is the default initialization
   - It will be consistent and semantically correct naming in Prod & Staging & Local 

### TODO:
 - Merge latest Hasura metadata
 - Make sure all the services talk to each other correctly, and update .env.example
 - Rewrite documentation for dev setup
   - Include documentation for developing with staging only
 - Create a vercel function (gqty or prisma?) to initialize circles in a parametrizable way
   - Perhaps available in prod/staging w/ super admin credentials 
 - Make sure it works for other devs (non M1 setups)
 
 
 ### Out of scope but seems doable
 - AWS S3 emulation w/ docker compose
 - hardhat in docker compose
   - (containerize mainnet fork??)


zashton:
whats the reason you are still using laravel for migrations ?

exrhizo:
Use larvel for migrations so that there is a single source of truth for the database state.

As I think about it, I could imagine that hasura is the source of truth for migrations, but then we would want to make sure that it doesn't break Laravel. 

I notice with this setup, hasura will fail to start if the postgres state doesn't match the expected metadata schema.

I think it can go either way, maybe we can talk as a team, I'd like to be able to setup a process that is clear, so everyone has the same mental model for the migration.

levity:
Being able to use Hasura as the single source of truth ASAP would be great.

zashton:
The eloquent model generator is just for ORM at the application layer doesn't modify the DB, unless we still intend to still develop on Laravel to continue build API on it this might not be useful at all.

My understanding this docker change is just a temp solution for setting up local env and we remove laravel totally after we have hasura ready. If yes we should have migration done at hasura, otherwise for the new DB changes you will need to "migrate" the new migration files from laravel to hasura when we phase laravel out which is additional work. Also I've already made DB changes to add views in hasura for user permissions.

exrhizo:
@zashton , right.
What I had imagined was that it would be possible to regenerate the models in postgres from the DB but it seems I was overcomplicating things.

I experimented, and laravel is fine with extra fields and tables being added, in that case it is straightforward for Hasura to be the place where migrations are done. And we can just wait until after we switch completely before we remove anything.

Note: When I added a field to users, laravel automatically returned it.

exrhizo:
Superseded by: https://github.com/coordinape/coordinape/pull/385

PR #405, opened 2022-01-10
Replace Material-UI SvgIcon with Stiches.dev SvgIcon

f1lander:
### Summary
Since we are implementing stitches.dev we want to replace for first steps much as we can from material-ui to our new design system convention.

### Goal
Replace the `import { SvgIcon } from '@material-ui/core';` component for ours `import { SvgIcon } from './SvgIcon';`

image

Example for usage:
image

As you see the naming is the same to leave almost the same, the other change is every `SvgIcon` into `styled` from our `stitches.config` to enable the different variants defined.

image

Example of usage:


package.json
```diff
@@ -20,6 +20,7 @@
     "@sentry/integrations": "6.14.1",
     "@sentry/react": "6.14.1",
     "@sentry/tracing": "6.14.1",
+    "@stitches/react": "^1.2.6",
```
exrhizo, line 23: pin `yarn add lib --exact`

src/icons/MinusCircleIcon.tsx
```diff
@@ -1,23 +1,35 @@
...
-    />
-  </SvgIcon>
+import { SvgIcon } from './SvgIcon';
+
+import { IconProps } from 'types';
+
+export const MinusCircleIcon = styled(
+  ({ color = 'currentColor', ...props }: IconProps) => {
+    return (
+      <SvgIcon {...props}>
```
exrhizo, line 13: I still think it is possible to remove a couple lines of boiler plate, I don't want to ask you to spend so much time on this..

whatif it was like: 
```
export const MinusCircleIcon = (props) =>
  <SvgIcon {...props}>
    ({ color }) => (
       <path d="M12 10 2 1" stroke={color} />
     )
   </SvgIcon>
   ```
   I think needing color is why you kept those parts of the interface out?

f1lander: I need to wrap into the `styled`

exrhizo:
image
is now black

image
`i` is bigger

image
I think pencil is bigger

exrhizo:
Given that icons are breaking in the app:
Lets do this incrementally, revert the changes for most of the icons, and create a folder in `ui/icons` and then put the new icons there.

Finally this is the interface I was suggesting, it might be "too clever"
the benefit is that we control want props the final icon accepts and `(color: string)` could be set by stitches, and we can even have something like `({color, colorBackground})` and have stitches variants set those.

https://gist.github.com/exrhizo/fc689bcc8df9656da19d3c9ce5fbadc6
I think this isn't exactly right, because I forgot to reinclude width and height.

I felt like all Svg props were overkill, but idk, the main thing is that creating icons is simple and using them is simple.

PR #419, opened 2022-01-17
Prep for laravel docker-compose. Add Git Submodule.

exrhizo:
Add some structure around how env vars are done.

Add laravel submodule.

vscode 2 space, wasn't defaulting for me.. So put it in the local config, it's our convention

Pin dependencies


src/lib/README
```diff
@@ -1,2 +1,4 @@
+# libraries
```
levity, line 1: Rename file to .md if you want to use markdown

src/config/env.ts
```diff
@@ -0,0 +1,44 @@
...
+  : '10000000-ffff-ffff-ffff-000000000001';
+export const INFURA_PROJECT_ID = getEnvValue(
+  'REACT_APP_INFURA_PROJECT_ID',
+  'missing-infura-id'
+);
+export const REACT_APP_HASURA_URL = getEnvValue(
+  'REACT_APP_HASURA_URL',
+  'https://missing-hasura-url.edu'
+);
+export const HASURA_ADMIN_SECRET = process.env.REACT_APP_HASURA_ADMIN_SECRET;
```
levity, line 38: I think we need to avoid putting this in the frontend build, no?

levity (changes requested):

src/lib/README
exrhizo: done

src/config/env.ts
exrhizo: done

levity (approved):

PR #421, opened 2022-01-18
Setup Hasura Cloud Preview with Github Actions

### LINKED ISSUE #400: Branch specific Hasura instances to work with Vercel branch previews
We currently (before Hasura) use Vercel to automatically deploy previews of every branch. This enables easy testing on staging and CI/CD workflows. See https://vercel.com/features/previews

This task is to setup a similar feature for Hasura - https://hasura.io/docs/latest/graphql/cloud/preview-apps.html - where Hasura will create a temporary schema on the same Postgres database instance to allow us to test and preview database schema changes.

With this setup, we should be able to have multiple virtual Postgrees databases running on the same postgres server.

The ideal end result is that for every PR, we have a isolated database and isolated preview that allows for easy development.

META-DREAMER:
closes #400 

This enables [Preview Apps](https://hasura.io/docs/latest/graphql/cloud/preview-apps.html) for PRs made to this repo that modify the Hasura state. It will create an ephemeral database in the staging Postgres server and try applying any new metadata / migrations to it in order to verify correctness before merging the PR.

Since a PR may also make frontend changes that depend on changes to the Hasura instance, the Github action will also trigger a Vercel deployment that has the ephemeral Hasura instance configured as the GraphQL endpoint using the `REACT_APP_HASURA_URL` env var.

Once the PR is merged, the `delete-hasura-preview.yml` action will cleanup the ephemeral database and Hasura instance since it's no longer needed.

### DIFF
--- a/.github/workflows/delete-hasura-preview.yml
+++ b/.github/workflows/delete-hasura-preview.yml
@@ -1,0 +1,21 @@
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  delete-hasura-preview:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Hasura Cloud Preview Apps
+        uses: hasura/hasura-cloud-preview-apps@v0.1.4
+        with:
+          name: "coordinape-${{github.env.GITHUB_HEAD_REF}}${{github.event.number}}"
+          delete: true
+          postgresDBConfig: |
+            POSTGRES_SERVER_CONNECTION_URI=${{secrets.POSTGRES_STAGING_CONNECTION_URI}}
+            PG_ENV_VARS_FOR_HASURA=PG_DATABASE_URL
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          HASURA_CLOUD_ACCESS_TOKEN: ${{secrets.HASURA_CLOUD_ACCESS_TOKEN}}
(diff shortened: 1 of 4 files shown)


crabsinger (approved):
@META-DREAMER this looks great! Nice work on the workflows and getting it all working. I'm not super sure how to test this, but the code looks good to me.
You said:
> The downside is that there might be a race condition where the env var is not set before the vercel build reads them, which can be addressed by just redeploying manually from vercel dashboard.

So we should keep an eye out for this behavior post merge. And if we're seeing vercel deploys which fail, and rebuilding works, then we're likely in this scenario correct?

I guess I'm curious if we'll see a specific error message ("Missing X env var") if we're in a race condition? Is that right?

META-DREAMER:
@crabsinger The build wont fail, it will just be pointing to the staging Hasura instance instead of the temporary one for the PR. In practice this wont really matter much.

The only time this would be an issue is if the same PR makes changes to the Hasura tables and also implements new frontend features that depend on those changes AND someone actually goes to the Vercel preview deployment to test it. The backend API url will be logged to the browser console so you'l be able to see there if it's pointing to the staging or temp instance. If its not pointing to temp instance you can redeploy from vercel and it should pick up the new env variable.

In terms of testing if this works, you can see a [passing deployment run here ](https://github.com/coordinape/coordinape/runs/4872928155)where it deployed the Hasura instance and [commented on the PR](https://github.com/coordinape/coordinape/pull/421#issuecomment-1016800070)  with the API url and console link.

PR #398, opened 2022-01-08
docker-compose w/ Laravel

exrhizo:
Superseding: https://github.com/coordinape/coordinape/pull/385
Merge w Laravel migration disable: https://github.com/coordinape/coordinape-backend/pull/74

Requires: Hasura cli version >= 2.1.1

Setup for Hasura local development w/ all sub systems

### Goals:
 - Hasura manages the DB migrations
 - `setup.sh` + `docker-compose up` + `yarn run seed` + `vercel dev`
   -  **Run everything needed** for all our subsystems to run **locally**, including auth connecting & laravel
 - Enable integration testing in CI
 - Enable contributors to easily work on big circles (100+)

### Notice
 - Using hasura database name "default" is preferable
   - w/ env variable `PG_DATABASE_URL` this is the default initialization
   - It will be consistent and semantically correct naming in Prod & Staging & Local 
   - see: [Setting up Hasura migrations](https://hasura.io/docs/latest/graphql/core/migrations/migrations-setup.html) Step 3.1: Initialize database migrations

### TODO:
 - Fix FE crashes because of null values in test data
 - Finish rewriting documentation
   - Include documentation for developing with staging only
 - Make sure it works for other devs (non M1 setups)
 
 
 ### Out of scope but seems doable
 - AWS S3 emulation w/ docker compose
 - hardhat in docker compose
   - (containerize mainnet fork??)

PR #423, opened 2022-01-18
Feature/create circle hasura

### LINKED ISSUE #411: Mutations/Permissions/Actions handler/Logic + Data validation for User Circle Creation
Action Handler
- Create circle
- if protocol_id is provided link to existing organisation else create a new organisation
- Add membership of circle creator into the new circle as a circle admin
- Add Coordinape user into circle to receive donations

Validation
- If protocol_id is provided, check if circle creator is an existing admin of another circle in the organisation.

Permissions
- Users cannot directly trigger a direct insert mutation on the circle data

zashton:
Fixes #411 

@META-DREAMER


api/hasura/actions/create_circle.ts
```diff
@@ -0,0 +1,221 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+import { GraphQLClient, gql } from 'graphql-request';
+import Validator from 'validatorjs';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
```
META-DREAMER, line 5: We should strongly type these using Hasura's codegen features: https://hasura.io/docs/latest/graphql/core/actions/codegen/index.html

api/hasura/actions/create_circle.ts
```diff
@@ -0,0 +1,221 @@
...
+    function (value) {
+      try {
+        JSON.parse(value);
+      } catch (e) {
+        return false;
+      }
+      return true;
+    },
+    'The :attribute is not a valid JSON.'
+  );
```
META-DREAMER, line 27: Would be cleaner to use a library like Zod so we can get proper integration with TS for type guarding and more reusable parsers / validators from the composability.

api/hasura/actions/create_circle.ts
```diff
@@ -0,0 +1,221 @@
...
+            }
+          ) {
+            id
+          }
+        }
+      `;
+      let data = await client.request(QUERY, {
+        address: variables.address,
+        protocol_id: variables.protocol_id,
+      });
```
META-DREAMER, line 63: We should just use GraphQL Zeus here as well so we get strong type safety etc. Can generate an "admin zeus client" separately from the zeus client generated for clientside so it only allows using queries and fields that are allowed for the given role.

hasura/metadata/actions.yaml
```diff
@@ -1,6 +1,15 @@
-actions: []
+actions:
+- name: create_circle
+  definition:
+    kind: synchronous
+    handler: http://host.docker.internal:3000/api/hasura/actions/create_circle
```
META-DREAMER, line 5: The handler base URL should be defined with an env variable

META-DREAMER (changes requested):
good stuff, just some suggestions on how to improve the typesafety and make dev experience a lot nicer / seamless

api/hasura/actions/create_circle.ts
zashton: Would i need to switch to node express framework to use hasura's codegen ?

exrhizo: This PR has the relevant mutations w/ zeus generated code
https://github.com/coordinape/coordinape/pull/398

But I am splitting it up into a few PRs

api/hasura/actions/create_circle.ts
exrhizo: Yes, using zod is preferred, we can share validation code between the front end and backend.

For that, we can have a folder in `lib` for zod schemas, and actually there is one for this purpose in `src/forms/CreateCircleForm.tsx`

you might need to split it into parts. Also we can remove uxresearch_json, there is a ticket to remove those fields.

package.json
```diff
@@ -33,6 +33,7 @@
     "ethers": "5.5.1",
     "gqty": "^2.1.0",
     "graphql": "^16.0.1",
+    "graphql-request": "^3.7.0",
```
exrhizo, line 36: pin dependencies `--exact`
e.g. `"3.7.0"`

exrhizo (commented):
I wonder if this could all be done without an action,
[Postgres: Data validations](https://hasura.io/docs/latest/graphql/core/databases/postgres/schema/data-validations.html)
The hard thing would be to do the check for if a user had permission to add a circle to an existing protocol.

I don't think we have a way to express the permission for org admin?


If using a hook:
After using both Zues and prisma, for this sort of thing, I feel like prisma might be easier, especially as it isn't so complicated as to involve multiple tables. It would definitely be simpler.

api/hasura/actions/create_circle.ts
exrhizo: by split into parts I mean:
```
const shared = {   user_name: z.string(), circle_name: z.string() };
const schema = { ...shared, captcha_token: z.string() }
```

api/hasura/actions/create_circle.ts
META-DREAMER: Hasura has a "typescript-zeit" generator that you can use for vercel style handler. Main thing to copy is the type definitions, ideally we can [create our own code generator](https://github.com/hasura/codegen-assets/blob/master/builder-kit/README.md) that does it in the exact right format we want.

zashton:
- added zod for validation
- used hasura codegen to type inputs
- integrated gql zeus for query/mutation

api/hasura/customTypes/createCircleCustomTypes.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export type Create_Circle_Input = {
+  user_name: Scalars['String'];
+  address: Scalars['String'];
+  circle_name: Scalars['String'];
+  protocol_name?: Maybe<Scalars['String']>;
+  protocol_id?: Maybe<Scalars['Int']>;
+};
+
+export type Create_Circle_Response = {
```
exrhizo, line 22: If you run `yarn generate`, I hope these get created automagically by zeus + hasura introspection.

We don't want to manually manage these types.

api/hasura/actions/create_circle.ts
```diff
@@ -0,0 +1,173 @@
...
+            returning: {
+              id: true,
+            },
+          },
+        ],
+      });
+      return res.status(200).json(insert_circles_one);
+    } else {
+      // create a new organisation
+      const { insert_organizations_one } = await gql.q('mutation')({
```
exrhizo, line 133: We will want to move these into a library file, like `src/lib/gql/GqlHasuraAdmin.ts`

api/hasura/actions/create_circle.ts
```diff
@@ -0,0 +1,173 @@
...
+  } catch (err) {
+    if (err instanceof z.ZodError) {
+      return res.status(422).json({
+        extensions: err.issues,
+        message: 'Invalid input',
+        code: '422',
+      });
+    }
+  }
+  const address = object.address.toLowerCase();
```
exrhizo, line 26: The zod parser should probably be rewritten so that the address is included and that lower case is also done with it. I can help with that.

.env.example
```diff
@@ -3,6 +3,7 @@ LOCAL_SEED_ADDRESS=0xYOUR_DEV_ADDRESS
 HARDHAT_OWNER_ADDRESS=0xYOUR_DEV_ADDRESS
 REACT_APP_INFURA_PROJECT_ID=infuraidhere
 ETHEREUM_RPC_URL=https://mainnet.infura.io/v3/infuraidhere
+COORDINAPE_USER_ADDRESS=0xfad763da9051953fea58c2304395719a3b6ba361
```
exrhizo, line 6: Is this address for staging or prod or local?

exrhizo (changes requested):

src/lib/zod/index.ts
```diff
@@ -0,0 +1,16 @@
+import { z } from 'zod';
+
+export const createCircleSchemaInput = z
+  .object({
+    user_name: z.string().min(3).max(255),
+    address: z.string().regex(/^(0x)?[0-9a-f]{40}$/i),
```
exrhizo, line 6: we have a address checker in `src/forms/formHelpers.ts`

package.json
```diff
@@ -36,6 +36,8 @@
     "gqty": "2.1.0",
     "graphql": "16.2.0",
     "graphql-zeus": "4.0.4",
+    "graphql-request": "3.7.0",
+    "history": "5.0.0",
```
exrhizo, line 40: These are no longer needed?

api/hasura/customTypes/createCircleCustomTypes.ts
zashton: it did create a`ValueTypes['create_circle_input']` type have removed this since its unneeded
But the fetch error is somehow back after i updated from main, can't query/mutate anymore from backend

.env.example
zashton: this is for prod, can be the same address for staging as well

src/lib/zod/index.ts
zashton: tried importing, backend seem to have no access to config/env

Cannot find module 'config/env'
Require stack:
- /yearn-coordinape/src/forms/formHelpers.ts
- /yearn-coordinape/src/lib/zod/index.ts
- /yearn-coordinape/api/hasura/actions/create_circle.ts
- /.config/yarn/global/node_modules/@vercel/node/dist/launcher.js
- /.config/yarn/global/node_modules/@vercel/node/dist/dev-server.js

levity: try using full relative paths (i.e. '../config/env' or whatever it should be)
it's only Webpack that makes "magic" paths like 'config/env' work, and ofc it doesn't run in the backend

api/hasura/customTypes/createCircleCustomTypes.ts
topocount: This is amazing

topocount (commented):
How does this run without `node-fetch`? I get fetch is not defined errors in the handler I'm writing without it.

zashton:
> How does this run without `node-fetch`? I get fetch is not defined errors in the handler I'm writing without it.

for now it should work if you add
`import './node-fetch-shim';`
into `api-lib/Gql.ts`

api/hasura/actions/create_circle.ts
zashton: Done

api/hasura/actions/create_circle.ts
zashton: i've left the tolowercase for now

exrhizo (approved):

PR #426, opened 2022-01-18
Rename Hasura DB from staging to default

exrhizo:
### Using hasura database name "default" is preferable
   - w/ env variable `HASURA_GRAPHQL_DATABASE_URL` this is the default initialization
   - It will be consistent and semantically correct naming in Prod & Staging & Local 
   - see: [Setting up Hasura migrations](https://hasura.io/docs/latest/graphql/core/migrations/migrations-setup.html) Step 3.1: Initialize database migrations


crabsinger (approved):
This looks good to me. Not sure if this will break our Hasura cloud instance with the rename, but if it breaks we'll know what to fix.

PR #432, opened 2022-01-19
feat: add remaining methods in apeDistributor hook

Niraj-Kamdar:


exrhizo (dismissed):
@levity @teeolendo 
I am not tracking if these interfaces make sense, I wonder if there is something this needs to be merged?

Or do we close it?

levity:
This is something that we need, but I haven’t merged it yet because it uses
a pattern that turned out to be unnecessary, so I wanted to modify it. Just
haven’t gotten around to it yet.

Issue #389, opened 2022-01-03
Edit background image doesn't work when choosing same image twice in a row


spydermonkee:
**Problem**
Background doesn't show preview on second try if same image was selected twice. (without saving by using the "Upload Background" button).

1. choose edit background button
2. choose file
3. hit "x" instead of "upload background"
4. edit background button
5. choose same file again

RESULT - nothing happens (no background change and no info)
EXPECTED - image preview on background 

Note - refresh allows same image to work again

**Solution**
Allow multiple tries with same image

**Stakeholders**
Dev team

CryptoGraffe:
i can check this out

PR #434, opened 2022-01-20
Fixing bug #389 - choosing the same background image twice in a row

CryptoGraffe:
Fixes bug #389 where choosing the same background image twice in a row doesn't work. This also fixes same for avatar and other future usages of `FormFileUpload`


exrhizo (approved):
LGTM! Thank you :)

CryptoGraffe:
I don't have perms to merge so someone else can merge this

PR #453, opened 2022-01-22
Configure Webpack to create Sentry release and tag it in the Frontend

### LINKED ISSUE #413: Have Sentry know about git commit
There is a way to get sentry set up so that when there is a deploy and new bugs start happening, sentry will automatically detect the likely commit.

A lead is that vercel has this env variable:
VERCEL_GIT_COMMIT_SHA

It may involve something with the deploy hooks as well.

topocount:
Along with enabling integrations in GitHub and Vercel, this change
allows for commits to be associated with deploys in sentry.io.

In local environments, including `vercel dev`, the webpack plugin
defaults to "dry run" mode and doesn't attempt to create a release or
associate commits in Sentry. Commits are only tied to a release in
the specified config when vercel deploys from the "production" branch
in this configuration.

Additional configuration is below.

Closes #413 

test plan:

The "production" branch head in https://github.com/topocount/coordinape
is currently throwing an error on pageload. That error is associated
with the tagged release (represented by the commit SHA) which is
associated with recent commits.

## Vercel Integration with Sentry

[These instructions](https://docs.sentry.io/product/integrations/deployment/vercel/) up to but excluding "Usage" will suffice. The required webpack config is in this PR.

## Github Integration with Sentry

- [x] Steps 1-3 in [these instructions](https://docs.sentry.io/product/releases/setup/release-automation/) are all that's necessary. Since we're deploying with Vercel and have this webpack config, a Github Action is not needed.

### DIFF
--- a/craco.config.js
+++ b/craco.config.js
@@ -35,7 +35,7 @@ module.exports = {
           dryRun: process.env.VERCEL_GIT_COMMIT_SHA ? false : true,
           ignore: ['node_modules', 'craco.config.js'],
           org: 'coordinape',
-          project: 'coordinape',
+          project: 'app',
         }),
       ],
     },
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,4 +1,5 @@
 const webpack = require('webpack');
+const SentryCliPlugin = require('@sentry/webpack-plugin');
 
 module.exports = {
   webpack: {
@@ -26,7 +27,17 @@ module.exports = {
           buffer: require.resolve('buffer/'),
         },
       },
-      plugins: [new webpack.ProvidePlugin({ Buffer: ['buffer', 'Buffer'] })],
+      plugins: [
+        new webpack.ProvidePlugin({ Buffer: ['buffer', 'Buffer'] }),
+        new SentryCliPlugin({
+          include: 'build',
+          release: process.env.VERCEL_GIT_COMMIT_SHA,
+          dryRun: process.env.VERCEL_GIT_COMMIT_SHA ? false : true,
+          ignore: ['node_modules', 'craco.config.js'],
+          org: 'coordinape',
+          project: 'coordinape',
+        }),
+      ],
     },
   },
 };
--- a/craco.config.js
+++ b/craco.config.js
@@ -1,6 +1,11 @@
 const webpack = require('webpack');
 const SentryCliPlugin = require('@sentry/webpack-plugin');
 
+const { VERCEL_GIT_COMMIT_SHA, SENTRY_PROJECT, SENTRY_AUTH_TOKEN } =
+  process.env;
+
+const shouldDryRun = !(SENTRY_PROJECT && SENTRY_AUTH_TOKEN);
+
 module.exports = {
   webpack: {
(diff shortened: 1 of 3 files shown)


craco.config.js
```diff
@@ -26,7 +27,17 @@ module.exports = {
...
-      plugins: [new webpack.ProvidePlugin({ Buffer: ['buffer', 'Buffer'] })],
+      plugins: [
+        new webpack.ProvidePlugin({ Buffer: ['buffer', 'Buffer'] }),
+        new SentryCliPlugin({
+          include: 'build',
+          release: process.env.VERCEL_GIT_COMMIT_SHA,
+          dryRun: process.env.VERCEL_GIT_COMMIT_SHA ? false : true,
+          ignore: ['node_modules', 'craco.config.js'],
+          org: 'coordinape',
+          project: 'coordinape',
```
topocount, line 38: These fields might need to be modified depending on the `org` and `project` slugs are configured in Sentry

package.json
```diff
@@ -121,6 +121,7 @@
     "@datamodels/identity-profile-basic": "0.1.2",
     "@glazed/types": "0.1.3",
     "@gqty/cli": "2.3.1",
+    "@sentry/webpack-plugin": "^1.18.4",
```
exrhizo, line 124: pin dependencies `--exact`

craco.config.js
exrhizo: org is `coordinape`
project is `app`

exrhizo (commented):
So then,

this isn't needed?

image


I think @crabsinger added the sentry integration to vercel, can we merge this and it will work?

@levity @crabsinger probably still need to do the github setup?

craco.config.js
levity, line 38:
```suggestion
          project: 'app',
```

levity (approved):
LGTM once you make the dependency change that @exrhizo suggested (personally i don't mind non-exact deps but i guess we should be consistent)

topocount:
> So then,
> 
> image this isn't needed? (because you did the webpack version of feeding these in?)

This is still needed because this step will generate a project-scoped auth token that is graanted to vercel. I think this is much cleaner than any one of the contributors or maintainers generating the auth token via the [integration platform](https://docs.sentry.io/product/integrations/integration-platform/), essentially a manual integration.

> # Next steps
> 
> What is still needed for merge?

It seems like this has all been done

- [ ] The next steps are just the config documentation I linked in the PR description. The Vercel-Sentry setup will be needed prior to merging this PR or Vercel builds will likely fail, since this is setup to attempt to communicate with Sentry when a `VERCEL_GIT_COMMIT_SHA` environment variable is present.

- [ ] The Sentry-Github integration will be needed for associated commits to begin showing up for each Release.

It looks like the mandatory vercel-Sentry config might be done to prevent breaking vercel deployments, but I've reified this list of todos just to be sure.

crabsinger:
I'll work with @topocount to setup the configs on Sentry and Vercel.

crabsinger:
Added the Sentry Env Vars to vercel.

PR #458, opened 2022-01-25
Fixes pfp issue on map page

### LINKED ISSUE #417: Profile image not showing on map view or allocation page
**Problem**
User updated profile image and profile image is not updated on map or on allocate page. 

**Map view:**
profile pic problem
**Allocation page view:**
profile pic problem2

Tried this in Firefox and Chrome.

f1lander:
fixes #417 

### problem
the `profile.avatar` is for some reason an empty string that's the reason why the map page is not loading the pfp.

image

### Solution

A non-empty string validation for the avatar value.

```
const p = profile ?? user?.profile;
let avatar = p?.avatar;
if (avatar === '') avatar = user?.profile?.avatar;
```
Screen Shot 2022-01-25 at 00 20 37

PS. @zashton address the `profile.updated_at` is not updated like the `user.updated_at` 
image

@exrhizo it could be something about the recoil state profile object request here? 
image

### DIFF
--- a/src/components/ApeAvatar/ApeAvatar.tsx
+++ b/src/components/ApeAvatar/ApeAvatar.tsx
@@ -18,10 +18,10 @@ export const ApeAvatar = ({
   profile?: IProfile;
 }) => {
   // TODO: simplify so all: <ApeAvatar path={getAvatarPath(p?.avatar)} />
-  const p = profile ?? user?.profile;
-  let avatar = p?.avatar;
-  if (avatar === '') avatar = user?.profile?.avatar;
-  const avatarPath = getAvatarPathWithFallback(avatar, user?.name);
+  const avatarPath = getAvatarPathWithFallback(
+    profile?.avatar || user?.profile?.avatar,
+    user?.name
+  );
   const src = path ?? avatarPath;
   return (
     <Avatar src={src} alt={user?.name} {...props}>
--- a/src/components/ApeAvatar/ApeAvatar.tsx
+++ b/src/components/ApeAvatar/ApeAvatar.tsx
@@ -19,7 +19,9 @@ export const ApeAvatar = ({
 }) => {
   // TODO: simplify so all: <ApeAvatar path={getAvatarPath(p?.avatar)} />
   const p = profile ?? user?.profile;
-  const avatarPath = getAvatarPathWithFallback(p?.avatar, user?.name);
+  let avatar = p?.avatar;
+  if (avatar === '') avatar = user?.profile?.avatar;
+  const avatarPath = getAvatarPathWithFallback(avatar, user?.name);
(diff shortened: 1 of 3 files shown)


zashton:
tried to clear the token gifts historical cache for circle on the backend upon pfp update, seem unrelated to this issue

src/components/ApeAvatar/ApeAvatar.tsx
```diff
@@ -19,7 +19,9 @@ export const ApeAvatar = ({
 }) => {
   // TODO: simplify so all: <ApeAvatar path={getAvatarPath(p?.avatar)} />
   const p = profile ?? user?.profile;
-  const avatarPath = getAvatarPathWithFallback(p?.avatar, user?.name);
+  let avatar = p?.avatar;
+  if (avatar === '') avatar = user?.profile?.avatar;
```
levity, line 23:
```suggestion
  const avatar = p?.avatar || user?.profile?.avatar;
```

src/recoilState/map.ts
```diff
@@ -181,9 +181,13 @@ export const rMapGraphData = selector<GraphData>({
             `Missing user of circle = ${epoch.circle_id} in rMapGraphData at ${profile.address}`
           );
 
+          const p = profile ?? user?.profile;
+          let avatar = p?.avatar;
+          if (avatar === '') avatar = user?.profile?.avatar;
```
levity, line 186:
```suggestion
          const avatar = p?.avatar || user?.profile?.avatar;
```

levity (changes requested):
do these more concise versions look good to you?

src/components/ApeAvatar/ApeAvatar.tsx
```diff
@@ -19,7 +19,9 @@ export const ApeAvatar = ({
 }) => {
   // TODO: simplify so all: <ApeAvatar path={getAvatarPath(p?.avatar)} />
   const p = profile ?? user?.profile;
```
levity, line 21:
```suggestion
```

exrhizo:
I think you are right edax, I can look at it tomorrow, 
if `rUserProfileMap` doesn't have avatars for some reason then that would cause this...
maybe because it uses `createFakeProfile(assertDef(us.first()))`

This is all convoluted, but the source of this sort of thing is that the profile may be missing for a given user and addresses can be changed.

This is one place that refactoring with the database to force consistency would be helpful

src/components/ApeAvatar/ApeAvatar.tsx
f1lander: this is clever, thanks! I was falling asleep 😪

levity (dismissed):

exrhizo (approved):
LGTM

Issue #450, opened 2022-01-21
Setup local s3 for image hosting in localdev


CryptoGraffe:

spydermonkee:
@CryptoGraffe - is this for Hasura Migration? I want to add it to the milestone if so. I assume it's for enabling development? Perhaps you can add a description?

PR #462, opened 2022-01-26
Adds a local s3 implementation using LocalStack

CryptoGraffe:
* addresses issue #450 
* this is hooked up to the laravel storage system, but will be used for future hasura image/file work
* I had to manually do a `docker-compose build` to rebuild the larevel Dockerfile changes that are in the submodule

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,8 @@ REACT_APP_API_BASE_URL=http://localhost:9000/api
 REACT_APP_HASURA_URL=http://localhost:8080/v1/graphql
 # NODE: Used for the node scripts and vercel serverless functions
 NODE_HASURA_URL=http://localhost:8080/v1/graphql
+# local version of s3 for local testing
+REACT_APP_S3_BASE_URL=http://coordinape.s3-website.localhost.localstack.cloud:4566
 
 # Staging Environment
 #REACT_APP_API_BASE_URL=https://staging-api.coordinape.com/api
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,31 @@ services:
       HASURA_GRAPHQL_AUTH_HOOK: http://host.docker.internal:${LOCAL_VERCEL_PORT}/api/hasura/auth
       ## HASURA_GRAPHQL_UNAUTHORIZED_ROLE: public
 
+
+  # for local s3 storage for images
+  localstack:
+    image: localstack/localstack:latest
+    hostname: s3
+    networks:
+      default:
+        aliases:
+          - s3.localhost.localstack.cloud
+          - s3-website.localhost.localstack.cloud
+          - coordinape.s3.localhost.localstack.cloud
+          - coordinape.s3-website.localhost.localstack.cloud
+    environment:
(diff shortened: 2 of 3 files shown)


.env.example
```diff
@@ -16,6 +16,8 @@ LOCAL_CIRCLE_ADMIN=0xYOUR_DEV_ADDRESS
 # Local Environment
 REACT_APP_API_BASE_URL=http://localhost:9000/api
 REACT_APP_HASURA_URL=http://localhost:8080/v1/graphql
+# local version of s3 for local testing
+REACT_APP_S3_BASE_URL=http://coordinape.s3-website.localhost.localstack.cloud:4566
```
exrhizo, line 20: I wonder if this can use a URL, that is just `http://localhost:4566` otherwise not a big deal.

docker-compose.yml
```diff
@@ -31,6 +31,31 @@ services:
...
+  localstack:
+    image: localstack/localstack:latest
+    hostname: s3
+    networks:
+      default:
+        aliases:
+          - s3.localhost.localstack.cloud
+          - s3-website.localhost.localstack.cloud
+          - coordinape.s3.localhost.localstack.cloud
+          - coordinape.s3-website.localhost.localstack.cloud
```
exrhizo, line 45: oh I see, this was done here, is it needed to have so many aliases?

I imagine we could do this with all of the docker images, but probably not worth it, although more readable

docker-compose.yml
```diff
@@ -31,6 +31,31 @@ services:
...
+    environment:
+      - HOSTNAME_EXTERNAL=localstack
+      - AWS_DEFAULT_REGION=us-east-1
+      - EDGE_PORT=4566
```
exrhizo, line 49: having the port as a .env variable seems good, with the other variables

exrhizo (commented):
This can be branched off of `main`

docker-compose.yml
CryptoGraffe: I was following examples from LocalStack, i think there is variance in how different versions of s3 refers to the host, older versions use folder paths and newer versions use subdomains. Hence the 2 versions. Then the two other hosts are the webserving hosts, so 4 total.

docker-compose.yml
CryptoGraffe: None of the other ports in this file have vars for them, how do you want to do it? There is a range for localstack but i dont think its necessary, we probably only need 4566.

CryptoGraffe:
> This can be branched off of `main`

Switched to main.

docker-compose.yml
CryptoGraffe: Never mind i see it in main, i'll sort this out

exrhizo: this is all that is needed to be ready to merge, we can do more for improving env variables in another PR

@levity has feelings about the env as well

exrhizo: eh, don't want to complexify the env, if it comes up, we can address it then

exrhizo:
I created https://github.com/coordinape/coordinape/pull/512
and will merge that shortly

Issue #464, opened 2022-01-26
LOCAL var mismatches in seeding scripts


CryptoGraffe:

exrhizo:
temp solution add:
`NODE_HASURA_ADMIN_SECRET=admin-secret`

exrhizo:
https://github.com/coordinape/coordinape/pull/472

exrhizo:
I don't know what the best thing would be, because this `api-lib` vars are used by both the serverless functions, therefore they need the NODE_HASURA_URL prefix. But when used for the local scripts, they should be set to the local versions....

There is a bigger question, which is what should the interface to the gql code look like
`src/lib/gql/index.ts` is user scoped
`src/lib/gql/GqlHasuraAdmin.ts` is admin scoped

@levity you changed the `src/lib/gql/index.ts` to be strictly tied to 
`import { getAuthToken } from 'services/api';`
That is complecting it with our react app, ideally it would be an independent package

I did a kind of stand in solution but these interfaces could really use work

exrhizo:
I didn't mean to close this, I am not good at github issues yet

PR #465, opened 2022-01-26
Renaming vars to match .env.example and adding missing var

CryptoGraffe:
Addresses #464

### DIFF
--- a/api-lib/Gql.ts
+++ b/api-lib/Gql.ts
@@ -1,5 +1,6 @@
 import { Gql } from '../src/lib/gql/GqlHasuraAdmin';
-import { NODE_HASURA_ADMIN_SECRET, NODE_HASURA_URL } from './config';
+
+import { LOCAL_HASURA_ADMIN_SECRET, LOCAL_HASURA_URL } from './config';
 
 export * from '../src/lib/gql/GqlHasuraAdmin';
-export const gql = new Gql(NODE_HASURA_URL, NODE_HASURA_ADMIN_SECRET);
+export const gql = new Gql(LOCAL_HASURA_URL, LOCAL_HASURA_ADMIN_SECRET);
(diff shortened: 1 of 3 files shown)


zashton:
was encountering the same issue when trying to run the seeder

crabsinger:
@CryptoGraffe unfortunately "addresses" is not on of these terms: https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

can you change the PR description to use one of these terms:
```

    close
    closes
    closed
    fix
    fixes
    fixed
    resolve
    resolves
    resolved
```

api-lib/Gql.ts
```diff
@@ -1,5 +1,6 @@
 import { Gql } from '../src/lib/gql/GqlHasuraAdmin';
-import { NODE_HASURA_ADMIN_SECRET, NODE_HASURA_URL } from './config';
+
+import { LOCAL_HASURA_ADMIN_SECRET, LOCAL_HASURA_URL } from './config';
```
exrhizo, line 3: This is confusing, and I don't know what the best thing would be, because these api-lib is used by both the serverless functions, therefore they need the `NODE_HASURA_URL` prefix.

But when used for the local scripts, they should be set to the local versions....

I am going to make a fix PR for this

exrhizo:
Three is a bigger question, which is what should the interface to the gql code look like
`src/lib/gql/index.ts` is user scoped
`src/lib/gql/GqlHasuraAdmin.ts` is admin scoped

@levity you changed the `src/lib/gql/index.ts` to be strictly tied to 
`import { getAuthToken } from 'services/api';`

because that is complecting it to our react app, but it would be best to be a package

I can take another go at this, but I won't get to it today.

exrhizo:
https://github.com/coordinape/coordinape/pull/472

api-lib/Gql.ts
CryptoGraffe: I looked at the usages but it was only used by the seed scripts in main, and wasn't sure how this is all mapped out

exrhizo: I didn't mean to say that what you did is confusing,
I mean what I implemented and the last minute change lawrence and I made led to a situation that was confusing

I appreciate you working on everything and helping the team get it to a sensible place

PR #482, opened 2022-01-27
Remove esrun in favor of more stable dependencies

topocount:
esrun passes raw output from esbuild as an argument to a new node
process. unfortunately, many machines only have about 200kb-300kb
allotted for arguments passed into new processes.

By creating a tempfile and passing that file path into a node
subprocess, we avoid this issue of argument size.

test plan:
Try out existing scripts:

- yarn db-seed
- yarn db-clean
- yarn db-seed && yarn db-seed will demonstrate error output, which is
  now bubbled up and logged to stderr


exrhizo (approved):
Sweet

PR #483, opened 2022-01-27
Switch to react router v6

exrhizo:
As far as I can tell, this should unblock storybook working w/ our app

Following this guide,
https://reactrouter.com/docs/en/v6/upgrading/v5


levity (approved):
This works well.

There's one issue, which I'll leave up to you to decide if it should block the merge. Will comment on it inline in a sec

src/routes/routes.tsx
```diff
@@ -24,26 +24,24 @@ import * as paths from './paths';
 // look into this.
 const LazyAssetMapPage = lazy(() => import('pages/AssetMapPage'));
 
-export const Routes = () => {
+export const AppRoutes = () => {
   const hasCircles = useHasCircles();
   const hasSelectedCircle = useHasSelectedCircle();
```
levity, line 29: one or both of these hooks return `false` initially even for users where they will eventually return `true`. so if i load e.g. /history, even though the right page will eventually load, i'll get `No routes matched location "/history"` in the console at first.

this prevents us from putting a catch-all route (`<Route path="*" element={<Navigate to={paths.getHomePath()} replace />} />`) at the end of the logged-out routes, because it will also prevent getting to logged-in routes.

the absence of a catch-all there means that if you're a user that's not in any circle yet, and you visit a route like /allocation, you don't get redirected to the "This wallet isn't associated with a circle" page. you just get a blank page and `No routes matched location "/allocation"` in the console.

exrhizo: Yes this is an existing issue and although I touched it, I don't feel like I am in it enough, that I want to resolve it. When we do the org overview panel, that will connect with this

exrhizo:
@levity Thank you for getting this to the finish line :)

f1lander:
@exrhizo thanks! And sorry because I was able to get this solution before but I didn't communicate this int that right ways and I was thinking of solve with current version was the only option.

PR #493, opened 2022-01-27
WIP, DO NOT MERGE: Feature/ape 404 check nominees

jsur:
WIP version of check-nominee.ts lambda function to check if `gql`, `zeus` etc. usage is what's wanted.

### DIFF
--- a/api/cron/check-nominee.ts
+++ b/api/cron/check-nominee.ts
@@ -1,6 +1,8 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node';
-import { GraphQLClient, gql } from 'graphql-request';
 
+import { gql } from '../../api-lib/Gql';
+
+// TODO: remove this
 /*
 
         $expired_nominees = $this->model->with('circle','nominations')->where('ended',0)->pastExpiryDate()->get();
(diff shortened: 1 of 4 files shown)


api/cron/check-nominee.ts
```diff
@@ -0,0 +1,93 @@
...
+          sendSocialMessage({
+            message: `Nominee ${n.name} has only received ${n.nominations.length} vouch(es) and has failed`,
+            circleId: n.circle_id,
+          })
+        )
+      );
+      return res.status(200).json({ update_nominees });
+    }
+
+    res.status(200).json({ message: 'No updates' });
```
jsur, line 86: What do we want to return when there's no updated nominees?

src/lib/gql/zeusHasuraAdmin/index.ts
```diff
@@ -16240,6 +16240,9 @@ const handleFetchResponse = (
 export const apiFetch =
   (options: fetchOptions) =>
   (query: string, variables: Record<string, any> = {}) => {
+    if (!!process.env.NODE_ENV) {
+      globalThis.fetch = nodeFetch as unknown as typeof fetch;
```
zashton, line 16244: how about splitting this to a separate PR ?

api/cron/check-nominee.ts
```diff
@@ -0,0 +1,93 @@
...
+              _eq: false,
+            },
+            expiry_date: { _lte: new Date().toISOString() },
+          },
+        },
+        {
+          id: true,
+          name: true,
+          circle_id: true,
+          nominations: [{}, { id: true }], // TODO: aggregate query somehow?
```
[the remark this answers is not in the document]
zashton: you could use
`nominations_aggregate {
      aggregate {
        count
      }
    }
`

api/cron/check-nominee.ts
zashton: its just saved inside logs, we could just keep it as no updates for now.

jsur: Ok!

api/cron/check-nominee.ts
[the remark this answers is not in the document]
jsur: Great, that works, thanks.

src/lib/gql/zeusHasuraAdmin/index.ts
jsur: Sure why not.

zashton: that will be great thanks! will def help unblock others

PR #496, opened 2022-01-28
Spike/storybook

f1lander:
### Summary
Enables Storybook into our project, the `Button` story is just an example, new stories will be added in another PR's.

BTW, I guess I can add the vercel config to build the storybook too.


package.json
```diff
@@ -111,6 +113,16 @@
     "extends": [
       "react-app",
       "react-app/jest"
+    ],
```
levity, line 116: does this work if you add it to `.eslintrc` instead? we should keep all rules there if possible

levity (commented):
did you mean to check in some story files? i don't see anything relating to Button in the diff

f1lander:
Sorry I forgot to push that change, I'll push later

package.json
f1lander: Let me try, I guess this was added automatically

exrhizo (dismissed):
LGTM, let's merge!

tsconfig.json
```diff
@@ -15,7 +15,11 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react-jsx",
-    "baseUrl": "src"
+    "baseUrl": "src",
+    "paths": {
+      "@emotion/core": ["./types/emotion-core-stub.d.ts"],
```
exrhizo, line 20: would it be useful to install emotion types?

exrhizo (approved):

PR #497, opened 2022-01-28
use node-fetch in serverless functions

jsur:
Node.js does not have `fetch` available globally, thus we need to include `node-fetch` when querying the db with Hasura from serverless functions.


exrhizo (approved):
LGTM

exrhizo:
maybe we change this at some point to be in a different location, I suspect we will refactor all of this before we really feel settled with the new environment

PR #499, opened 2022-01-30
Feature/ape 404: check nominees cron trigger

jsur:
**Deploy TODO:**

- Add env var `HASURA_API_BASE_URL` to staging and prod.
- Add env car `TELEGRAM_BOT_TOKEN` to staging and prod

**Points open for discussion:**

- Telegram API usage still unclear, where does `chat_id` come from? It's available at least in table `profiles`. ✅
- What webhook url should a cron trigger call? Does `docker-compose` need something for cron triggers to work? ✅
- Caesar the Coordinape bot avatar url needs to be filled in for Discord notifs ✅

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,9 @@ HASURA_EVENT_SECRET=event-secret
 # Same as prisma but replace s/localhost/postgres
 HASURA_GRAPHQL_DATABASE_URL=postgres://postgres:postgrespassword@postgres:5432/postgres
 
+# Social messages
+TELEGRAM_BOT_TOKEN=<your Telegram testing bot token>
+
 # Local Environment
 REACT_APP_API_BASE_URL=http://localhost:9000/api
 REACT_APP_HASURA_URL=http://localhost:8080/v1/graphql
--- a/.env.example
+++ b/.env.example
@@ -26,17 +26,14 @@ HASURA_EVENT_SECRET=event-secret
 # Same as prisma but replace s/localhost/postgres
 HASURA_GRAPHQL_DATABASE_URL=postgres://postgres:postgrespassword@postgres:5432/postgres
 
-# Social messages
-TELEGRAM_BOT_TOKEN=<your Telegram testing bot token>
-
 # Local Environment
 REACT_APP_API_BASE_URL=http://localhost:9000/api
 REACT_APP_HASURA_URL=http://localhost:8080/v1/graphql
 REACT_APP_S3_BASE_URL=http://coordinape.s3-website.localhost.localstack.cloud:4566
 # NODE: Used for the node scripts and vercel serverless functions
 NODE_HASURA_URL=http://localhost:8080/v1/graphql
-HASURA_EVENT_SECRET=event-secret
-HASURA_GRAPHQL_ADMIN_SECRET=admin-secret
+REACT_APP_S3_BASE_URL=http://coordinape.s3-website.localhost.localstack.cloud:4566
+TELEGRAM_BOT_TOKEN=<your Telegram testing bot token>
 
 # Staging Environment
 #REACT_APP_API_BASE_URL=https://staging-api.coordinape.com/api
--- a/api-lib/sendSocialMessage.ts
+++ b/api-lib/sendSocialMessage.ts
@@ -47,52 +47,42 @@ export async function sendSocialMessage({
   const [{ discord_webhook, telegram_id }] = circles;
 
   if (channels?.discord && discord_webhook) {
-    try {
-      const discordWebhookPost = {
-        content: msg,
-        username: DISCORD_BOT_NAME,
-        avatar_url: DISCORD_BOT_AVATAR_URL,
-      };
-      const res = await fetch(discord_webhook, {
-        method: 'POST',
-        body: JSON.stringify(discordWebhookPost),
-        headers: {
-          'Content-Type': 'application/json',
-        },
-      });
+    const discordWebhookPost = {
+      content: msg,
+      username: DISCORD_BOT_NAME,
+      avatar_url: DISCORD_BOT_AVATAR_URL,
+    };
+    const res = await fetch(discord_webhook, {
+      method: 'POST',
+      body: JSON.stringify(discordWebhookPost),
+      headers: {
+        'Content-Type': 'application/json',
+      },
+    });
 
-      if (!res.ok) {
-        throw new Error(JSON.stringify(await res.json()));
-      }
-    } catch (e) {
-      // TODO: what/where do we want to log?
-      console.log('Discord webhook error:', e);
+    if (!res.ok) {
+      throw new Error(JSON.stringify(await res.json()));
     }
   }
 
   if (channels?.telegram && telegram_id) {
-    try {
-      const telegramBotPost = {
-        chat_id: telegram_id,
-        text: msg,
-      };
-      const res = await fetch(
-        `https://api.telegram.org/bot${process.env.TELEGRAM_BOT_TOKEN}/sendMessage`,
-        {
-          method: 'POST',
-          body: JSON.stringify(telegramBotPost),
-          headers: {
-            'Content-Type': 'application/json',
-          },
-        }
-      );
-
-      if (!res.ok) {
-        throw new Error(JSON.stringify(await res.json()));
+    const telegramBotPost = {
+      chat_id: telegram_id,
+      text: msg,
+    };
+    const res = await fetch(
+      `https://api.telegram.org/bot${process.env.TELEGRAM_BOT_TOKEN}/sendMessage`,
+      {
+        method: 'POST',
+        body: JSON.stringify(telegramBotPost),
+        headers: {
+          'Content-Type': 'application/json',
+        },
       }
-    } catch (e) {
-      // TODO: what/where do we want to log?
-      console.log('Telegram bot error:', e);
+    );
+
+    if (!res.ok) {
+      throw new Error(JSON.stringify(await res.json()));
     }
   }
 }
--- a/api-lib/sendSocialMessage.ts
+++ b/api-lib/sendSocialMessage.ts
@@ -3,6 +3,8 @@ import fetch from 'node-fetch';
 import { gql } from './Gql';
 
 const DISCORD_BOT_NAME = 'Caesar The Coordinape Bot';
+const DISCORD_BOT_AVATAR_URL =
+  'https://coordinape-prod.s3.amazonaws.com/assets/static/20220131-coordinape-pfp-ape.jpg';
 
 type SocialMessage = {
   message: string;
@@ -49,7 +51,7 @@ export async function sendSocialMessage({
       const discordWebhookPost = {
         content: msg,
         username: DISCORD_BOT_NAME,
-        // avatar_url: '' TODO
+        avatar_url: DISCORD_BOT_AVATAR_URL,
       };
       const res = await fetch(discord_webhook, {
         method: 'POST',
--- a/api-lib/sendSocialMessage.ts
+++ b/api-lib/sendSocialMessage.ts
@@ -45,7 +45,6 @@ export async function sendSocialMessage({
   const [{ discord_webhook, telegram_id }] = circles;
 
   if (channels?.discord && discord_webhook) {
-    // https://discord.com/api/webhooks/937364127595245569/dZGmCuPxt3knbTEl1QtvVQnyppK8CoHV9dvBdsvhobAPV7gMC-WL6Cpct5uQTgiRfpsB
     try {
       const discordWebhookPost = {
         content: msg,
@@ -70,14 +69,13 @@ export async function sendSocialMessage({
   }
 
   if (channels?.telegram && telegram_id) {
-    // https://api.telegram.org/bot<your-bot-token>/sendMessage?chat_id=<chat-id>&text=TestReply
     try {
       const telegramBotPost = {
-        chat_id: 'TODO', // what goes here?
-        text: encodeURIComponent(msg),
+        chat_id: telegram_id,
+        text: msg,
       };
       const res = await fetch(
-        `https://api.telegram.org/bot${telegram_id}/sendMessage`,
+        `https://api.telegram.org/bot${process.env.TELEGRAM_BOT_TOKEN}/sendMessage`,
         {
           method: 'POST',
           body: JSON.stringify(telegramBotPost),
--- a/api-lib/sendSocialMessage.ts
+++ b/api-lib/sendSocialMessage.ts
@@ -49,7 +49,7 @@ export async function sendSocialMessage({
     }
   }
 
-  if (channels?.telegram && telegram_id) {
+  if (TELEGRAM_BOT_BASE_URL && channels?.telegram && telegram_id) {
     const telegramBotPost = {
       chat_id: telegram_id,
       text: msg,
--- a/api-lib/sendSocialMessage.ts
+++ b/api-lib/sendSocialMessage.ts
@@ -1,0 +1,98 @@
+import fetch from 'node-fetch';
+
+import { gql } from './Gql';
+
+const DISCORD_BOT_NAME = 'Caesar The Coordinape Bot';
+
+type SocialMessage = {
+  message: string;
+  circleId: number;
+  sanitize: boolean;
+  channels: {
+    discord?: boolean;
+    telegram?: boolean;
+  };
+};
+
+function cleanStr(str: string) {
+  return str.replace(/:|-|\/|\*|_|`/g, '');
+}
+
+export async function sendSocialMessage({
+  message,
+  circleId,
+  sanitize = true,
+  channels,
+}: SocialMessage) {
+  const msg = sanitize ? cleanStr(message) : message;
+
+  const { circles } = await gql.q('query')({
+    circles: [
+      {
+        where: {
+          id: {
+            _eq: circleId,
+          },
+        },
+      },
+      {
+        discord_webhook: true,
+        telegram_id: true,
+      },
+    ],
+  });
+
+  const [{ discord_webhook, telegram_id }] = circles;
+
+  if (channels?.discord && discord_webhook) {
+    // https://discord.com/api/webhooks/937364127595245569/dZGmCuPxt3knbTEl1QtvVQnyppK8CoHV9dvBdsvhobAPV7gMC-WL6Cpct5uQTgiRfpsB
+    try {
+      const discordWebhookPost = {
+        content: msg,
+        username: DISCORD_BOT_NAME,
+        // avatar_url: '' TODO
+      };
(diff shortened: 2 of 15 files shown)


zashton:
> Points open for discussion:
> 
> * Telegram API usage still unclear, where does `chat_id` come from? It's available at least in table `profiles`.
> * What webhook url should a cron trigger call? Does `docker-compose` need something for cron triggers to work?
> * Caesar the Coordinape bot avatar url needs to be filled in for Discord notifs

Cron webhook url will be similar to how you add your action endpoint, but its in the events tab. You'll need to add a base url env var in the yml file.

For discord notif, because of the multi-tenant nature of the system, I made it so that each circle will need to create their own discord webhook within their own discord guild, theres a walkthrough guide here. [LINK](https://docs.coordinape.com/welcome/discord_bot) . You can create a circle [here](https://app.coordinape.com/new-circle) in the live server to try it out. 

For the telegram notif , the `chat_id` is actually a unique key to identify each chatroom/conversation, in this case its referencing from the `circles` table. I made it very early on just for specific organisations (I filled in the `chat_id` manually), currently for the new organisations none of them uses it. After you are done setting up your circle, create a telegram group do let me know thru DM, i can invite in my bot in your chatroom and you can try and play and test its notifications

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  const [{ discord_webhook, telegram_id }] = circles;
+
+  if (channels?.discord && discord_webhook) {
+    try {
+      const discordWebhookPost = {
+        content: msg,
+        username: DISCORD_BOT_NAME,
+        // avatar_url: '' TODO
```
crabsinger, line 52: https://coordinape-prod.s3.amazonaws.com/assets/static/20220131-coordinape-pfp-ape.jpg

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,96 @@
...
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      });
+
+      if (!res.ok) {
+        throw new Error(JSON.stringify(await res.json()));
+      }
+    } catch (e) {
+      // TODO: what/where do we want to log?
```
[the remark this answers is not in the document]
crabsinger: Would also be nice to be able to retry these for example if Discord goes down, how do we retry them once it's API is back up?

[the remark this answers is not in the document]
META-DREAMER: Hasura has its internal ping logging system so we can just start with that. If we return the proper HTTP error responses in the event handler, we can configure retries etc within Hasura: https://hasura.io/docs/latest/graphql/core/event-triggers/create-trigger.html#retry-logic

In this case I would probably setup separate triggers / handlers for discord and telegram webhooks so that we can manage their error state and retries individually.

hasura/metadata/cron_triggers.yaml
```diff
@@ -1 +1,5 @@
-[]
+- name: check-nominee
+  webhook: '{{HASURA_GRAPHQL_CRON_URL}}/check-nominee'
```
crabsinger, line 2: We should add a deploy TODO in the PR description to make sure this ENV_VAR is set for staging and prod environemnts too.

docker-compose.yml
```diff
@@ -69,7 +71,7 @@ services:
     tty: true
     ports:
       - ${LOCAL_LARAVEL_PORT}:80
-      - "443:443"
```
crabsinger, line 72: @jsur is this just your editor, or is there a reason we should be using single quotes vs double quotes?

crabsinger:
@jsur nice work on this PR! Seems like a great start with a few additions before merge ready.

Things I'm aware of right now:
-  logging
- separate telegram and discord into separate jobs (either can fail separately)
- add bot image URL

api-lib/sendSocialMessage.ts
jsur: Added.

docker-compose.yml
jsur: That seems to be Prettier using `.prettierrc.json` from the project root level:

```
["INFO" - 10:38:46 PM] Detected local configuration (i.e. .prettierrc or .editorconfig), VS Code configuration will not be used
["INFO" - 10:38:46 PM] Prettier Options:
{
  "filepath": "/Users/juliussuominen/code/coordinape/docker-compose.yml",
  "parser": "yaml",
  "singleQuote": true,
  "arrowParens": "avoid"
}
["INFO" - 10:38:46 PM] Formatting completed in 0.017ms.
```

hasura/metadata/databases/databases.yaml
```diff
@@ -6,4 +6,4 @@
         from_env: HASURA_GRAPHQL_DATABASE_URL
       isolation_level: read-committed
       use_prepared_statements: false
-  tables: '!include default/tables/tables.yaml'
+  tables: "!include default/tables/tables.yaml"
```
jsur, line 9: @crabsinger whereas this double quote is Hasura editing metadata files when something is saved from the console.

hasura/metadata/cron_triggers.yaml
jsur: Added.

api/hasura/cron/check-nominee.ts
```diff
@@ -0,0 +1,75 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { gql } from '../../../api-lib/Gql';
+import { sendSocialMessage } from '../../../api-lib/sendSocialMessage';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  try {
+    const { nominees } = await gql.q('query')({
+      nominees: [
```
exrhizo, line 9: shared libraries

api-lib/sendSocialMessage.ts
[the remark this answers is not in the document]
jsur: If only `sendSocialMessage` throws, the mutation that updates `nominees.ended` as `true` has already been committed in the database. Thus the nominee that was updated on the first handler execution won't be picked up on the retry.

Is there a way to rollback a GraphQL mutation on error?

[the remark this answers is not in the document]
zashton: Maybe you can try @META-DREAMER 's suggestion and use an event trigger onto nominee table's update event
then you won't need to do error handling since it runs async in the background

https://hasura.io/docs/latest/graphql/core/event-triggers/payload.html
could just within the event handler logic do a check on the payload old vs new ended =  false > true if yes trigger the send social message, and when creating the event trigger you can set the retries/timeout/intervals.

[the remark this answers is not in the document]
jsur: Ah right, I didn't fully understand what I read - I'll take a look at event triggers next. One more thing: `check-nominee.ts` already queries nominees with this where clause:

```
          where: {
            ended: {
              _eq: false,
            },
            expiry_date: { _lte: new Date().toISOString() },
          },
```

If the event trigger handler should also check the payload for `nominee.ended = false > true` etc., isn't that essentially duplicating the business logic into two places?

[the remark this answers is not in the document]
jsur: Yeah, I agree - I was just wondering if duplicating this sort of logic into to places could be confusing in the future if there's a lot of similar use cases (cron trigger + event handler), but there also doesn't seem to be a better option if we want to use Hasura retry logic. Anyhow, I'll work on the event triggers based on this convo 👍

api/hasura/cron/check-nominee.ts
jsur: Not following, should this query be moved somewhere else or?

jsur:
I opened this PR now, feel free to comment!

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+
+export async function sendSocialMessage({
+  message,
+  circleId,
+  sanitize = true,
+  channels,
+}: SocialMessage) {
+  const msg = sanitize ? cleanStr(message) : message;
+
+  const { circles } = await gql.q('query')({
```
exrhizo, line 31: see how Zashton moved these to `import { Gql } from '../src/lib/gql/GqlHasuraAdmin';`
and actually it doesn't need to be admin, so just regular '../src/lib/gql';`

api/hasura/cron/check-nominee.ts
exrhizo: yes, I gave more details above

api/hasura/event-triggers/check-nominee-discord.ts
```diff
@@ -0,0 +1,20 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { EventTriggerPayload } from '../types';
+
+import handleCheckNomineeMsg from './utils/handleCheckNomineeMsg';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
```
exrhizo, line 7: I noticed that topocount made a verification middleware, I think we will want the same one here before merging

exrhizo (commented):
Hey I have looked some, but not enough to really understand it, we need someone to do that

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  if (channels?.telegram && telegram_id) {
+    const telegramBotPost = {
+      chat_id: telegram_id,
+      text: msg,
+    };
+    const res = await fetch(
+      `https://api.telegram.org/bot${process.env.TELEGRAM_BOT_TOKEN}/sendMessage`,
```
exrhizo, line 74: Move this to `api-lib/config.ts`

hasura/metadata/cron_triggers.yaml
exrhizo: I think it is better to just write it out? or did hasura generate that itself?
Does hasura automagically create that variable for itself, because it does that for the other `HASURA_GRAPHQL` variables I believe

zashton: > I think it is better to just write it out? or did hasura generate that itself? Does hasura automagically create that variable for itself, because it does that for the other `HASURA_GRAPHQL` variables I believe

@META-DREAMER recommended the use of env var for webhook handlers, so i advised @jsur to do the same
(https://github.com/coordinape/coordinape/pull/423#discussion_r786488767)

api/hasura/event-triggers/check-nominee-discord.ts
zashton: we can let @topocount 's PR merge first to main, so @jsur can easily just pull in the changes

hasura/metadata/cron_triggers.yaml
jsur: Yup, the env var also has a new name: `HASURA_CRON_BASE_URL`.

.env.example
```diff
@@ -22,6 +22,9 @@ HASURA_GRAPHQL_ADMIN_SECRET=admin-secret
 # Same as prisma but replace s/localhost/postgres
 HASURA_GRAPHQL_DATABASE_URL=postgres://postgres:postgrespassword@postgres:5432/postgres
 
+# Social messages
```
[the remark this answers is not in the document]
jsur: No problem, changed.

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,88 @@
+import fetch from 'node-fetch';
+
+import { gql } from './Gql';
+
+const DISCORD_BOT_NAME = 'Caesar The Coordinape Bot';
+const DISCORD_BOT_AVATAR_URL =
+  'https://coordinape-prod.s3.amazonaws.com/assets/static/20220131-coordinape-pfp-ape.jpg';
```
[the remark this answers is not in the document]
jsur: 👍

zashton:
@jsur 
hooked up my telegram & discord bots to try out your handlers - cron/event triggers working well & notifications are sent out :thumbsup:

api-lib/sendSocialMessage.ts
jsur: 👍

api/hasura/event-triggers/check-nominee-discord.ts
jsur: 🆒

api-lib/sendSocialMessage.ts
jsur: I see, do you want these exact queries into `../src/lib/gql` or more general ones? In this case the more general one would be e.g. `getCircleById`?

jsur: Btw, `../src/lib/gql` doesn't export an api object, but rather the comment in it tells to do this:

```
import { REACT_APP_HASURA_URL } from 'config/env';
import { getAuthToken } from 'services/api';
import { getGql } from 'lib/gql';

const api = getApi(REACT_APP_HASURA_URL, getAuthToken);
await api.updateProfile();
```

It seems like it'd be easier to export the `api` to avoid having to refer to `REACT_APP_HASURA_URL` etc. all over app code.

levity: Auth token isn't available until after the user logs in

exrhizo: I think that comment is confusing, rather, I think we ought to link to the usages..

But like lawrence says this instruction is for the react app.

For the vercel functions, see`api-lib/Gql.ts`
which has the admin client.

api-lib/sendSocialMessage.ts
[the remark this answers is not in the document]
META-DREAMER: > https://hasura.io/docs/latest/graphql/core/event-triggers/payload.html could just within the event handler logic do a check on the payload old vs new nominee.ended = false > true + its past the nominee expiry date & insufficient vouches. if yes trigger the send social message, and when creating the event trigger you can set the retries/timeout/intervals.

Yes this is the ideal way to do it because it guarantees that a notification will be sent no matter how that field gets updated. e.g. If theres another mutation we create in the future also updates this field then we don't have to duplicate sending the notification. Unidirectional data flow as much as possible.

hasura/metadata/cron_triggers.yaml
META-DREAMER: Another way to handle this is to just have a single endpoint to forward all cron events to and then branch the logic out from there. When Hasura calls the event handler it gives all the info on what event was triggered: https://hasura.io/docs/latest/graphql/core/event-triggers/payload.html#json-payload

something like:

```
type EventName = 'check-nominee' | ''
const EventHandlers = {
  'check-nominee': 
}


const handler = (hasuraEventPayload: EventTriggerPayload) => {
    
}
```

jsur: Makes sense, I can try that approach on the next cron job once this PR is done.

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,69 @@
...
+  message,
+  circleId,
+  sanitize = true,
+  channels,
+}: SocialMessage) {
+  const msg = sanitize ? cleanStr(message) : message;
+
+  const {
+    circles_by_pk: { discord_webhook, telegram_id },
+  } = await gql.getCircle(circleId);
```
jsur, line 31: @exrhizo I now used the admin client here. Do you want me to use it in `check-nominee.ts` for the query and mutation as well?

jsur:
TODOs left for this PR: 

- Possible Hasura admin client (`api-lib/Gql.ts`) usage in `check-nominee.ts`
- Once @topocount's PR #508 is merged, I'll add `verifyHasuraAdminMiddleware` usage in the handlers

api-lib/sendSocialMessage.ts
zashton: yeap , @jsur please do so

jsur: Cool, will take a look today.

zashton:
Fixes #522

api-lib/sendSocialMessage.ts
jsur: Added @zashton

jsur: Added also to `handleCheckNomineeMsg`

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,69 @@
...
+        'Content-Type': 'application/json',
+      },
+    });
+
+    if (!res.ok) {
+      throw new Error(JSON.stringify(await res.json()));
+    }
+  }
+
+  if (channels?.telegram && telegram_id) {
```
zashton, line 52: will be good to do a check whether `TELEGRAM_BOT_BASE_URL` has value here as well

docker-compose.yml
```diff
@@ -31,6 +31,9 @@ services:
       HASURA_GRAPHQL_ADMIN_SECRET: ${HASURA_GRAPHQL_ADMIN_SECRET}
       HASURA_EVENT_SECRET: ${HASURA_EVENT_SECRET}
       HASURA_GRAPHQL_AUTH_HOOK: http://${DOCKER_GATEWAY_HOST:-host.docker.internal}:${LOCAL_VERCEL_PORT}/api/hasura/auth
+      HASURA_CRON_BASE_URL: http://host.docker.internal:${LOCAL_VERCEL_PORT}/api/hasura/cron
+      HASURA_EVENT_TRIGGER_BASE_URL: http://host.docker.internal:${LOCAL_VERCEL_PORT}/api/hasura/event-triggers
+      TELEGRAM_BOT_TOKEN: ${TELEGRAM_BOT_TOKEN}
```
zashton, line 36: this shouldn't be required for the event hook to work, its getting from the config directly

zashton (changes requested):
All good, just some minor changes.

api-lib/sendSocialMessage.ts
jsur: Added to PR.

docker-compose.yml
jsur: Right, deleted from this file.

jsur:
Added `HASURA_API_BASE_URL ` and removed individual Hasura event env variables from `docker-compose`.

zashton (approved):
looks good.
@crabsinger can help add `HASURA_API_BASE_URL` env var to hasura staging ? and we can merge this

zashton:
@META-DREAMER any clue why the hasura preview is failing ?

crabsinger:
HASURA_API_BASE_URL has been added to hasura staging env vars.

crabsinger (approved):

topocount (approved):

topocount:
TELEGRAM_BOT_TOKEN exists in staging and prod. HASURA_API_BASE_URL is configured in docker-compose, and can be set in hasura console if it isn't already.

PR #500, opened 2022-01-30
Update setup w prisma and fetch-shim

exrhizo:
Fix the build instructions, running `prisma generate` is needed to create prisma's node_modules

Also, the global fetch to be included with an import
Having it in the generated files is hard to maintain.

Rename `setup-hardhat.sh` to `setup.sh`

### DIFF
--- a/api-lib/node-fetch-shim.ts
+++ b/api-lib/node-fetch-shim.ts
@@ -1,0 +1,8 @@
+import fetch from 'node-fetch';
+
+// This injects fetch into the global context for node.
+if (!globalThis.fetch) {
+  globalThis.fetch = fetch as unknown as typeof globalThis.fetch;
+}
+
+export { fetch };
--- a/scripts/generate-zeus.sh
+++ b/scripts/generate-zeus.sh
@@ -12,7 +12,7 @@ function generate() {
   local MV_PATH=$GEN_PATH/$1; shift
 
   # pass the rest of the arguments to zeus
-  zeus $HASURA_GRAPHQL_ENDPOINT $GEN_PATH --ts --rq $@
+  (set -x; zeus $HASURA_GRAPHQL_ENDPOINT/v1/graphql $GEN_PATH --ts --rq $@)
   test -d $GEN_PATH/zeus
   test -d "$MV_PATH" && rm -r $MV_PATH
   mv $GEN_PATH/zeus $MV_PATH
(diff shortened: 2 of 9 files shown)


scripts/generate-zeus.sh
```diff
@@ -12,7 +12,7 @@ function generate() {
   local MV_PATH=$GEN_PATH/$1; shift
 
   # pass the rest of the arguments to zeus
-  zeus $HASURA_GRAPHQL_ENDPOINT $GEN_PATH --ts --rq $@
+  (set -x; zeus $HASURA_GRAPHQL_ENDPOINT/v1/graphql $GEN_PATH --ts --rq $@)
```
levity, line 15: Why did this change?

levity (dismissed):

scripts/generate-zeus.sh
exrhizo: because it was wrong, from when I wrote it

exrhizo:
@levity 
 
Maybe I should just rename `setup.sh` back but this vercel command doesn't live in the repo?
 
 
image

levity:
>  @levity 
> 
>  
> 
> Maybe I should just rename `setup.sh` back but this vercel command doesn't live in the repo?

Right, either rename back or change the Vercel config

scripts/generate-zeus.sh
levity: that's weird, it was working for me as it was--i ran this several times to test it when i refactored it

api-lib/node-fetch-shim.ts
```diff
@@ -0,0 +1,8 @@
+import fetch from 'node-fetch';
+
+// This injects fetch into the global context for node.
+if (!globalThis.fetch) {
+  globalThis.fetch = fetch as unknown as typeof globalThis.fetch;
```
levity, line 5: do we still need the `@types/node-fetch` changes from ba558f9c? if not, can they be removed?

exrhizo: That was done wrong, I made a mistake merging it, @jsur put that in the generated code, if we want it in a place like that it should be in a place like `api-lib/Gql.ts` but I want to refactor all of that

scripts/generate-zeus.sh
exrhizo: weird, maybe it was something with renaming variables actually

api-lib/node-fetch-shim.ts
levity: sorry, i should have been more specific i guess. i meant specifically adding `@types/node-fetch` to `devDependencies` in package.json. if that's not needed, let's remove it

scripts/generate-zeus.sh
levity: hmm. ideally we wouldn't leave this `/v1/graphql` bit here when we have it in the env variable in other cases, for consistency's sake

PR #502, opened 2022-02-01
Remove unused codes

exrhizo:
These files and folders are unused


levity:
Looks like the removal of src/global causes Vercel to fail

levity:
Not sure if that's Vercel-specific or maybe just a reference that wasn't removed... try building locally?

PR #508, opened 2022-02-02
Create Hasura createUser Action

topocount:
I've defaulted to camelCase naming within graphQL because
that's the convention, but I'm happy to go back to snake case if that's
preferred.

I've grouped all potential updates/inserts into one mutation,
to minimize the i/o and better utilize graphQL.

This PR includes some authentication middlewares that will be useful for future endpoint implementations.

## Pre-merge todo

add `HASURA_EVENT_SECRET` with a  randomly generated value to the hasura cloud environment and/or vercel environment. Protip: they should be the same value


topocount:
@zashton @exrhizo @META-DREAMER 

I'm opening this up for review.

I'm not sure of the needed shape of the response after poking around for a while, and I still need to set up authentication as noted.

api/hasura/actions/create_user.ts
```diff
@@ -0,0 +1,116 @@
...
+          name: { _eq: createUserParams.name },
+          ended: { _eq: false },
+        },
+      },
+      { returning: { id: true, address: true, name: true } },
+    ],
+    // Create a profile if none exists yet
+    insert_profiles_one: [
+      {
+        object: { address: createUserParams.address },
```
META-DREAMER, line 91: FYI you can also do a nested insert instead of multiple mutations at the top level since it's for a related table to the user insert: https://hasura.io/docs/latest/graphql/core/databases/postgres/mutations/insert.html#one-to-one-one-to-many-relationships

This is fine for now too, just wanted to shout out nested inserts because they can be super useful (don't have to specify the primary key since it takes it from the foreign key inserted into the user table)

META-DREAMER (approved):
Few minor things, looks good otherwise!

scripts/generate-zeus.sh
```diff
@@ -12,16 +13,34 @@ function generate() {
...
+  if [[ "$PLATFORM" == "OSX" || "$PLATFORM" == "BSD" ]]; then
+    sed -i "" 's,bigint"]:any,bigint"]:number,g' "$MV_PATH"/index.ts
+    sed -i "" 's,bigint"]:unknown,bigint"]:number,g' "$MV_PATH"/index.ts
+  elif [ "$PLATFORM" == "LINUX" ]; then
+    sed -i 's,bigint"]:any,bigint"]:number,g' "$MV_PATH"/index.ts
+    sed -i 's,bigint"]:unknown,bigint"]:number,g' "$MV_PATH"/index.ts
+  else
+    echo "unknown platform; exiting"
+    exit 1
+  fi
```
topocount, line 43: @exrhizo let me know what you think about this

levity: You could get this from the output of `uname -a` (though this might break Windows? Not a Windows user)

zashton:
> I still need to write the authentication code. Because the serverless function is hosted publicly by vercel, we should auth on a pre-shared environmental secret that'd be passed as a request header. We should probably be doing this on all vercel-hosted actions to ensure we don't see unauthorized db writes. Only hasura should be able to do anything with these functions.

+1 you could add an `action secret` env var and the authorization code in a shared middleware.

scripts/generate-zeus.sh
topocount: I don't think executing bash scripts in windows is a generally supported thing. As I understand it, it's okay to tell Windows devs to use [WSL](https://docs.microsoft.com/en-us/windows/wsl/about) and then do what's needed to support that environment.

topocount: @levity does this script work without issue on your M1 macbook?

src/lib/zod/index.ts
```diff
@@ -14,3 +16,19 @@ export const createCircleSchemaInput = z
...
     'Either Protocol name should be filled in or a Protocol should be selected.'
   );
+
+// currently mirroring the shape of the current endpoint
+// it might be preferable to fold circle_id into the object
+export const createUserSchemaInput = z.object({
+  circle_id: z.number(),
+  object: z.object({
+    name: z.string().min(3).max(255),
+    address: zEthAddress,
```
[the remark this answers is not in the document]
topocount: thanks good catch. Seems like we need a backend version of the zEthAddress parser that doesn't return an ENS address, though we would like to be able to resolve them to the address and then save that, I think

api/hasura/actions/create_user.ts
```diff
@@ -0,0 +1,116 @@
...
+  // I'd prefer to add this uniqueness constraint into the database
+  const { object: createUserParams, circle_id } = result.data;
+
+  const { users: existingUsers } = await gql.q('query')({
+    users: [
+      {
+        limit: 1,
+        where: {
+          name: { _eq: createUserParams.name },
+          address: { _eq: createUserParams.address },
```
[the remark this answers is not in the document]
exrhizo: Ideally we only have lower case, that way our database is consistent

api/hasura/actions/create_user.ts
```diff
@@ -0,0 +1,116 @@
...
+      { object: { ...createUserParams, circle_id } },
+      { id: true },
+    ],
+    // End any active nomination
+    update_nominees: [
+      {
+        _set: { ended: true },
+        where: {
+          circle_id: { _eq: circle_id },
+          address: { _eq: createUserParams.address },
```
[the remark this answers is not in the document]
exrhizo: lets prefer lower case

api/hasura/actions/create_user.ts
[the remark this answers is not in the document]
topocount: yeah the parser lowercases all inputs, so we don't have to do any partial string matches here.

api/hasura/actions/create_user.ts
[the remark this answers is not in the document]
topocount: okay belt and suspenders it is

api/hasura/actions/create_user.ts
```diff
@@ -0,0 +1,104 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+import fetch from 'node-fetch';
+
+import { gql } from '../../../api-lib/Gql';
+import { validateVerificationKey } from '../../../api-lib/validate';
+import { profiles_constraint } from '../../../src/lib/gql/zeusHasuraAdmin';
+import { createUserSchemaInput } from '../../../src/lib/zod';
+
+// @ts-ignore
+global.fetch = fetch;
```
exrhizo, line 10: zashton fixed this by adding it to api-lib/Gql

.env.example
```diff
@@ -20,6 +20,7 @@ PRISMA_DATABASE_URL=postgres://postgres:postgrespassword@localhost:5432/postgres
 # Hasura CLI (used with ./hasura/config.yaml)
 HASURA_GRAPHQL_ENDPOINT=http://localhost:8080
 HASURA_GRAPHQL_ADMIN_SECRET=admin-secret
+HASURA_EVENT_SECRET=event-secret
```
exrhizo, line 23: same as `HASURA_GRAPHQL_ADMIN_SECRET`?

api-lib/validate.ts
```diff
@@ -0,0 +1,29 @@
+import { VercelRequest, VercelResponse, VercelApiHandler } from '@vercel/node';
+
+export const validateVerificationKey = (handler: VercelApiHandler) => {
+  return async (req: VercelRequest, res: VercelResponse) => {
+    try {
+      if (
+        !req.headers.verification_key ||
+        (req.headers.verification_key as string) !==
+          process.env.HASURA_EVENT_SECRET
```
exrhizo, line 9: variables like these to be loaded in `api-lib/config.ts`

api-lib/validate.ts
```diff
@@ -0,0 +1,29 @@
+import { VercelRequest, VercelResponse, VercelApiHandler } from '@vercel/node';
+
+export const validateVerificationKey = (handler: VercelApiHandler) => {
```
exrhizo, line 3: name it something with event, maybe `verifyHasuraAdminMiddleware`

@levity ?

src/lib/zod/index.ts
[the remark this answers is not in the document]
exrhizo: An issue with querying infura is that we can run into quota (maybe that wouldn't be a problem in practice?) I feel like we should expect non ens address in our endpoints until in the future if we decide to improve support.

api/hasura/actions/create_user.ts
exrhizo: @zashton I wonder if this is possible for circle with a user in it

scripts/generate-zeus.sh
exrhizo: I have a macbook m1 and it worked

api/hasura/actions/create_user.ts
zashton: yeah possible if its a newly added user that has never login to our system, they would have no profile.

.env.example
topocount: it seems more secure to separate them out, since what we are authenticating here is a bit different.

hasura/metadata/actions.yaml
```diff
@@ -1,15 +1,29 @@
 actions:
+- name: createUser
+  definition:
+    kind: synchronous
+    handler: '{{HASURA_ACTION_BASE_URL}}/create_user'
+    forward_client_headers: true
+    headers:
+    - name: verification_key
+      value_from_env: HASURA_EVENT_SECRET
+  comment: creates a user in a circle and creates a profile if none exists
```
zashton, line 10: missing permissions for role - user, superadmin. 
to be able to trigger this mutation. 
For this also i wonder if you are able to add an optional condition to the middleware, only users of `role` = 1 of the same `circle_id` are allowed to use this mutation. A couple of other mutations (to be migration) would also require this authentication e.g updating users/ circles

api-lib/config.ts
```diff
@@ -31,3 +31,7 @@ export const LOCAL_SEED_ADDRESS = <string>(
   getEnvValue('LOCAL_SEED_ADDRESS', 'missing')
 );
 export const IS_LOCAL_ENV = process.env.NODE_ENV === 'development';
+
+export const HASURA_EVENT_SECRET = <string>(
+  getEnvValue('HASURA_EVENT_SECRET', '')
```
exrhizo, line 36: I don't think this should have a default value, because the vercel should crash without this

api/hasura/actions/create_user.ts
```diff
@@ -0,0 +1,100 @@
...
+        message: 'invalid input',
+        code: 422,
+      });
+  }
+
+  // External Constraint Validation
+  // I'd prefer to add this uniqueness constraint into the database
+  const { object: createUserParams, circle_id } = result.data;
+
+  const { users: existingUsers } = await gql.q('query')({
```
exrhizo, line 27: eventually we will want to have a nice library for all these things, although if it is unlikely to be reused it might as well be kept in the action

src/forms/formHelpers.ts
```diff
@@ -17,17 +17,24 @@ export const zStringISODateUTC = z
...
 export const zEthAddress = z
   .string()
+  // `resolveName` throws outright if it is passed a malformed eth address
+  // so we need to refine it first so safeParseAsync doesn't throw
+  // unexpectedly
+  .refine(validAddressOrENS, 'Wallet address is invalid')
   .transform(s => provider.resolveName(s))
   .transform(s => s || '')
-  .refine(s => ethers.utils.isAddress(s), 'Wallet address is invalid')
+  .refine(s => s, 'unresolved ENS name')
```
exrhizo, line 26: the `s => s | ''` insures this never happens, I think these don't make sense now..
Maybe this one we call `zEthAddressEns` and the other `zEthAddress`

META-DREAMER:
@topocount I'm wondering why the Hasura Preview action didnt run on this PR: https://github.com/coordinape/coordinape/blob/main/.github/workflows/hasura-preview.yml

Is your branch up to date with main?

api/hasura/actions/create_user.ts
topocount: once we get a uniqueness contraint on the users table this particular query should be unnecessary

api/hasura/actions/create_user.ts
```diff
@@ -0,0 +1,100 @@
...
+
+import { gql } from '../../../api-lib/Gql';
+import { verifyHasuraAdminMiddleware } from '../../../api-lib/validate';
+import { profiles_constraint } from '../../../src/lib/gql/zeusHasuraAdmin';
+import { createUserSchemaInput } from '../../../src/lib/zod';
+
+async function handler(request: VercelRequest, response: VercelResponse) {
+  // Input Validation
+  const result = await createUserSchemaInput.safeParseAsync(request.body.input);
+  // a `switch` gets typescript to behave soundly
```
crabsinger, line 11: "to behave soundly"... what's going on here? Is the `if(result.success) == true` not working?

topocount: @crabsinger the code you commented is good js, but fails the typescript typechecker.


NBD because @exrhizo proposed some syntax that works as expected above.

topocount:
> @topocount I'm wondering why the Hasura Preview action didnt run on this PR: https://github.com/coordinape/coordinape/blob/main/.github/workflows/hasura-preview.yml
> 
> Is your branch up to date with main? maybe something up with the path filter although I did test it to be working when I set it up.

@META-DREAMER do you know why this is failing now

api-lib/circleAdmin.ts
```diff
@@ -0,0 +1,36 @@
...
+      if (role !== 1) {
+        return res.status(401).json({
+          message: 'User not circle admin',
+          code: 401,
+        });
+      }
+    }
+
+    await handler(req, res);
+  };
```
topocount, line 33: @zashton here's the circle auth middleware

META-DREAMER: instead of assuming it's admin, can we check the headers to verify it explicitly? Hasura should pass the auth/role info to the handler: https://hasura.io/docs/latest/graphql/core/remote-schemas/auth/remote-schema-forward-auth.html#passing-headers-from-hasura-to-your-remote-schema

.env.example
crabsinger, line 23: I'm assuming this needs to be set in staging/production? If so, can you make a TODO in the PR of what needs to happen before/post deploy?

api-lib/circleAdmin.ts
```diff
@@ -0,0 +1,36 @@
...
+import { z } from 'zod';
+
+import { getUserFromAuthHeader } from './findUser';
+import { verifyHasuraAdminMiddleware } from './validate';
+
+const middleware =
+  (handler: VercelApiHandler) =>
+  async (req: VercelRequest, res: VercelResponse) => {
+    // skip auth check if no header is passed
+    // because we can assume the user is a hasura-admin
```
crabsinger, line 17: comment no longer relevant, seems you added the check below.

api-lib/circleAdmin.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (req.headers.authorization) {
+      const circleId = z.number().parse(req.body.input.circle_id);
+      const { role } = await getUserFromAuthHeader(
+        req.headers.authorization,
+        circleId
+      );
+      if (role !== 1) {
```
crabsinger, line 24: small nit: I usually find logic patterns that are "if X then Y" easier to reason about and understand than "if not X, then not Y".

I can imagine it being nice to have a few helper functions here around the roles, so we could do something like this:

```
if role(admin) { 
do action
} else {
raise 401, unauthorized.
}
```

but don't need to refactor this right now, we can do it once we see a broader pattern clearly in a few of these PRs.

hasura/metadata/actions.yaml
```diff
@@ -1,15 +1,32 @@
...
+- name: createUser
+  definition:
+    kind: synchronous
+    handler: '{{HASURA_ACTION_BASE_URL}}/create_user'
+    forward_client_headers: true
+    headers:
+    - name: verification_key
+      value_from_env: HASURA_EVENT_SECRET
+  permissions:
+  - role: user
```
crabsinger, line 11: https://github.com/coordinape/coordinape/pull/508/files#diff-f3e37af54b5b154e8a705d0f72ed476d9b054ed166a0f8fcf25aece0d8ba4e7cR24 seems to only allow this for admins, but we also allow this action for `role: user` too ?

crabsinger (commented):
looks good - a few thoughts.

api-lib/circleAdmin.ts
topocount: thank you! I was confused when I couldn't find these in the http headers.

hasura/metadata/actions.yaml
topocount: Yeah, some users are circle admins, but they don't carry any special privileges above user in hasura, since anyone can become a circle admin simply by creating a circle.

We have a middleware (that I added today) that checks the requesting user is a circle admin for the circle they're attempting to modify, and that's how we're apparently gating this logic after some discussion with @exrhizo @META-DREAMER and @zashton in the dev core channel.

api-lib/circleAdmin.ts
topocount: sure, I bias towards "erroring early" but I definitely appreciate the desire to eliminate another thing that needs to be loaded into one's mental stack.

topocount: i've created functions that describe the logic in words, since there are multiple checks now.

api-lib/circleAdmin.ts
```diff
@@ -0,0 +1,57 @@
...
+        .transform(Number.parseInt)
+        .parse(sessionVariables['x-hasura-user-id']);
+      const { role } = await getUserFromProfileId(profileId, circleId);
+      if (isNotCircleAdmin(role)) {
+        res.status(401).json({
+          message: 'User not circle admin',
+          code: 401,
+        });
+        return;
+      }
```
zashton, line 34:
```suggestion
      try {
        const { role } = await getUserFromProfileId(profileId, circleId);
        if (isNotCircleAdmin(role)) {
          res.status(401).json({
            message: 'User not circle admin',
            code: 401,
          });
          return;
        }
      }
      catch(e) {
        res.status(401).json({
          message: 'User not circle admin',
          code: 401,
        });
        return;
      }
```

zashton: Think you might need to do a try catch here, if user doesn't exist in circle might throw an error

zashton (dismissed):
middleware looks good, just a small suggestion regarding error catching. 👍

zashton (approved):

META-DREAMER (approved):
Few more small things, main thing would be talking to the DB through Hasura vs using Prisma

hasura/metadata/actions.graphql
```diff
@@ -12,6 +19,17 @@ input create_circle_input {
   protocol_id: Int
 }
 
+input createUserInput {
+  address: String!
```
META-DREAMER, line 23: We shouldn't allow the user to specify the address in the mutation. Instead we should set the ETH address of the authenticated user in the auth handler as `X-Hasura-Eth-Address` and extract the value in this action handler. That way its impossible to create a user with someone elses Eth address.

META-DREAMER (changes requested):
Just realized something about passing address field in the mutation vs getting it from auth context (thanks for the callout @CryptoGraffe)

Should fix it before merging since its a security issue

EDIT: Just saw Zashton posted a fix here: https://github.com/coordinape/coordinape/pull/508#discussion_r803123126

Would be good to rebase this on that

src/lib/zod/index.ts
```diff
@@ -16,3 +16,19 @@ export const createCircleSchemaInput = z
...
+    name: z.string().min(3).max(255),
+    address: zEthAddressOnly,
+    non_giver: z.boolean().optional(),
+    starting_tokens: z.number().optional(),
+    give_token_remaining: z.number().optional(),
+    fixed_non_receiver: z.boolean().optional(),
+    non_receiver: z.boolean().optional(),
+    role: z.number().min(0).max(1).optional(),
+  }),
+});
```
META-DREAMER, line 34: We should also set this to [strict](https://github.com/colinhacks/zod#strict) so Zod invalidates it if it contains any extra fields we don't expect the user to be able to set / modify.

topocount:
@META-DREAMER I'm not sure what's wrong with the hasura metadata. There's no diff between the old config that I've been building off and the current config post-rebase.

exrhizo (dismissed):

META-DREAMER:
> @META-DREAMER I'm not sure what's wrong with the hasura metadata. There's no diff between the old config that I've been building off and the current config post-rebase.
> 
> 
> 
> It also applies successfully locally
> 
> 
> 
> edit: a retry made it work
> 
> 

Yea it seems like their APIs have been flaky for some reason, not an issue on our side.

META-DREAMER (dismissed):

exrhizo (approved):
I put the env variables in vercel, for:
`HASURA_GRAPHQL_ADMIN_SECRET`
and
`HASURA_EVENT_SECRET`
the two event secret values were randomly generated by my mac `/dev/urandom`

PR #514, opened 2022-02-07
Jest x Hardhat part 1: Ganache in CI

levity:
This PR contains work to set up the codebase to allow writing frontend integration tests that connect to a local testchain. This is part of the work for #527 

As a result of the changes in this PR, running `yarn test:ci` will start up a Ganache instance with a mainnet fork, run the hardhat tests against it, run the Jest tests (none of which use the testchain yet), and then shut down the Ganache process.


levity:
The main thrust of the work so far has been to make the test environment less specific to hardhat's built-in testchain, and work with ganache as well. My expectation is that ganache will be easier to wrangle into starting and stopping automatically around a Jest test suite, because I've done exactly that before. (see e.g. [makerdao/testchain](/makerdao/testchain))

hardhat/utils/unlockSigner.ts
```diff
@@ -2,9 +2,26 @@ import { Signer } from 'ethers';
...
+    await provider.request({ method: 'evm_addAccount', params: [address, ''] });
+    await provider.request({
+      method: 'personal_unlockAccount',
+      params: [address, '', 0],
+    });
+
+    // we do this because Hardhat's provider code doesn't allow
+    // the account to be used even after it's unlocked
+    const newProvider = new ethers.providers.JsonRpcProvider(
+      // FIXME don't hardcode this!
```
levity, line 16: FIXME

hardhat/scripts/start-ganache.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# read .env, filtering out comments
+export $(cat ../.env | sed 's/^#.*$//' | xargs)
+
+exec ./node_modules/.bin/ganache \
+  -p 8555 \
+  -m coordinape \
+  -f $ETHEREUM_RPC_URL \
+  --fork.blockNumber $HARDHAT_FORK_BLOCK \
+  --miner.defaultGasPrice 0x7735940000
```
levity, line 11: this is 100x the default price... was getting errors:
```
Transaction's maxFeePerGas (2000000000) is less than the block's baseFeePerGas (129277809532)
```

hardhat/scripts/start-ganache.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# read .env, filtering out comments
+export $(cat ../.env | sed 's/^#.*$//' | xargs)
+
+exec ./node_modules/.bin/ganache \
+  -p 8555 \
```
levity, line 7: TODO: parameterize this... the idea is to avoid using the default port, so a developer can be using one testchain instance for persistent local data, & running another one on a different port just for tests

crabsinger:
@levity curious your thinking of Ganache vs just using Hardhat's internal default network? I've had a bit of buggyness with Ganache, and seemed to see a number of folks experiencing Hardhat's network as more reliable, although without the UI.

levity:
> 

I used ganache extensively at Maker & it was pretty reliable after some wrangling. 

Hardhat has been pretty decent, but once you stray from its default usage scenario, it becomes unclear how to do things. e.g. I want to be able to run two separate instances simultaneously, so that a developer can use one with persistent data for their local dev, and one that's constantly getting modified and reset as automated testing runs in the background. With hardhat, it's not even documented how to change the port number for its internal node, for example.

levity:
@exrhizo @crabsinger tagged you as reviewers just in case you can get to it soon, but don't worry about it if you have other priorities

package.json
```diff
@@ -82,7 +82,8 @@
     "hardhat:install": "yarn --cwd hardhat install",
     "hardhat:test": "yarn --cwd hardhat test",
     "test": "craco test",
-    "test:ci": "CI=1 craco test --coverage && yarn hardhat:test",
+    "test:all": "yarn hardhat:test && CI=1 craco test --coverage",
+    "test:ci": "./hardhat/scripts/start-ganache.sh --exec yarn test:all",
```
crabsinger, line 86: Can we update the README to include these commands now?

I see it currently has https://github.com/coordinape/coordinape/blob/main/README.md#hardhat with a few commands (that errored for me) but these aren't there.

.env.example
```diff
@@ -1,6 +1,8 @@
 # Shared Environment
 LOCAL_SEED_ADDRESS=0xYOUR_DEV_ADDRESS
 HARDHAT_OWNER_ADDRESS=0xYOUR_DEV_ADDRESS
+HARDHAT_FORK_BLOCK=13500000
+HARDHAT_GANACHE_PORT=8555
 REACT_APP_INFURA_PROJECT_ID=infuraidhere
```
crabsinger, line 6: I'm getting this error: 
"Starting RPC server
CodedError: project ID does not have access to archive state"

which makes sense because I passed in a free Infura ID. I assume the forking requires "archive" data access which will require a paid INFURA id. 

Anyhow, maybe we document this more clearer, and maybe there's a way to run some of the tests without needing the fork?

crabsinger: https://infura.io/docs/ethereum/add-ons/archiveData

crabsinger:
> 



> @crabsinger I used ganache extensively at Maker & it was pretty reliable after some wrangling.
> 
> Hardhat has been pretty decent, but once you stray from its default usage scenario, it becomes unclear how to do things. e.g. I want to be able to run two separate instances simultaneously, so that a developer can use one with persistent data for their local dev work, and one that's constantly getting modified and reset as automated testing runs in the background. With hardhat, it's not even documented how to change the port number for its internal node, for example.

makes sense - thanks for the rationale!

crabsinger (commented):
Code looks good - made a few comments. I was unable to get the 

`❯ yarn test:ci`
`❯ yarn test:all`

commands to succeed (but I think that's related to not having an Infura ID with archival data enabled?

hardhat/hardhat.config.ts
```diff
@@ -94,9 +96,19 @@ task('mint', 'Mints the given token to specified account')
...
+
+const hardhatNetwork = {
+  live: false,
+  allowUnlimitedContractSize: true,
+  gas: 'auto' as const,
+  gasPrice: 'auto' as const,
+  gasMultiplier: 1,
+  chainId: 1337,
+  accounts: {
+    mnemonic: 'coordinape',
```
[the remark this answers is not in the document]
levity: We're not using them for anything atm. The most important thing is to set HARDHAT_OWNER_ADDRESS, because the test deployment scripts will fund that address with tokens.

levity:
> Code looks good - made a few comments. I was unable to get the
> 
> `❯ yarn test:ci` `❯ yarn test:all`
> 
> commands to succeed (but I think that's related to not having an Infura ID with archival data enabled?

Yep! You can also try signing up for [Alchemy](https://www.alchemy.com/pricing); their free plan has archive data

.env.example
levity: As-is, the tests are pretty dependent on the fork. I'll explore reducing this dependency in the future.

levity: Added a line in the README about needing access to archive data

crabsinger:
> > Code looks good - made a few comments. I was unable to get the
> > `❯ yarn test:ci`
> > `❯ yarn test:all`
> > commands to succeed (but I think that's related to not having an Infura ID with archival data enabled?
> 
> Yep! You can also try signing up for [Alchemy](https://www.alchemy.com/pricing); their free plan has archive data

Alchemy - ah, I'll get one of those keys. Maybe useful to change the name from INFURA_* to ALCHEMY_* or document this somewhere. @levity

PR #515, opened 2022-02-07
Hides "My Team" path when team selection is false

### LINKED ISSUE #377: Hide "My Team" link in avatar menu when team selection is disabled
**Problem**
When team selection is disabled for the current circle, the "My Team" link in the avatar menu (upper right) just loads the same page as the "My Epoch" link.

**Solution**
Remove the "My Team" link when team selection is disabled for the current circle.

**Implementation**
See MyAvatarMenu.tsx, MenuNavigationLinks.tsx, routes/paths.ts. The code that currently renders the menu links does not check the current circle, so this check will have to be added, using the useSelectedCircle hook. See CirclesHeaderSection.tsx for an example of this.

f1lander:
fixes #377

### DIFF
--- a/src/components/MenuNavigationLinks/MenuNavigationLinks.tsx
+++ b/src/components/MenuNavigationLinks/MenuNavigationLinks.tsx
@@ -34,36 +34,42 @@ const useStyles = makeStyles(theme => ({
   },
 }));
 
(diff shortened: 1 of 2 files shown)


exrhizo (approved):
Thank you for getting this!

PR #526, opened 2022-02-09
fix spoofing address issue for create circle

zashton:
- remove address as an input

@CryptoGraffe


api-lib/config.ts
```diff
@@ -31,3 +31,7 @@ export const LOCAL_SEED_ADDRESS = <string>(
   getEnvValue('LOCAL_SEED_ADDRESS', 'missing')
 );
 export const IS_LOCAL_ENV = process.env.NODE_ENV === 'development';
+
+export const COORDINAPE_USER_ADDRESS = <string>(
+    getEnvValue('COORDINAPE_USER_ADDRESS', '')
```
exrhizo, line 36: dont add the '' because this should cause a fail if that isn't available

exrhizo (commented):
This makes sense to me

api/hasura/auth.ts
```diff
@@ -39,6 +39,7 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
     res.status(200).json({
       'X-Hasura-User-Id': tokenRow.tokenable_id.toString(),
       'X-Hasura-Role': profile.admin_view ? 'superadmin' : 'user',
+      'X-Hasura-Address': profile.address,
```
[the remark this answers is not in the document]
META-DREAMER: Yea this is good, we can use it like that. It can work for admin-secret too, just have to specify `X-Hasura-User-Id` in addition to the admin secret and it will allow the admin to "act like" a given user.

api-lib/config.ts
zashton: updated

topocount (dismissed):
I understand why this is needed for this endpoint, but don't believe it is needed for #508.

api/hasura/actions/create_circle.ts
```diff
@@ -36,7 +37,8 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
     const ret: ValueTypes['create_circle_response'] =
       await gql.insertCircleWithAdmin(
         input,
-        process.env.COORDINAPE_USER_ADDRESS
+        req.body.session_variables['x-hasura-address'],
```
topocount, line 40: I wrote a composable request body parser in #508. Can this be rebased on top of that with the session variable zod parser updated?

topocount: it's here: https://github.com/coordinape/coordinape/blob/c3d60f85c48045820557ebe00d3dfcb751dcb4f7/src/lib/zod/index.ts#L42-L64

zashton: Sure

zashton:
> @zashton actually can we include the api folder in `tsconfig.json`? I tried doing it in my branch but was getting errors from the create_circle function when running `npx tsc --noEmit`, which is a failure in Ci

saw the error , let me try fixing it

src/lib/zod/index.ts
```diff
@@ -5,14 +5,78 @@ import { zEthAddressOnly } from '../../../src/forms/formHelpers';
 export const createCircleSchemaInput = z
   .object({
     user_name: z.string().min(3).max(255),
-    address: zEthAddressOnly,
     circle_name: z.string().min(3).max(255),
     protocol_id: z.number().int().positive().optional(),
     protocol_name: z.string().min(3).max(255).optional(),
   })
+  .strict()
   .refine(
```
zashton, line 13: @topocount any workaround how i can make the refine work with your zod parser ? fails on CI

topocount: Sorry, the changes merged to main should work, if you rebase. I updated the type to accept everything except `passthrough` zod parsers.

topocount (approved):

PR #528, opened 2022-02-09
Set X-Hasura-Eth-Address in auth context

META-DREAMER:
Allows us to verify / prepopulate Eth addresses in Action handlers / mutations when we want to ensure the user can't take actions using someone elses ETH address.


META-DREAMER:
Closing since I just noticed #526

PR #533, opened 2022-02-11
Feature/469 enable distribute form

### LINKED ISSUE #469: Build (Vaults) Submit Distribution: Part 2.1 Update Allocations Table
### Specification 

- the contents of the table should reflect the form selection in part 1, #468 
    - the token type in the last column will be determined by the vault choice
    - the token amounts in the last column will be determined by the vault choice & the total amount
      - update amounts when the total amount input is blurred

teeolendo:
- Enable form dynamic calculation as part of closes #469


src/pages/DistributePage/AllocationsTable.tsx
```diff
@@ -30,10 +44,10 @@ const AllocationTable = ({ users }: { users: IUser[] }) => {
         {
           label: 'Name',
           accessor: 'name',
-          render: function UserName(u: IUser) {
+          render: function UserName(u: IAllocateUser) {
             return (
               <div className={classes.avatarCell}>
-                <ApeAvatar user={u} className={classes.avatar} />
+                <NewApeAvatar name={u.name} className={classes.avatar} />
```
levity, line 50: Good opportunity here to switch this over to using Stitches. Pretty close to having this whole file have no dependency on material-ui

src/components/NewApeAvatar/NewApeAvatar.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
levity, line 1: is this still a WIP? it looks similar enough to ApeAvatar that maybe it would be better to change that to be able to handle simple props instead of user/profile objects?

src/pages/DistributePage/AllocationsTable.tsx
```diff
@@ -44,26 +58,41 @@ const AllocationTable = ({ users }: { users: IUser[] }) => {
...
-          render: () => '22',
+          render: (u: IAllocateUser) =>
+            u.received_gifts_aggregate?.aggregate?.count && '-',
         },
         {
           label: '% of Epoch',
-          render: () => '2.5%',
+          render: (u: IAllocateUser) =>
+            u.received_gifts.length > 0
+              ? `${((u.received_gifts[0].tokens / totalGive) * 100).toFixed(
```
levity, line 77: this and below should be summing `tokens` from all gifts

levity (changes requested):

levity:
I'm noticing that the table doesn't seem to reliably update when I change the amount in the text field. Maybe there's some overly-aggressive caching going on in the AllocationsTable?

src/pages/DistributePage/DistributePage.tsx
```diff
@@ -183,11 +197,16 @@ const DistributePage = () => {
           </Box>
         </Box>
         <Box css={{ m: '$lg' }}>
-          <AllocationTable users={users} />
+          <AllocationTable
+            users={users as IAllocateUser[]}
+            totalAmountInVault={updateAmount as number}
+            tokenName={selectedVault}
+            totalGive={totalGive as number}
```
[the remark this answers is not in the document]
teeolendo: Because the received gifts can be undefined, we'd have to cast it in this way. I could be wrong though but I think it goes back to the database data type from Hasura generates the types.

src/hooks/gql.ts
```diff
@@ -45,8 +46,47 @@ export function useCurrentOrg() {
...
+      {
+        id: true,
+        ended: true,
+        circle_id: true,
+        number: true,
+        circle: {
+          id: true,
+          name: true,
+          users: [
+            { where: { received_gifts: { epoch_id: { _eq: epochId } } } },
```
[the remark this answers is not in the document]
teeolendo: I think we shouldn't submit users who don't have gifts but, as for presentation, I'd suggest an order by Give so that you can use who's getting the most and then it tapers off naturally. I'm thinking about this in the context of a team member that has an undervalued contribution, just good for them to know when they are submitting so there are no surprises.

src/hooks/gql.ts
```diff
@@ -45,8 +46,47 @@ export function useCurrentOrg() {
   }).data?.organizations_by_pk;
 }
 
-export function useCircleIdForEpoch(epochId: number) {
+export function useEpochIdForCircle(epochId: number) {
```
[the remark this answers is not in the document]
teeolendo: I'm gonna compromise and move it to the same folder as DistributePage.tsx. I think code is more readable when side effects are kept in separate files.

src/components/NewApeAvatar/NewApeAvatar.tsx
teeolendo: My only issue leaving this here is scope. I just created a ticket here https://github.com/coordinape/coordinape/issues/545 that specifies what should be done. I might finish later today.

src/pages/DistributePage/DistributePage.tsx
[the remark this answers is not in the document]
levity: It would be safest to define `totalGive` such that even in edge cases like that, it defaults to a number value like 0

levity:
> I'm noticing that the table doesn't seem to reliably update when I change the amount in the text field. Maybe there's some overly-aggressive caching going on in AllocationsTable?

I might have caused this myself with some of the changes to AllocationsTable; not sure 😅 gotta run to dinner now though

teeolendo:
> > I'm noticing that the table doesn't seem to reliably update when I change the amount in the text field. Maybe there's some overly-aggressive caching going on in AllocationsTable?
> 
> I might have caused this myself with some of the changes to AllocationsTable; not sure 😅 gotta run to dinner now though

No, I don't think so. I think it was kinda wonky before. I have a few changes to push so let's see how that goes.

levity (approved):

src/pages/DistributePage/AllocationsTable.tsx
levity: ^ attn @f1lander

levity (approved):

teeolendo (commented):
I'll refactor this then.

PR #534, opened 2022-02-11
Create a body factory that accepts user session variables

topocount:
Sometimes, you don't want to automatically parse or validate admin
requests in a Hasura Action.

I tried to implement this case in the existing body parser factor but
ZodUnions don't easily compose statically at the parameter level. I
have a feeling discriminated unions will play much nicer.


CryptoGraffe (approved):
lg2m as far as I can understand it. We pass in the desired schema as a generic now and get it nicely typed

Issue #518, opened 2022-02-08
Migrate v2/{circle_id}/admin/{address}: Update Circle User (Admin)


zashton:
**Middleware**
- Validate user is circle admin of updating user's circle

**Additional Validations**
- Verify if address is updated - no duplicates in circle
- Block the update if `starting_tokens` is changed and there is an active epoch ongoing

**Logic**

Input: `name`, `address`,`starting_tokens`, `non_giver`, `fixed_non_receiver`,`non_receiver`,`role`
- Check If user is forced set to opt out from opt in, **Purge all `pending_gifts` rows with no notes received by the user. Set all gifts that has notes with token value to 0.** Set user's `tokens_received` to 0. Iterate through all users that sent tokens to this user and add back the tokens to their `tokens_remaining`
- If address of user is changed validate if a profile with that address exists, if not create an empty profile for that address.
- if `starting_tokens` is changed update `give_token_remaining` to the same value
- **Check if admin set `non_giver` from `0` to `1`. Purge all `pending_gifts` rows with no notes sent by the user, Set all gifts that has notes with token value to 0. Iterate through all users that received tokens from this user and deduct the tokens from their tokens_received**

**DB Tables**
`users`, `pending_token_gifts`

**Notes**
Notifications
- Opt out notification via Telegram/Discord (if set) to notify respective channel/groups with refunded total to specific users.
Once you are opted out all the tokens you received will be refunded. The notification will indicate how much of your tokens was refunded + (breakdown of all those senders amount). if you didn't receive any then it will just indicate 0 is refunded.
e.g
$name opted out of the current epoch!
A total of 30 GIVE was refunded
 senderName: 10
 senderName2: 20

Use event triggers for the sending of notifications, so it doesn't block the mutation if the service fails and has an retry mechanism.

`circles` table: `telegram_id` and `discord_webhook` are the 2 fields used for notifications.
`telegram_id` would be the `chat_id` of your chatgroup/channel. your telegram bot would need to be an admin to receive webhook updates from telegram

PR #538, opened 2022-02-11
Implement AdminUpdateUser Hasura Endpoint

topocount:
In addition to implementing the endpoint I've made the following
changes:

1. Refactored optional profile creation (as a part of other mutations)
   into an event trigger based on user creation or updates
2. Created composable discord and telegram factories that work with
   logic handlers for event triggers that emit bot messages, based on
   @jsur's existing work. I'm considering combining the factories into
   a single function and then allowing the triggers to emit the variable
   that determines whether or not discord or telegram is the target.
3. Extended @jsur's event trigger payload type into a dynamic type
   that changes based on the table and type of operation on the table
   that emits the event.

# Test Plan

1. Follow the README instructions to set up the environment
2. find the next epoch in the database, if one isn't already active
and adjust the start date to before the current date. Take note of the
circle
3. Allocate some give to members of the Circle in the UI.
4. Set up a discord webhook for the circle and add a telegram channel id
in the circle's database entry
5. pick a user in the circle give has been allocated to and execute the
following mutation in the graphiql interface:
```gql
mutation UpdateUserTest {
  adminUpdateUser(
    object: {
      address: "<recipient's address>",
      circle_id: <recipient circleID>,
      non_receiver: true
    }
  ) {
    address
  }
}
```

6. verify that discord and telegram entries are created for the opt-out
   and the refund. The total refunded from the opted-out user will
   likely be incorrect due to the mock data in the database, but refund
   to the user you allocated from should be correct.
7. Notice that a new profile was created with a lower-case version of
   the user's address, since the existing address is checksum-mixed
   case. This can be verified in the create-profile event trigger's
   "processed events" tab.
8. run other queries and verify that fixed_non_receiver and non_receiver
   behave correctly as inputs together, along with give_token_remaining.

# TODO
- [x] run a couple more tests and ensure error responses are sensible
- [x] implement telegram and discord bot hooks.
- [x] convert profile check and creation into an event trigger
- [x] configure and test event triggers in hasura
- [x] Seems like @jsur has made some progress on bot comms in hasura so I'll try to borrow from their work.

resolves https://github.com/coordinape/coordinape/issues/518


api-lib/handleRefundGiveMsg.ts
```diff
@@ -0,0 +1,32 @@
+import assert from 'assert';
```
CryptoGraffe, line 1: I'm not expert with `assert` in this environment, but usually in other frameworks it's only for things you really don't expect to happen. I bring it up not exactly because It's used here but because I've seen it in other project code. i'm not sure if there is a better pattern here, especially wrt HTTP errors. In this case, it does seem really unexpected, but perhaps we could wrap it with something that that could take an http error, e.g `assertWithStatus(users_by_pk, 500)` (or 404).

api-lib/handleRefundGiveMsg.ts
```diff
@@ -0,0 +1,32 @@
...
+
```
CryptoGraffe, line 2: We should probably figure out how to standardize filenames for this project, I really don't have a strong opinion but we have snake case for actions, then mixed lowerCamelCase and kebab-case in this package, and then CamelCase for types/classes (which I think is good). Anyway I'm open to ideas but a style guide might be nice. I've also struggled a bit with the db columns vs json vs var names. I have been using snake for db and lowerCamel for vars.

src/lib/gql/GqlHasuraAdmin.ts
```diff
@@ -87,6 +81,26 @@ export class Gql {
     ).circles;
   }
 
+  async getCurrentEpoch(
```
CryptoGraffe, line 84: Is there something about this that is admin only?

hasura/metadata/actions.yaml
```diff
@@ -39,8 +51,9 @@ custom_types:
   enums: []
   input_objects:
   - name: create_circle_input
-  - name: createUserInput
   - name: upload_image_input
+  - name: CreateUserInput
```
CryptoGraffe, line 55: mixed casing very obv here

hasura/metadata/actions.yaml
```diff
@@ -1,4 +1,15 @@
 actions:
+- name: adminUpdateUser
+  definition:
+    kind: synchronous
+    handler: '{{HASURA_API_BASE_URL}}/actions/admin_update_user'
+    forward_client_headers: true
+    headers:
+    - name: verification_key
+      value_from_env: HASURA_EVENT_SECRET
```
CryptoGraffe, line 9: this isn't unique to this PR but is there some way we can do this automatically for all actions/events? also the env var has EVENT in the name which is a little misleading

hasura/metadata/actions.graphql
```diff
@@ -24,12 +30,13 @@ type Mutation {
...
 input create_circle_input {
   user_name: String!
+  address: String!
   circle_name: String!
   protocol_name: String
   protocol_id: Int
 }
 
-input createUserInput {
+input CreateUserInput {
```
CryptoGraffe, line 39: this casing seems intentional, but most of the custom inputs are not cased like this, they are snakey. again I don't really care which (i don't have a great basis) but i'm fond of consistency

api/hasura/event-triggers/refund-pending-gift.ts
```diff
@@ -0,0 +1,66 @@
...
+      users_by_pk: [
+        { id: sender_id },
+        {
+          give_token_remaining: true,
+          circle_id: true,
+          name: true,
+        },
+      ],
+    });
+    assert(users_by_pk, `dangling give without user: ${sender_id}`);
```
CryptoGraffe, line 27: do we have something like sentry or bugsnag set up for this kind of stuff? does it work with the vercel serverless functions? it would be nice.

api/hasura/event-triggers/refund-pending-gift.ts
```diff
@@ -0,0 +1,66 @@
+import assert from 'assert';
+
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { gql } from '../../../api-lib/Gql';
+import { EventTriggerPayload } from '../../../api-lib/types';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
```
CryptoGraffe, line 9: I don't have a lot of context here so I would love a comment describing under what circumstances this is called and what is supposed to happen. I can grok it from the code but it's nuanced enough I think some comments are warranted.

api/hasura/event-triggers/opt-out-telegram.ts
```diff
@@ -0,0 +1,7 @@
+import handleOptOutMsg from '../../../api-lib/handleOptOutMsg';
+import makeTelegramEventHandler from '../../../api-lib/make-telegram-event';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+
+const handler = makeTelegramEventHandler(handleOptOutMsg);
+
+export default verifyHasuraRequestMiddleware(handler);
```
CryptoGraffe, line 7: nice how small these trigger are

api/hasura/event-triggers/create-profile.ts
```diff
@@ -0,0 +1,54 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { gql } from '../../../api-lib/Gql';
+import { EventTriggerPayload } from '../../../api-lib/types';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+import { profiles_constraint } from '../../../src/lib/gql/zeusHasuraAdmin';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
```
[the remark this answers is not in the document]
CryptoGraffe: can you give some more details here @zashton ?

api/hasura/event-triggers/create-profile.ts
```diff
@@ -0,0 +1,54 @@
...
+          },
+        ],
+      });
+      if (insert_profiles_one) {
+        res.status(200).json({
+          message: `profile created for ${insert_profiles_one.address}`,
+        });
+        return;
+      }
+      res
```
CryptoGraffe, line 42: I'm new to gql - is this how you know an insert had a pk constraint violation? it just returns undefined (rather than throw) ?

api/hasura/actions/admin_update_user.ts
```diff
@@ -0,0 +1,175 @@
...
+                epoch_id: { _eq: currentEpoch.id },
+                recipient_id: { _eq: user.id },
+              },
+            },
+            { __typename: true },
+          ],
+        });
+      }
+    }
+
```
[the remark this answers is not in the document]
CryptoGraffe: I was wondering about how transactions worked in gql, i could use one in the vouching stuff i'm working on. It's not clear to me how you pass in multiple , so i'm interested in how this progresses

api-lib/types.d.ts
```diff
@@ -0,0 +1,75 @@
...
+
+type OperationTypes = 'INSERT' | 'UPDATE' | 'DELETE' | 'MANUAL';
+
+// This is not an exhaustive list of all object and array relation types,
+// and should be added to
+// as more relationships are encountered in the GQL api.
+// Only scalar types are returned by event triggers, so this is needed
+// to strip out faulty types from the autogenerated zeus types
+//
+// Maybe one day typescript will allow us to pick based on the type of
```
CryptoGraffe, line 11: This seems like a pain! trying to understand it. This led me to some docs that helped me understand TS a little bit better tho. I think Partial can help me with some of my other struggles.

api-lib/make-telegram-event.ts
```diff
@@ -0,0 +1,29 @@
...
+  ) => Promise<boolean>
+) {
+  return async function handler(req: VercelRequest, res: VercelResponse) {
+    try {
+      const sent = await msgHandler(req.body, { telegram: true });
+      res
+        .status(200)
+        .json({ message: `Telegram message ${sent ? 'sent' : 'not sent'}` });
+    } catch (e) {
+      res.status(500).json({
```
CryptoGraffe, line 23: Eventually we should move all these res.status(x).json() errors to use `ErrorResponse` or `ErrorResponseWithStatus`

CryptoGraffe (commented):
I left a bunch of comments but they are all really cosmetic, and are not showstoppers for merging this. I'd be fine reflecting some of the concepts in new tickets.

api/hasura/event-triggers/create-profile.ts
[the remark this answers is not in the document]
zashton: the event trigger runs as a separate thread than the main mutation as its asynchronous, so the profile may get inserted only after the response is sent hence profile would be null for the response
@CryptoGraffe

api/hasura/actions/admin_update_user.ts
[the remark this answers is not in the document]
zashton:
```
const { insert_vouches_one, insert_users_one} = await this.q('mutation')({ 
insert_vouches_one:[ .... ],
insert_users_one:[... ]
})
```
@CryptoGraffe you'll just need to chain them in the same mutation call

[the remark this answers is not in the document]
topocount: > i tried throwing an error here to check if hasura will rollback the `delete_pending_token_gifts` mutation (but it didn't), how about merging these 2 into 1 mutation call ?

Good catch! I'll do this

api/hasura/actions/admin_update_user.ts
```diff
@@ -0,0 +1,175 @@
...
+    const mutationResult = await gql.q('mutation')({
+      update_users: [
+        {
+          _set: {
+            ...input,
+            // reset remaining tokens to starting tokens
+            give_token_remaining:
```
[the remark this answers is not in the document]
topocount: I think i need to modify both of these things, so I'm going to keep the `give_token_remaining` logic and add the `give_token_received` entry

api/hasura/event-triggers/create-profile.ts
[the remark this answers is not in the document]
topocount: @zashton is this a cricitcal business requirement? The logic here is pretty simple regardless, based on my understanding. If the profile is null in the response, then it's a new profile and all that really exists in the database row is the eth address.

hasura/metadata/actions.graphql
```diff
@@ -65,3 +96,14 @@ type update_profile_response {
   id: Int!
 }
 
+type AdminUpdateUserResponse {
```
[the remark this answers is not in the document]
topocount: I just recently figured out how to do this. Will set it up.

api/hasura/event-triggers/refund-pending-gift.ts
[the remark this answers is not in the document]
topocount: I think this is the https://3factor.app/ way to do it. IF we NEED to access this synchronusly, I can refactor it, but I don't see why that would be the case right now.

api/hasura/event-triggers/create-profile.ts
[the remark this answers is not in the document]
levity: > the event trigger runs as a separate thread than the main mutation as its asynchronous, so the profile may get inserted only after the response is sent hence profile would be null for the response

jumping in here after seeing this comment; apologies if i'm missing some context. but in general, it would be good to have a way to delay the response until we're sure that all triggers have completed, so that we can return data in the response that was created/modified by triggers.

api/hasura/actions/admin_update_user.ts
```diff
@@ -0,0 +1,175 @@
...
+              {
+                start_date: true,
+                end_date: true,
+                id: true,
+              },
+            ],
+          },
+        },
+      ],
+    });
```
META-DREAMER, line 71: I can imagine this being reused in other places so probably worth extracting into a separate `queries/getUserInCircle.ts` file

api-lib/make-telegram-event.ts
topocount: Yeah theres a good amount of cleanup to do

src/lib/gql/GqlHasuraAdmin.ts
topocount: yes I should do this

api/hasura/event-triggers/create-profile.ts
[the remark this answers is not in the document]
topocount: > jumping in here after seeing this comment; apologies if i'm missing some context. but in general, it would be good to have a way to delay the response until we're sure that all triggers have completed, so that we can return data in the response that was created/modified by triggers.
> 
> otherwise our client pattern gets more complicated: send request, wait for some time, fetch modified data, wait if it's not available yet, repeat...

@levity I think this is a question of how the new profile db entries are used, then. It doesn't seem like there is a situation where a profile needs to be synchronously accessible, since only a 3rd party creates a profile as a side effect of some other interaction. Therefore this seems like a great candidate for event-based business logic. Please correct me if this is not the case

If something does need a synchronous guarantee, then it should be included in the action logic directly.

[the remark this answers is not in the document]
topocount: Also, is there any plan to migrate to graphQL subscriptions for state in the frontend?

src/lib/zod/index.ts
```diff
@@ -17,8 +17,19 @@ export const createCircleSchemaInput = z
     'Either Protocol name should be filled in or a Protocol should be selected.'
   );
 
-// this shape mirrors the shape of the original rest endpoint
-// it might be preferable to fold circle_id into the object
+export const adminUpdateUserSchemaInput = z
+  .object({
+    circle_id: z.number(),
+    address: zEthAddressOnly,
```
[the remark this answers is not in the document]
topocount: @zashton I'm very eager to enforce uniqueness at the DB level lol

api/hasura/event-triggers/create-profile.ts
topocount: @CryptoGraffe no this is just logging for Hasura. The config of this call "swallows" any conflict errors. See here for more details: 

https://hasura.io/docs/latest/graphql/core/databases/postgres/mutations/upsert.html

api-lib/handleRefundGiveMsg.ts
topocount: This is a succinct way to cause a throw. It's much easier than using an if statement, then generating an error manually.

api-lib/handleRefundGiveMsg.ts
topocount: I'll push a PR with the formatting we discussed today and see what @zashton thinks

api/hasura/event-triggers/refund-pending-gift.ts
topocount: We could/should do something like this: https://docs.sentry.io/platforms/node/

topocount: We could inject it into the hasura validation middleware so it's easy to deploy haha

src/lib/zod/index.ts
```diff
@@ -17,8 +17,20 @@ export const createCircleSchemaInput = z
...
+    new_address: zEthAddressOnly.optional(),
```
topocount, line 24: @zashton I've decided to add a this instead

src/lib/gql/GqlHasuraAdmin.ts
CryptoGraffe: does it have to be in both places?

CryptoGraffe (dismissed):
This all looks good. I'd like the following changes to be followed up on beyond this PR:

1. change asserts to some wrapper we have around assert that has status codes
2. eliminate all the res.status(x) calls with a wrapper around that (success and errors)
3. the casing changes we discussed
4. submit a ticket to hasura about the verification key on every action

api/hasura/actions/admin_update_user.ts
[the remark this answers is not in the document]
zashton: i'm reminded something that was missed out. I've just updated the issue specs
**Check if admin set `non_giver` from 0 to 1. Purge all pending_gifts rows with no notes sent by the user, Set all gifts that has notes with token value to 0**

api/hasura/event-triggers/create-profile.ts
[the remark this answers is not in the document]
zashton: tried this mutation and indeed the profile was null, i'm not sure why `give_token_received` gets set correctly to 0 even though it is also an event trigger. 
```gql
mutation UpdateUserTest {
  adminUpdateUser(object: {address: "<address>", 
    new_address: "<new_address>", 
    circle_id: 1, non_receiver: true}) {
    id
    UserResponse {
      address
      pending_received_gifts{
      	tokens
        note
      }
      give_token_received
      profile {
        address
        id
      }
    }
  }
}

```

hasura/metadata/actions.yaml
```diff
@@ -71,4 +83,14 @@ custom_types:
       type: object
       field_mapping:
         id: id
+  - name: UserResponse
```
zashton, line 86: I think we can just name this `user` relationship

src/lib/zod/index.ts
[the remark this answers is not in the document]
zashton: yeah we should def do that, just tested working

api/hasura/event-triggers/create-profile.ts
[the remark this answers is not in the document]
zashton: > Also, is there any plan to migrate to graphQL subscriptions for state in the frontend?

this would be nice to have.

api/hasura/event-triggers/refund-pending-gift.ts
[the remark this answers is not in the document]
zashton: On second thought, you are right. this mutation wouldn't need this piece of data synchronously. The critical one is the `token_received` which is already updated. I'm good with this 👍

api-lib/handleRefundGiveMsg.ts
zashton: i changed to camel case for actions/input/response in my latest merged PR 😄

zashton:
@topocount 

For me the only remaining critical issue is the `non_giver` scenario that i've missed in my spec previously.  
To clarify the Opt out definition in the UIs means you cannot receive `tokens` but still can send them. the `non_giver` flag would be the one to restrict you from sending tokens, but still can send notes with no token value.

api/hasura/event-triggers/refund-pending-gift.ts
```diff
@@ -0,0 +1,95 @@
...
+    assert(recipient, `dangling give without user: ${recipient_id}`);
+    const {
+      give_token_remaining: originalGive,
+      name: senderName,
+      non_giver,
+    } = sender;
+
+    const { name: recipientName, non_receiver } = recipient;
+
+    if (tokensToRefund > 0 && (non_giver || non_receiver)) {
```
zashton, line 63: just missing an mutation update  to deduct `give_token_received` from the recipient when switching to `non_giver` of the sender

zashton (changes requested):

src/lib/zod/index.ts
```diff
@@ -17,8 +17,20 @@ export const createCircleSchemaInput = z
...
+    name: z.string().min(3).max(255).optional(),
+    starting_tokens: z.number().optional(),
+    non_giver: z.boolean().default(false),
+    fixed_non_receiver: z.boolean().default(false),
+    non_receiver: z.boolean().default(false),
+    role: z.number().min(0).max(1).default(0),
```
zashton, line 30: i realise i lost my admin status after the update, so this should be optional instead of default

zashton (changes requested):

zashton:
lgt staging once these 2 issues are fixed
@topocount

zashton (approved):
2 issues addressed , lets merge it 👍

zashton:
@topocount should be `hasura-next` branch

PR #540, opened 2022-02-14
Circle Integrations: database table, migrations and permissions

davidfant:
Short demo of what a v1 of a Dework integration that uses circle integrations could look like: https://www.loom.com/share/b1bf242508f647a983c5a8b4c7d7c74e

This PR adds a new table `circle_integrations`. Circle admins can insert and delete integrations. Everyone can read circle integrations.

In a future PR I will make the frontend use this table to create integrations and read/show epoch contributions using integrations

### DIFF
--- a/hasura/metadata/actions.graphql
+++ b/hasura/metadata/actions.graphql
@@ -9,6 +9,7 @@ type Mutation {
     object: create_circle_input!
   ): create_circle_response
 }
+
 input create_circle_input {
   user_name: String!
   circle_name: String!
@@ -43,3 +44,4 @@ type createUserResponse {
   role: Int!
   starting_tokens: Int!
 }
+
--- a/hasura/metadata/actions.graphql
+++ b/hasura/metadata/actions.graphql
@@ -9,7 +9,6 @@ type Mutation {
     object: create_circle_input!
   ): create_circle_response
 }
-
 input create_circle_input {
   user_name: String!
   circle_name: String!
@@ -44,4 +43,3 @@ type createUserResponse {
   role: Int!
   starting_tokens: Int!
 }
-
--- a/hasura/metadata/databases/default/tables/public_circle_integrations.yaml
+++ b/hasura/metadata/databases/default/tables/public_circle_integrations.yaml
@@ -1,0 +1,45 @@
+table:
+  name: circle_integrations
+  schema: public
+object_relationships:
+- name: circle
+  using:
+    foreign_key_constraint_on: circle_id
+insert_permissions:
+- permission:
+    backend_only: false
+    check:
+      circle:
+        users:
+          _and:
+          - role:
+              _eq: 1
+          - address:
+              _eq: X-Hasura-Address
+    columns:
+    - circle_id
+    - data
+    - name
+    - type
+  role: user
+select_permissions:
+- permission:
+    columns:
+    - circle_id
+    - id
+    - data
+    - name
+    - type
+    filter: {}
+  role: user
+delete_permissions:
+- permission:
+    filter:
+      circle:
+        users:
+          _and:
+          - role:
+              _eq: 1
+          - address:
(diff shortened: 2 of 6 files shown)


crabsinger (dismissed):
This looks good to me. @topocount @META-DREAMER @zashton any thoughts you want to add?

zashton (changes requested):
migrations looks good, but input wise i think it seems to be lacking validation , the admin is able to insert any circle_id (he might not even belong to that circle) /data/name/type.

davidfant:
@zashton maybe I'm misunderstanding how the `insert_permissions` work, but wouldn't this (1) join the circle referenced by circle_id, and then join that circle's users? In that case all users fetched are connected to the circle, and if someone is an admin, they're authorized

zashton:
> @zashton maybe I'm misunderstanding how the `insert_permissions` work, but wouldn't this (1) join the circle referenced by circle_id, and then join that circle's users? In that case all users fetched are connected to the circle, and if someone is an admin, they're authorized

yeap, you are correct . the permissions should cover that circle id part, so you'll just need to clone the same set of permissions for the other role `superadmin`. Would type and data require specific values for your integration to work ?

davidfant:
> > @zashton maybe I'm misunderstanding how the `insert_permissions` work, but wouldn't this (1) join the circle referenced by circle_id, and then join that circle's users? In that case all users fetched are connected to the circle, and if someone is an admin, they're authorized
> 
> yeap, you are correct . the permissions should cover that circle id part, so you'll just need to clone the same set of permissions for the other role `superadmin`. Would `type` and `data` require specific values for your integration to work ?

Added permissions for the `superadmin role`!

For type and data: type would be e.g. "dework" or "github" and data is a JSON blob that contains whatever the integration needs to fetch tasks. For dework, that would be "{ organizationId: string }" which then is used to fetch Dework tasks. For Github it might be an oauth access/refresh token

zashton:
> > > @zashton maybe I'm misunderstanding how the `insert_permissions` work, but wouldn't this (1) join the circle referenced by circle_id, and then join that circle's users? In that case all users fetched are connected to the circle, and if someone is an admin, they're authorized
> > 
> > 
> > yeap, you are correct . the permissions should cover that circle id part, so you'll just need to clone the same set of permissions for the other role `superadmin`. Would `type` and `data` require specific values for your integration to work ?
> 
> @zashton I added permissions for the `superadmin role`!
> 
> For type and data: type would be e.g. "dework" or "github" and data is a JSON blob that contains whatever the integration needs to fetch tasks. For dework, that would be "{ organizationId: string }" which then is used to fetch Dework tasks. For Github it might be an oauth access/refresh token

Thanks, do commit and i'll review

zashton:
> @zashton forgot to push... Check now!

Do you need to put an unique constraint ?  i could insert multiple rows of the same circle id. I assume your FE integration will ignore the whole row of data if `type` and `data` is of the wrong format.

davidfant:
@zashton 
Unique constraint: I left it open for now. For example, if we'd want project-specific Dework connections and you wanted to add 2 boards, it might be easiest of that would be 2 integrations. Lmk if you disagree or think another path forward would make more sense.

Integrations with unknown type/data: yes they'll be ignored. That way we can keep adding integrations over time and the client will simply pick up the integrations it knows about/supports

zashton:
> @zashton Unique constraint: I left it open for now. For example, if we'd want project-specific Dework connections and you wanted to add 2 boards, it might be easiest of that would be 2 integrations. Lmk if you disagree or think another path forward would make more sense.
> 
> Integrations with unknown type/data: yes they'll be ignored. That way we can keep adding integrations over time and the client will simply pick up the integrations it knows about/supports

yeap make sense to me if a circle can have multiple boards ! looks good to me 👍

zashton (approved):

Issue #466, opened 2022-01-26
Build: Allocate Funds/Set Allowance


spydermonkee:
### Specification

Build the Allocate funds modal: 
https://www.figma.com/file/wXF7xGVv1j2SqwWfgbamS8/Coordinape-Design-v8?node-id=2136%3A23884

- only the vault owner can do this—this modal should be inaccessible to others
- see  https://www.notion.so/On-Chain-Vaults-27a8a1da30f04967a240df9dd4ddae76 for details of allowance-setting logic
- show a warning when the amount is larger than the funds in the vault, but don't prevent the form from being submitted
    - the vault owner can add funds later
- repeating
    - if the epoch repeats weekly, show a checkbox, "repeat funding weekly"
    - if it repeats monthly, show a checkbox, "repeat funding monthly"
        - this means: set `cooldownInterval` to 28 days
    - if the epoch has some non-standard repeating period, don’t show any repeating options
    - if repeating is enabled, the form needs to have a number input that specifies the number of epochs that the allowance should last for
        - "0" equals "don't repeat at all", "1" equals "repeat once", etc.
        - hide the number input if repeating is disabled

sshmm:
from what I understood from the description and notion:
1- This modal will be launched by Add Another Budget which has a separate issue
2- It will allocate funds from the vault balance to one of the circles
3- it doesn't have a design in Figma but the modal in #541 is close to what is required

levity:
We're going to hold off on this one for now; the design is still being
iterated on

PR #541, opened 2022-02-15
Allocate Fund Modal

f1lander:
project issue: #466 

### summary
UI ready, wired to the allocate fund button

### screenshots

image

### questions

1. are `fundsAvailable` should be required?
2. get fundsAvailable from real data
3. fields.repeat_monthly the flag to know the recurring label?
4. how do I get the epoch Period?

I need to understand how the `Fund Monthly/Weekly` works regarding the AdminVaultForm.FormController

### updates
02/21 Adds the allocate card to the AllocaModal


levity:
As we discussed in the meeting, please wire this component up in the app so that it replaces the existing `AllocateModal`

f1lander:
> As we discussed in the meeting, please wire this component up in the app so that it replaces the existing `AllocateModal`

done white_check_mark

src/ui/TextFieldFund/TextFieldFund.tsx
```diff
@@ -0,0 +1,116 @@
...
+        gap: '$sm',
+      }}
+    >
+      <Text
+        css={{
+          fontSize: '$2',
+          color: '$text',
+          fontWeight: '$light',
+        }}
+      >{`AVAILABLE: ${props.fundsAvailable.toLocaleString(
```
levity, line 67: This doesn't need to be wrapped in `{}`

src/ui/TextFieldFund/TextFieldFund.tsx
```diff
@@ -0,0 +1,116 @@
...
+        'en-US'
+      )} ${coinType}`}</Text>
+      <Box
+        css={{
+          display: 'flex',
+          backgroundColor: '$lightBackground',
+          borderRadius: '16px',
```
levity, line 74: maybe we can add this as radius `$5`?

src/ui/TextFieldFund/TextFieldFund.tsx
```diff
@@ -0,0 +1,116 @@
...
+          alignItems: 'center',
+          justifyContent: 'space-between',
+          width: '308px',
```
levity, line 77: we should probably not set width, and instead have it determined by parent elements, no?

src/ui/TextFieldFund/TextFieldFund.tsx
```diff
@@ -0,0 +1,116 @@
...
+        </Box>
+        <TextField
+          ref={inputRef}
+          value={inputFundValue}
+          onChange={handleChange}
+          variant="fund"
+        />
+        <Box
+          css={{
+            fontSize: 'calc($2 + 1px)',
```
levity, line 104: oh cool, glad this works! 🎉

src/ui/TextFieldFund/TextFieldFund.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+interface TextFieldFundProps {
+  fundsAvailable: number;
+  onChange(value: number): void;
+  value: number;
+  coinType?: string;
+}
+
+const parseFundToNumber = (value: string) => Number(value.replace(/,/g, ''));
+const parseFundToLocale = (value: number) => value.toLocaleString('en-US');
```
levity, line 13: This needs to be slightly different in order to let people type decimal values. Right now you can't type e.g. `12.50` because `12.` gets auto-converted to `12`

src/components/AllocateFundsCard/AllocateFundsCard.tsx
```diff
@@ -0,0 +1,134 @@
...
+          fontSize: '$4',
+          color: '$text',
+          fontWeight: '$light',
+          '@sm': {
+            fontSize: '$3',
+            textAlign: 'center',
+          },
+        }}
+      >
+        {`(Repeats ${props.recurringLabel})`}
```
levity, line 88: no need for `{}` here

levity (changes requested):

levity:
hey @f1lander -- using `AdminVaultForm` in `AllocateModal` is not actually what we should be doing, going forward. that was just a temporary thing that was done to pass type-checking when the modals were still non-functional. Sorry this wasn't clear; we should have done a better job of marking this in the code with comments. (Actually we shouldn't have done this in the first place... hindsight is 20/20, as they say)

So part of the work for this task is to set up a form handler that is specific to the needs of this modal, not use `AdminVaultForm`. You can continue to use the `createForm` helper, but it might be a better idea to use `react-hook-form` instead, like @teeolendo did in #550.

levity:
@f1lander ok, I looked into the code I was talking about, and fortunately I was able to remove all the other uses of AdminVaultForm. So you can keep on using it if you want. If you do, please rename it to `AllocateFundsForm`. You can still switch to `react-hook-form` if you want.

Now, I have more feedback on the fields in the form. I realized the form data doesn't actually need to contain the length of the repeating period, because this data is already present on the epoch. When we create a transaction, we can use both the form data and the epoch data. So the form data can actually be as simple as:

```
{ amount: number; repeat: boolean; repeat_times?: number }
```

f1lander:
> hey @f1lander -- using `AdminVaultForm` in `AllocateModal` is not actually what we should be doing, going forward. that was just a temporary thing that was done to pass type-checking when the modals were still non-functional. Sorry this wasn't clear; we should have done a better job of marking this in the code with comments. (Actually we shouldn't have done this in the first place... hindsight is 20/20, as they say)
> 
> 
> 
> So part of the work for this task is to set up a form handler that is specific to the needs of this modal, not use `AdminVaultForm`. You can continue to use the `createForm` helper that @exrhizo wrote, but it's probably a better idea to use `react-hook-form` instead, like @teeolendo did in #550.

@levity understood, at this point I was focusing on the ui as Tony and Spyder said, I didn't know about AdminVaultForm was temporary thing, sorry, I can start doing that, do you want to me create another branch and another PR for this or do it in this same branch?

levity:
@f1lander let's stay in the same branch. Make sure you check out my second comment, which I posted just 1 minute before you responded

f1lander:
> @f1lander ok, I looked into the code I was talking about, and fortunately I was able to remove all the other uses of AdminVaultForm. So you can keep on using it if you want. If you do, please rename it to `AllocateFundsForm`. You can still switch to `react-hook-form` if you want.
> 
> 
> 
> Now, I have more feedback on the fields in the form. I realized the form data doesn't actually need to contain the length of the repeating period, because this data is already present on the epoch. When we create a transaction, we can use both the form data and the epoch data. So the form data can actually be as simple as:
> 
> 
> 
> ```
> 
> { amount: number; repeat: boolean; repeat_times?: number }
> 
> ```
> 
> 

Perfect, I'll check it tomorrow

teeolendo:
Branch is stale. These changes will be implemented elsewhere.

PR #547, opened 2022-02-16
fixes #435 and fixes #439 by porting uploadProfileAvatar and uploadProfileBackground from laravel to hasura

### LINKED ISSUE #435: Migrate POST v2/upload-avatar
## Summary:
Migrate Upload Avatar Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `POST v2/upload-avatar` to Hasura.

We'll need to use a Hasura action to upload the data to S3, and then write the mutation to the database.

**Middleware:** auth token

validate max 10mb, file type jpg,bmp,gif,png

**Logic:**
80% image quality compression, append current timestamp to filename, delete previous image if exists and upload modified image to s3

**DB Tables:**
profiles

#### Notes:

 - Ngnix limits us to 1mb, let's use 2mb max, like twitter.

- We'll need to setup an S3 upload client in TS. Figure out what env vars are needed, and we can provision those env vars in our different environments (staging/prod). Let's find a well used library.

- We could convert the filename to wallet id pattern we'll be using in ceramic.

### LINKED ISSUE #439: Migrate POST v2/upload-background
## Summary:
Migrate Upload Avatar Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `POST v2/upload-background` to Hasura.

**Middleware:**
auth token

validate max 10mb, file type jpg,bmp,gif,png

**Logic:**
80% image quality compression, append current timestamp to filename, delete previous image if exists and upload modified image to s3

**DB Tables:**
profiles

CryptoGraffe:
This includes porting the FE as well. There is some unfinished business in here w/ FE patterns but it is functional for now and I left good comments for future improvements. 

fixes #435 

Test plan after deploy to staging/production:

1. visit My Profile by clicking your avatar in top right
2. click Edit profile
3. click the pencil icon on avatar
4. choose an avatar image
5. click the little upload button that shows up on the avatar
5. make sure it finishes ok
6. right click on the avatar now and copy the image url, open it in a new tab , should be an s3 image url
7. change the avatar again (steps 3-5)
8. reload the other tab, that image should be gone/deleted now

Perform a similar test for edit background, it's on the main profile page and has same workflow. 

*FOR DEPLOYMENT* 
These api keys need to be set or the instance that is running vercel needs to have the proper EC2 Role, I'm not sure how that works. 
```
IMAGES_AWS_ACCESS_KEY_ID=tbd
IMAGES_AWS_SECRET_ACCESS_KEY=tbd
```


api/hasura/actions/upload_profile_avatar.ts
```diff
@@ -0,0 +1,107 @@
...
+    HasuraUserSessionVariables
+  ).parse(req.body);
+
+  try {
+    // base64 decode the provided image data
+    const imageBytes = Buffer.from(input.image_data_base64, 'base64');
+
+    // file size check
+    if (imageBytes.byteLength > MAX_IMAGE_BYTES_LENGTH) {
+      return res.status(400).json({
```
crabsinger, line 34: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413 might be a nice code to more accurately descripe the error here.

src/hooks/gql.ts
```diff
@@ -37,6 +41,33 @@ function useTypedQuery<
   );
 }
 
+// this isn't used yet, but it would be great to be able to use it
+// I couldn't figure out how to get it to work with variables -CryptoGraffe
```
crabsinger, line 45: @META-DREAMER @topocount maybe can help get the variables to work here.

CryptoGraffe:
I'm not sure why the hasura-cloud-preview step is failing on here, not familiar w/ it

api/hasura/actions/upload_profile_avatar.ts
CryptoGraffe: I switched this to 413 and moved it to images.ts

CryptoGraffe:
> 

this appears to be fixed now

api/hasura/actions/upload_profile_avatar.ts
```diff
@@ -0,0 +1,96 @@
...
+    session_variables: sessionVariables,
+  } = composeHasuraActionRequestBodyWithSession(
+    updateProfileAvatarInput,
+    HasuraUserSessionVariables
+  ).parse(req.body);
+
+  try {
+    const imageBytes = parseBase64Image(res, input.image_data_base64);
+    if (!imageBytes) {
+      return res;
```
topocount, line 26: should this return some form of a 422 error?

api/hasura/actions/upload_profile_avatar.ts
```diff
@@ -0,0 +1,96 @@
...
+          avatar: true,
+        },
+      ],
+    });
+
+    // see if there was a previous avatar so we can delete it after new one is saved
+    let previousAvatar: string | undefined = undefined;
+    if (profiles_by_pk) {
+      previousAvatar = profiles_by_pk.avatar;
+    }
```
topocount, line 45: this is a nit, but can we do this after the graphql mutation completes?

api/hasura/actions/upload_profile_avatar.ts
```diff
@@ -0,0 +1,96 @@
...
+        {
+          id: true,
+          avatar: true,
+          address: true,
+        },
+      ],
+    });
+
+    if (mutationResult.update_profiles_by_pk) {
+      if (previousAvatar) {
```
topocount, line 81: let's first check the `previousAvatar` existence here, and clean up the code I mentioned above

api/hasura/images.ts
```diff
@@ -0,0 +1,45 @@
+import { VercelResponse } from '@vercel/node';
```
topocount, line 1: This file should be in `api-lib`, not `api`.

api/hasura/images.ts
```diff
@@ -0,0 +1,45 @@
...
+): Buffer | undefined {
+  // base64 decode the provided image data
+  const imageBytes = Buffer.from(imageDataBase64, 'base64');
+  // file size check
+  if (imageBytes.byteLength > MAX_IMAGE_BYTES_LENGTH) {
+    res.status(413).json({
+      message: `image size is larger than maximum allowed: ${imageBytes.byteLength}/${MAX_IMAGE_BYTES_LENGTH}`,
+      code: '413',
+    });
+  }
```
topocount, line 18: love this

src/hooks/gql.ts
topocount: I think we should put a pin in this and attack this issue in separate task soon after this one. To be clear, I don't want to deprioritize this for @CryptoGraffe, I just think it should be a separate PR that closely follows this one.

There's already a lot going on in this one.

api/hasura/images.ts
```diff
@@ -0,0 +1,45 @@
...
+import sharp from 'sharp';
+
+import { MAX_IMAGE_BYTES_LENGTH } from '../../src/lib/images';
+
+export function parseBase64Image(
+  res: VercelResponse,
+  imageDataBase64: string
+): Buffer | undefined {
+  // base64 decode the provided image data
+  const imageBytes = Buffer.from(imageDataBase64, 'base64');
```
topocount, line 11: Can we abstract this into a subfunction that lives in src/lib and rename this function `parseBase64ImageWithResponse`? It'd be fun to make this outer function more generic together, but I think keeping the outer function the interface as-is for now is fine

api-lib/s3.ts
```diff
@@ -0,0 +1,30 @@
+import aws from "aws-sdk";
+
+const s3 = new aws.S3({
+  accessKeyId: process.env.AWS_ACCESS_KEY_ID || 'missing_env_var',
+  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || 'missing_env_var',
+  endpoint: process.env.AWS_ENDPOINT ? process.env.AWS_ENDPOINT : undefined,
+});
+
+const uploadImage = (Key: string, Body: Buffer) => {
+  const p = {
```
topocount, line 10: nit: can we give this a name? a proper name is much more information-rich, even if the variable is trivial than a letter.

topocount (commented):
If possible, I'd strongly prefer to combine these actions into a single action that optionally takes either/both a profile or background image payload and behaves accordingly. @zashton I'm curious what you think about this idea.

Otherwise it's looking quite good

zashton:
> If possible, I'd strongly prefer to combine these actions into a single action that optionally takes either/both a profile or background image payload and behaves accordingly. @zashton I'm curious what you think about this idea.
> 
> Otherwise it's looking quite good

Either or is good, I would not do both .You'll need to roll things back + delete image if the 2nd upload fails. doing separate mutations might be more clean

api/hasura/images.ts
CryptoGraffe: The reason it's not in src/lib is because it refers to VercelResponse, which seemed inappropriate for the FE, since this is all backend code.

api/hasura/images.ts
CryptoGraffe: Same comment as below, api-lib is imported by the FE and it seemed inappropriate to import VercelResponse in there. Perhaps some other folder structure would be helpful here.

api/hasura/actions/upload_profile_avatar.ts
CryptoGraffe: It will either return the imageBytes, or a 413, as far as I can see.

api/hasura/actions/upload_profile_avatar.ts
CryptoGraffe: sure good call

CryptoGraffe: Actually no we can't, or I don't understand - we are doing the delete after the mutation succeeds, but this lookup has to happen beforehand or there is no way to grab the previousAvatar (it will be gone). Maybe you meant something else?

CryptoGraffe:
I guess the whole api-lib thing is Backend, really. Its just that gql.ts is imported by the FE, so maybe that is the lone part that needs to be refactored.

CryptoGraffe:
> I guess the whole api-lib thing is Backend, really. Its just that gql.ts is imported by the FE, so maybe that is the lone part that needs to be refactored.

Ok ignore my delusions here. api-lib is all backend stuff. I'm moving stuff around now.

src/hooks/gql.ts
CryptoGraffe: I'm content leaving this code in here and addressing it as we will out the FE later.

api/hasura/actions/upload_profile_avatar.ts
CryptoGraffe: I think it has to happen in the order as written, because previous has to be fetched before the mutation, otherwise it will be the mutated value

api/hasura/images.ts
CryptoGraffe: moving this, i was tripping

api/hasura/images.ts
CryptoGraffe: moving this and renaming, i'm tripping

api-lib/s3.ts
CryptoGraffe: done

api/hasura/actions/upload_profile_avatar.ts
zashton: Its possible to do an event trigger here, so the delete image can happen async in the background (as its not time sensitive nor critical). In the event handler just do a check on the payload for the old vs new column value for avatar column and execute the s3 delete.

.env.example
```diff
@@ -34,6 +35,8 @@ REACT_APP_S3_BASE_URL=http://coordinape.s3-website.localhost.localstack.cloud:45
 NODE_HASURA_URL=http://localhost:8080/v1/graphql
 HASURA_EVENT_SECRET=event-secret
 HASURA_GRAPHQL_ADMIN_SECRET=admin-secret
+AWS_ENDPOINT=http://s3.localhost.localstack.cloud:4566
```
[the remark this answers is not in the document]
CryptoGraffe: No, its only needed for local dev as well. It's the url for the aws sdk to connect to s3 instead of using REAL s3. In production this will be unset.

.env.example
```diff
@@ -44,3 +47,9 @@ HASURA_GRAPHQL_ADMIN_SECRET=admin-secret
 #REACT_APP_API_BASE_URL=https://coordinape.me/api
 #REACT_APP_HASURA_URL=https://coordinape-prod.hasura.app/v1/graphql
 #REACT_APP_S3_BASE_URL=https://coordinape-prod.s3.amazonaws.com/
+
+
+# These will need to be set to something meaningful for deployment, but in local can be unset
+AWS_ACCESS_KEY_ID=tbd
```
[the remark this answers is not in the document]
CryptoGraffe: We could prefix them with IMAGES_ or something since that is the only use. I can make that change. On your comment about the event handler - It's a good concept but i think its unnecessarily complex. This is a low impact operation (should be couple milliseconds) and it happens very infrequently.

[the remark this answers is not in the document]
CryptoGraffe: Ok I renamed them to be `IMAGES_AWS_X`

.env.example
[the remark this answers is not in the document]
CryptoGraffe: the `REACT_APP_S3_BASE_URL` is for serving images over http, which is diff

.env.example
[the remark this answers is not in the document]
zashton: @CryptoGraffe sure, have no issues with deleting the image in the same mutation, as long as we ensure it doesn't throw an exception if the delete fails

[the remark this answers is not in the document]
CryptoGraffe: Good call, I wrapped it in a try catch

CryptoGraffe:
I went through and refactored a bunch of the mutation/previous image delete / etc code to be shared. This will help for circle logo implementation as well which I am doing today. It should all be in the most recent commit.

src/hooks/useApiWithProfile.ts
```diff
@@ -31,15 +46,18 @@ export const useApiWithProfile = () => {
 
   const updateAvatar = useRecoilLoadCatch(
     () => async (newAvatar: File) => {
-      await getApiService().uploadAvatar(newAvatar);
+      // TODO: ideally we would use useTypedMutation instead of this but I couldn't get the variables to work w/ mutation -CryptoGraffe
+      const image_data_base64 = await fileToBase64(newAvatar);
```
zashton, line 50: We could switch this and updateBackground back temporarily to the old endpoint until the hasura prod instance is ready

api-lib/profileImageMutations.ts
```diff
@@ -0,0 +1,69 @@
...
+export const profileUpdateImageMutation = (
+  hasuraProfileId: number,
+  _set: (fileName: string) => any
+) => {
+  // save the new image id in the db
+  return async (fileName: string): Promise<any> => {
+    const mutationResult = await gql.q('mutation')({
+      update_profiles_by_pk: [
+        {
+          _set: _set(fileName),
```
topocount, line 36: Is this needed to override some annoying type mismatch? If so, consider returning `unknown` here instead of `any` for the `_set` function. Also, it'd be great to leave a comment here explaining why this is needed since this effectively turns off the zeus type tooling.

api-lib/profileImageMutations.ts
```diff
@@ -0,0 +1,69 @@
...
+  hasuraProfileId: number,
+  returning: () => any
+): Promise<{ avatar?: string; background?: string } | undefined> => {
+  // Figure out if there was a previous avatar, because we'll need to delete it
+  const { profiles_by_pk } = await gql.q('query')({
+    profiles_by_pk: [
+      {
+        id: hasuraProfileId,
+      },
+      returning(),
```
topocount, line 65:
```suggestion
      {background: true, avatar: true},
```

CryptoGraffe: this is a good easy change, i've pushed it

api-lib/profileImageMutations.ts
```diff
@@ -0,0 +1,69 @@
...
+        profile: mutationResult.update_profiles_by_pk,
+      };
+    }
+    throw 'mutation was not successful';
+  };
+};
+
+export const profilePreviousImageQuery = async (
+  hasuraProfileId: number,
+  returning: () => any
```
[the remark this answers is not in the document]
CryptoGraffe: done

src/hooks/useApiWithProfile.ts
CryptoGraffe: I'm not sure that we need to do that since main isn't going to prod

api-lib/profileImageMutations.ts
CryptoGraffe: pushed a small change here, still dont love having to pass in a func

api-lib/profileImages.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    if (!mutationResult.update_profiles_by_pk) {
+      throw 'profile mutation was not successful';
+    }
+
+    return mutationResult.update_profiles_by_pk;
+  };
+};
+
+export const profileImages = async (
```
topocount, line 65:
```suggestion
export const getProfileImages = async (
```

topocount (commented):
@CryptoGraffe the code looks good to me.

My only other request is that you provide an explicit test plan for validating the code at runtime in the PR description.

Once I complete that part of the review, I'll approve

src/hooks/useApiWithProfile.ts
zashton: quite sure main auto-deploys to prod based on vercel. We'll just do a quick check on the upload mutation functionality once this gets merged.

zashton (approved):
nothing else for me, we'll do a quick check on the mutations once this gets merged to main to make sure its not pointing to the hasura prod instance

zashton:
@CryptoGraffe @topocount went through the test plan all working good on local,
broken on the preview site though. guess we can ignore that for now
image

topocount (approved):
:tada:

PR #551, opened 2022-02-21
Fixes #440 by implementing hasura action and FE changes for circle lo…

### LINKED ISSUE #440: Migrate POST v2/{circle_id}/admin/upload-logo
## Summary:
Migrate Upload Avatar Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `POST v2/upload-background` to Hasura.

**Middleware:**
**auth token & circle admin**

validate max 10mb, file type jpg,bmp,gif,png

**Logic:**
resize image to 100 px width while keeping its aspect ratio, 80% image quality compression, append current timestamp to filename, delete previous image if exists and upload modified image to s3

**DB Tables:**
circles

#### Notes:

 - Different in that it requires the circle admin

CryptoGraffe:
Fixes #440 by implementing hasura action and FE changes for circle logo upload. Currently targeting merge into 435-hasura-upload-avatar but will move to main after that is merged. 

Test plan after deploy to staging/local:

set 
`REACT_APP_HASURA_ENABLED=true`
in .env to enable this. It's disabled by default so it doesn't disrupt production. 

1. visit the admin page  for a circle
2. click Edit settings
3. click the pencil icon on the circle logo
4. choose an image
5. click save at the bottom of the modal dialog
6. make sure it finishes ok
7.  close the modal
8. reopen the edit settings modal
7. right click on the logo now and copy the image url, open it in a new tab , should be an s3 image url
8. change the logo again (steps 3-6)
9. reload the other tab, that image should be gone/deleted now

### DIFF
--- a/api-lib/images.ts
+++ b/api-lib/images.ts
@@ -23,6 +23,8 @@ function parseBase64(imageDataBase64: string): Buffer {
 }
 
 async function resizeAvatar(imageBytes: Buffer) {
+  // convert to a small jpeg w/ 80% quality
+  // This is small because of the map view which needs to show lots of small avatars in a canvas
   const img = sharp(imageBytes);
   return img
     .resize({
@@ -37,6 +39,7 @@ async function resizeAvatar(imageBytes: Buffer) {
 }
 
 async function resizeBackground(imageBytes: Buffer) {
+  // convert to jpeg w/ 80% image quality
   const img = sharp(imageBytes);
   return img
     .jpeg({
@@ -46,6 +49,7 @@ async function resizeBackground(imageBytes: Buffer) {
 }
 
 async function resizeCircleLogo(imageBytes: Buffer) {
+  // convert to jpeg w/ 80% image quality
   const img = sharp(imageBytes);
   return img
     .jpeg({
--- a/api-lib/images.ts
+++ b/api-lib/images.ts
@@ -45,4 +45,13 @@ async function resizeBackground(imageBytes: Buffer) {
     .toBuffer();
 }
 
-export { resizeAvatar, resizeBackground };
+async function resizeCircleLogo(imageBytes: Buffer) {
+  const img = sharp(imageBytes);
+  return img
+    .jpeg({
+      quality: 80,
+    })
+    .toBuffer();
+}
+
+export { resizeAvatar, resizeBackground, resizeCircleLogo };
--- a/api-lib/validate.ts
+++ b/api-lib/validate.ts
@@ -5,21 +5,22 @@ import { HASURA_EVENT_SECRET } from './config';
 export const verifyHasuraRequestMiddleware = (handler: VercelApiHandler) => {
   return async (req: VercelRequest, res: VercelResponse) => {
     if (
-        !req.headers.verification_key ||
-        (req.headers.verification_key as string) !== HASURA_EVENT_SECRET
+      !req.headers.verification_key ||
+      (req.headers.verification_key as string) !== HASURA_EVENT_SECRET
     ) {
       // return here to prevent further execution
       res.status(401).json({
         message: 'Unauthorized access',
-        code: 401,
+        code: '401',
       });
       return;
     }
     try {
       await handler(req, res);
     } catch (error: any) {
+      console.error(error);
       res.status(500).json({
-        code: 500,
+        code: '500',
         message: 'internal error: ' + error,
       });
     }
--- a/api-lib/validate.ts
+++ b/api-lib/validate.ts
@@ -18,7 +18,6 @@ export const verifyHasuraRequestMiddleware = (handler: VercelApiHandler) => {
     try {
       await handler(req, res);
     } catch (error: any) {
-      console.error(error);
       res.status(500).json({
         code: '500',
         message: 'internal error: ' + error,
--- a/api/hasura/actions/upload_circle_logo.ts
+++ b/api/hasura/actions/upload_circle_logo.ts
@@ -1,6 +1,6 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node';
 
-// import { authCircleAdminMiddleware } from '../../../api-lib/circleAdmin';
+import { authCircleAdminMiddleware } from '../../../api-lib/circleAdmin';
 import { gql } from '../../../api-lib/Gql';
 import { ErrorResponse } from '../../../api-lib/HttpError';
 import { resizeCircleLogo } from '../../../api-lib/images';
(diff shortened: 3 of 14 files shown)


api/hasura/actions/upload_circle_logo.ts
```diff
@@ -0,0 +1,77 @@
...
+      throw 'circle mutation was not successful';
+    }
+    return {
+      id: mutationResult.update_circles_by_pk.id,
+    };
+  };
+}
+
+// TODO: this does not work and I'm not sure why
+//export default authCircleAdminMiddleware(handler);
```
zashton, line 77: I tried commenting out and using this middleware works fine for me

zashton (changes requested):
i had to do a yarn generate to make things work as its missing some types from the circle_integrations PR. After that + including the circleadmin middleware, upload works according to the test case

CryptoGraffe:
I had to do a yarn generate and push those changes because the circle_integrations work from previous PR hadn't been generated & committed

api-lib/validate.ts
```diff
@@ -5,23 +5,24 @@ import { HASURA_EVENT_SECRET } from './config';
...
         message: 'Unauthorized access',
-        code: 401,
+        code: '401',
       });
       return;
     }
     try {
       await handler(req, res);
     } catch (error: any) {
+      console.error(error);
```
zashton, line 21: remove the console

zashton (dismissed):
good to go for the 3 mutations updated, just needs to remove the console

src/lib/gql/index.ts
```diff
@@ -55,10 +55,35 @@ export function getGql(url: string, getToken: () => string) {
...
+    circleId: number,
+    image_data_base64: string
+  ) =>
+    makeQuery(url, getToken)('mutation')(
+      {
+        upload_circle_logo: [
+          {
+            object: {
+              image_data_base64: $`image_data_base64`,
+              circle_id: $`circleId`,
```
topocount, line 68: are the templates needed here?

api/hasura/actions/upload_profile_background.ts
```diff
@@ -8,8 +8,9 @@ import {
   profileUpdateBackgroundMutation,
   userAndImageData,
 } from '../../../api-lib/profileImages';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
 
-export default async function (req: VercelRequest, res: VercelResponse) {
+async function handler(req: VercelRequest, res: VercelResponse) {
```
topocount, line 13: kinda embarrassed that i missed this :grimacing:

api/hasura/actions/upload_profile_avatar.ts
```diff
@@ -8,8 +8,9 @@ import {
   profileUpdateAvatarMutation,
   userAndImageData,
 } from '../../../api-lib/profileImages';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
 
-export default async function (req: VercelRequest, res: VercelResponse) {
+async function handler(req: VercelRequest, res: VercelResponse) {
```
topocount, line 13: and this

api-lib/images.ts
```diff
@@ -45,4 +45,13 @@ async function resizeBackground(imageBytes: Buffer) {
     .toBuffer();
 }
 
-export { resizeAvatar, resizeBackground };
+async function resizeCircleLogo(imageBytes: Buffer) {
+  const img = sharp(imageBytes);
+  return img
+    .jpeg({
+      quality: 80,
```
topocount, line 52: Checking out this API, and it's not completely clear what an 80% quality change means. I imagine this is just a simple lossy operation that decreases the storage footprint of the operation, but does this conform to some api specification or does this just follow some convention? I also noticed this is the default and could probably be omitted.

I guess I'm saying it'd be good to explain why this is being done with a comment or something.

api/hasura/actions/upload_circle_logo.ts
```diff
@@ -0,0 +1,77 @@
...
+    // session_variables: sessionVariables,
+  } = composeHasuraActionRequestBodyWithSession(
+    uploadCircleImageInput,
+    HasuraUserSessionVariables
+  ).parse(req.body);
+
+  try {
+    const previousLogo = await getPreviousLogo(input.circle_id);
+
+    const updater = new ImageUpdater<{ id: number }>(
```
topocount, line 26: The more I look at the `ImageUpdater` class the more i appreciate it. 

I might be wrong about this, but you shouldn't need to explicitly add the <`T`> when instantiating the class. Typescript should be able to infer this from the mutation function you pass in. If you want to explicitly pass it in, that's cool, as long as typechecking fails when it doesn't match the actual return.

topocount (commented):
This looks quite good! A couple requests:

1. Please `npm i` the typescript typedefs for `uuid`. This is leftover from the preceding changeset and I forgot to mention it.

topocount (dismissed):

api-lib/images.ts
CryptoGraffe: This converts any base type to jpeg and uses jpeg 80% quality which is a standard 0-100 slider in jpeg land. I'd much rather have the number in here than have the mysterious default cuz that's really at the taste of the library. I also think we are going to tweak these going forward - I'm just matching the laravel impl for now but I have thoughts about how we might handle some of this differently in the future (e.g. multiples copies of images at various qualities/sizes).

CryptoGraffe: I left a comment

api/hasura/actions/upload_circle_logo.ts
CryptoGraffe: I like being explicit 🤷

CryptoGraffe:
> 

this is already in package.json , what am I missing?

```
    "@types/uuid": "^8.3.4",
```

src/lib/gql/index.ts
CryptoGraffe: I'm on the fence about this, it seems like good practice is to always use the gql variable pattern, but its also pretty verbose. In either case, we should be consistent about it.

api/hasura/actions/upload_circle_logo.ts
CryptoGraffe: I actually think the generic aspect of it might be unnecessary altogether since it always returns the same thing in the current impl, I was just geeking out a bit and trying to be flexible. There may be a version in future where we want to return { id: number, profile: x} or something for example

topocount (approved):

PR #552, opened 2022-02-22
Feature/create nominee Hasura migration

### LINKED ISSUE #455: Migrate Create Nominee Endpoint Hasura
Logic
- Use the circle's nomination day limit and add today's date to it to calculate the expiry date of the nomination and also capture the `min_vouches` value from the circle. 
- Check if `min_vouches` is = 1 if yes create a new user in the circle with the address
- Create a new profile record for the address if it doesn't exist already.

Validation
- Check if the nominee address exists as a user in the circle or exists in the nomination table of the circle

zashton:
Closes #455 

### Test Plan

Prerequisite for testing notifications : Telegram Bot + `chat_id` & discord webhook

1. Go to `http://localhost:3000/admin/circles` Settings page enable Vouching and set minimum vouches to 2
2. In the DB go to the circle you wish to test on and update `telegram_id` to your `chat_id` and the `discord_webhook` accordingly.
2. run this mutation in your hasura console.
```gql
mutation MyMutation {
  createNominee(payload: {address: "<address>", description: "<desc>", name: "<name>", circle_id: <circle_id>}) {
    nominee {
      id
      circle_id
      description
      name
      nominated_by_user_id
      nominated_date
      expiry_date
      ended
      created_at
      address
      vouches_required
    }
  }
}
```
3. You should recieve a notification in both discord and telegram chat to indicate that a user has been nominated.
4. go to `http://localhost:3000/vouching` and you should see the nomination appear on the page.
5. go back to `http://localhost:3000/admin/circles` Settings page and set the minimum vouches to 1.
6. Run the above mutation again with a different address.
7. This time you should 2 notifications each, first is the nomination message and the second is the message indicating the nomination has received enough vouches and user has been added to circle.
8. Go to `http://localhost:3000/admin/circles` and you should see the nomination added as a user in the table


api-lib/createNomineeMutations.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export const insertNominee = async (
+  nominatedByUserId: number,
+  circleId: number,
+  address: string,
+  name: string,
+  description: string,
+  nominationDaysLimit: number,
+  vouchesRequired: number
+) => {
```
META-DREAMER, line 94: should make this a single object param instead of multiple params like this, cleaner that way and lets you specify fields in any order.
```suggestion
export const insertNominee = async (params: {
  nominatedByUserId: number,
  circleId: number,
  address: string,
  name: string,
  description: string,
  nominationDaysLimit: number,
  vouchesRequired: number
}) => {
```

api-lib/profileImages.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+};
+
+export const profileUpdateBackgroundMutation = (hasuraProfileId: number) => {
+  return profileUpdateImageMutation(hasuraProfileId, (fileName: string) => {
+    return { background: fileName };
+  });
+};
+
+const profileUpdateImageMutation = (
```
META-DREAMER, line 39: It would be better to organize mutations / queries in a subfolder under api-lib, instead of spreading them out across multiple files in a flat way. Will be easier to maintain and track that way. e.g. having a single `imageHelpers.ts` file that contains all the non graphql stuff, then having `api-lib/mutations/profile.ts` that contains profile update mutations etc. 

Also, for naming functions should try to word them in a way that describes what they **do**. e.g. for this, a better name would be `createProfileUpdateImageMutation`

api-lib/profileImages.ts
```diff
@@ -0,0 +1,84 @@
...
+
+import {
+  composeHasuraActionRequestBodyWithSession,
+  HasuraUserSessionVariables,
+  uploadImageInput,
+} from '../src/lib/zod';
+
+import { gql } from './Gql';
+
+export const userAndImageData = (
```
META-DREAMER, line 11: Better name would be `parseUploadImageRequest`

src/lib/zod/index.ts
```diff
@@ -39,6 +48,10 @@ export const circleIdInput = z
   })
   .strip();
 
+export const uploadImageInput = z
+  .object({ image_data_base64: z.string() })
+  .strict();
```
META-DREAMER, line 53: The name of this variable makes it sound like it's a function that will upload an image. Might be better to Capitalize it instead and name it `UploadImageInput` or `UploadImageInputSchema`

src/lib/zod/index.ts
```diff
@@ -17,6 +17,15 @@ export const createCircleSchemaInput = z
     'Either Protocol name should be filled in or a Protocol should be selected.'
   );
 
+export const createNomineeSchemaInput = z
```
META-DREAMER, line 20: `CreateNomineeInputSchema`

api-lib/images.ts
```diff
@@ -0,0 +1,48 @@
+import sharp from 'sharp';
+
+import { MAX_IMAGE_BYTES_LENGTH } from '../src/lib/images';
+
+import { PayloadTooLargeError } from './HttpError';
+
+export type ImageResizer = (imageBuffer: Buffer) => Promise<Buffer>;
+
+export function parseBase64Image(imageDataBase64: string): Buffer {
```
META-DREAMER, line 9: should name this file `imageHelpers.ts` instead since its not actually images but a set of helpers to manipulate images

api-lib/profileImages.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    if (!mutationResult.update_profiles_by_pk) {
+      throw 'profile mutation was not successful';
+    }
+
+    return mutationResult.update_profiles_by_pk;
+  };
+};
+
+export const profileImages = async (
```
META-DREAMER, line 65: should name this `getProfileImages`

api-lib/sendSocialMessage.ts
```diff
@@ -0,0 +1,67 @@
+import fetch from 'node-fetch';
+
+import { TELEGRAM_BOT_BASE_URL } from './config';
+import { DISCORD_BOT_NAME, DISCORD_BOT_AVATAR_URL } from './constants';
+import { gql } from './Gql';
+
+type SocialMessage = {
```
META-DREAMER, line 7: Feels like this should be named `BotMessage`, `sendBotMessage` etc

api/hasura/actions/create_nominee.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    const {
+      input: { object: input },
+      session_variables: sessionVariables,
+    } = composeHasuraActionRequestBody(createNomineeSchemaInput).parse(
+      req.body
+    );
+    const { circle_id, address, name, description } = input;
+
+    if (sessionVariables.hasuraRole !== 'admin') {
```
META-DREAMER, line 28: Instead of putting all the logic in an if block like this, it would be easier to follow if you instead terminated early if the condition was not met. 

e.g. 

```ts
if (sessionVariables.hasuraRole === 'admin')
  return res.status(...);

// rest of the logic
```

Otherwise its difficult to figure out the code path

api/hasura/event-triggers/utils/handleCheckNomineeMsg.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (nominees_by_pk) {
+      await sendSocialMessage({
+        message: `Nominee ${nominees_by_pk.name} has only received ${nominees_by_pk.nominations_aggregate?.aggregate?.count} vouch(es) and has failed`,
+        circleId: nominees_by_pk.circle_id,
+        channels,
+      });
+      return true;
+    }
+  }
+  return false;
```
META-DREAMER, line 29: Instead of wrapping the main logic in an if block and returning false at the end, it would be easier to follow if you returned early by inverting the if condition and then put the main logic after that.

```suggestion
  const shouldSendMessage = 
    data.old.ended === false &&
    data.new.ended === true &&
    new Date(data.new.expiry_date) < new Date();
    
  if (!shouldSendMessage)  return false;

  const { nominees_by_pk } = await gql.getNominee(data.new.id);

  if (!nominees_by_pk) return false;
  
   await sendSocialMessage({
      message: `Nominee ${nominees_by_pk.name} has only received ${nominees_by_pk.nominations_aggregate?.aggregate?.count} vouch(es) and has failed`,
      circleId: nominees_by_pk.circle_id,
      channels,
    });
  return true;
```

api/hasura/event-triggers/utils/handleCheckNomineeMsg.ts
```diff
@@ -0,0 +1,30 @@
+import { gql } from '../../../../api-lib/Gql';
+import { sendSocialMessage } from '../../../../api-lib/sendSocialMessage';
+import { EventTriggerPayload } from '../../types';
+
+export default async function handleCheckNomineeMsg(
```
META-DREAMER, line 5: this file should be outside of the event-triggers folder since its not actually an event trigger itself.

META-DREAMER (changes requested):
Looks like this PR is trying to do a bunch of things at once (nominee stuff, but also handling image uploads etc), would be good to split it up next time. 

Bunch of comments on better naming for things (always the hardest part!) and general code organization

zashton:
> Looks like this PR is trying to do a bunch of things at once (nominee stuff, but also handling image uploads etc), would be good to split it up next time.
> 
> Bunch of comments on better naming for things (always the hardest part!) and general code organization

not sure why the image uploads code was appearing in my PR, rebased and its gone now 😆

api/hasura/event_triggers/createVouchedUser.ts
```diff
@@ -0,0 +1,80 @@
...
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { gql } from '../../../api-lib/Gql';
+import { EventTriggerPayload } from '../../../api-lib/types';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    event: { data },
+  }: EventTriggerPayload<'nominees', 'INSERT'> = req.body;
```
zashton, line 12: @CryptoGraffe this is the event trigger that creates the user from the nomination, we could create a similar one that triggers when `vouches` `INSERT` with the same business logic

api/hasura/event_triggers/createVouchedUser.ts
```diff
@@ -0,0 +1,80 @@
...
+        insert_users_one: [
+          { object: { name, address, circle_id } },
+          {
+            id: true,
+          },
+        ],
+        // End current nomination
+        update_nominees: [
+          {
+            _set: { ended: true },
```
zashton, line 56: I will need to set the user_id here , any clue how to get it within the same mutation ?

api/hasura/actions/createNominee.ts
```diff
@@ -0,0 +1,97 @@
...
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  try {
+    const {
+      input: { payload: input },
+      session_variables: sessionVariables,
+    } = composeHasuraActionRequestBody(createNomineeSchemaInput).parse(
+      req.body
+    );
+    if (sessionVariables.hasuraRole === 'admin') {
```
CryptoGraffe, line 27: You can use `composeHasuraActionRequestBodyWithSession` and it verifies user and not admin

CryptoGraffe:
Looks good to me, I need to test locally though.

src/lib/zod/index.ts
topocount: I want to second this suggestion

CryptoGraffe:
Fixes #455

PR #572, opened 2022-02-25
restoring non-graphql endpoints and adding a HASURA_ENABLED env var to re-enable

CryptoGraffe:
Follow the test plan from #547 to verify this works with laravel, but realy this needs to be tested in prod, which does not have hasura configured properly.


zashton (changes requested):
tested working on local, could we add the env var to `env.example` as well

zashton (approved):
LGTM

PR #574, opened 2022-02-25
Port Vouching to Hasura

### LINKED ISSUE #519: Hasura migration: Vouching
Validation
- Check if voucher exists in the same circle as the nominee
- Check if voucher already has an existing vouch for the nominee

Business Logic
- Check if `vouches_required` of the nominee is now equal than the number of vouches+1 (nomination = 1 vouch, so only 1 vouch is needed if `vouches_required` is 2). If vouches is sufficient, create a user in the circle with the nominee's address & link the `user_id` to the nominee record (make sure no user with this address exists in the circle) , create a profile with the address if it doesn't exist, end the nomination by setting it `end` = true . 

Notifications via Telegram/Discord (if set)
- nominee_name has been vouched for by voucher_name!
- nominee_name has received enough vouches and is now in the circle!

CryptoGraffe:
Fixes #519 by porting vouching to hasura. 

To test: 

1. have a bearer token set up in hasura console for a user in circle_id 6
2. run mutation for a user who is nominated but needs vouching. I used this: 
```

mutation MyMutation {
  vouch(payload: {circle_id: 6, nominee_id: 61}) {
    nominee {
      address
      ended
    }
  }
}
```
should return ended false

3. run again and should get a dupe error
4. go into sql or use a mutation to fudge the user id of the vouch, i ran `update vouches set voucher_id=607;` where my real userid is 606
5. run vouch mutation again, now ended should be true and user is in the circle (i just verified in DB cuz UI is giving me hell)

### DIFF
--- a/api-lib/HttpError.ts
+++ b/api-lib/HttpError.ts
@@ -6,6 +6,7 @@ export interface HttpError extends Error {
 }
 
 export class InternalServerError extends Error implements HttpError {
+  // TODO: it seems like hasura doesn't like 500s so we should pick a 4xx as the default error
   httpStatus = 500;
 }
 
@@ -13,19 +14,35 @@ export class PayloadTooLargeError extends Error implements HttpError {
   httpStatus = 413;
 }
 
-export function ErrorResponse(res: VercelResponse, error: any): VercelResponse {
-  const statusCode = error.httpStatus || 500;
-  return ErrorResponseWithStatusCode(res, error, statusCode);
+export class NotFoundError extends Error {
+  httpStatus = 404;
 }
 
-export function ErrorResponseWithStatusCode(
+export class ForbiddenError extends Error {
+  httpStatus = 403;
+}
+
+export class BadRequestError extends Error {
+  httpStatus = 400;
+}
+
+export function errorResponse(res: VercelResponse, error: any): VercelResponse {
+  const statusCode = error.httpStatus || 500;
+  return errorResponseWithStatusCode(res, error, statusCode);
+}
+
+export function errorResponseWithStatusCode(
   res: VercelResponse,
   error: any,
   statusCode: number
 ): VercelResponse {
   return res.status(statusCode).json({
-    error: `${statusCode}`,
     message: error.message || 'Unexpected error',
+    // included at this level for backwards compat w/ older hasura, perhaps can remove
+    code: `${statusCode}`,
+    extensions: {
+      code: `${statusCode}`,
+    },
   });
 }
 
--- a/api/hasura/actions/vouch.ts
+++ b/api/hasura/actions/vouch.ts
@@ -1,0 +1,153 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { getUserFromProfileId } from '../../../api-lib/findUser';
+import { gql } from '../../../api-lib/Gql';
+import {
+  BadRequestError,
+  errorResponse,
+  ForbiddenError,
+  InternalServerError,
+  NotFoundError,
+} from '../../../api-lib/HttpError';
+import { Awaited } from '../../../api-lib/ts4.5shim';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+import {
+  composeHasuraActionRequestBodyWithSession,
+  HasuraUserSessionVariables,
+  vouchInput,
+} from '../../../src/lib/zod';
+
+type Voucher = Awaited<ReturnType<typeof getUserFromProfileId>>;
+type Nominee = Awaited<ReturnType<typeof getNominee>>;
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  try {
+    const {
+      input: { payload: input },
+      session_variables: sessionVariables,
+    } = composeHasuraActionRequestBodyWithSession(
+      vouchInput,
+      HasuraUserSessionVariables
+    ).parse(req.body);
+
+    const { hasuraProfileId: voucherProfileId } = sessionVariables;
+    const { nominee_id: nomineeId } = input;
+
+    // validate that this is allowed
+    const { voucher } = await validate(nomineeId, voucherProfileId);
+
+    // vouch and build user if needed
+    const updatedNominee = await vouch(nomineeId, voucher);
+
+    return res.status(200).json({ id: updatedNominee.id });
+  } catch (e: any) {
+    return errorResponse(res, e);
+  }
+}
+
+async function validate(nomineeId: number, voucherProfileId: number) {
+  // Get the nominee
+  const nominee = await getNominee(nomineeId);
+
+  // make sure circle exists
+  if (!nominee.circle) {
+    throw new NotFoundError('circle not found');
+  }
+
+  // Check if voucher exists in the same circle as the nominee
+  // TODO: this uses assert for error handling
+  const voucher = await getUserFromProfileId(
+    voucherProfileId,
+    nominee.circle_id
+  );
+
+  // If circle only allows giver to vouch, make sure voucher is a giver
+  if (nominee.circle.only_giver_vouch && voucher.non_giver) {
+    throw new ForbiddenError(
+      "voucher is a 'non-giver' so is not allowed to vouch"
+    );
+  }
+
+  // make sure the nomination period hasn't ended
+  if (nominee.ended) {
+    throw new BadRequestError('nomination has already ended for this nominee');
+  }
+
+  // TODO: could this be handled by a unique index in the vouches table?
+  // Check if voucher already has an existing vouch for the nominee
+  if (nominee.nominated_by_user_id === voucher.id) {
+    throw new ForbiddenError(
+      "voucher nominated this nominee so can't additionally vouch"
+    );
+  }
+
+  const { vouches } = await gql.getExistingVouch(nomineeId, voucher.id);
+
+  if (vouches.pop()) {
+    throw new ForbiddenError('voucher has already vouched for this nominee');
+  }
+
+  return {
+    voucher,
+  };
+}
+
+async function vouch(nomineeId: number, voucher: Voucher) {
+  // vouch for the nominee
+
+  // this inserts the vouch and also fetches the nominee with updated vouch count
+  const insert_vouches = await gql.insertVouch(nomineeId, voucher.id);
+  if (!insert_vouches?.nominee) {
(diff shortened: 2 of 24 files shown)


src/lib/gql/GqlHasuraAdmin.ts
```diff
@@ -520,4 +543,59 @@ export class Gql {
...
+    const { insert_vouches_one } = await this.q('mutation')({
+      insert_vouches_one: [
+        {
+          object: {
+            nominee_id: nomineeId,
+            voucher_id: voucherId,
+          },
+        },
+        {
+          id: true,
```
zashton, line 575: we can return the `nominee` object here so you wouldn't need to refetch after every insert

src/lib/zod/index.ts
```diff
@@ -62,6 +62,13 @@ export const uploadCircleImageInput = z
   })
   .strict();
 
+export const vouchInput = z
+  .object({
+    nominee_id: z.number(),
+    circle_id: z.number(),
```
zashton, line 68: Don't think `circle_id` is necessary, you can get the `circle_id` from the `nominee` object. 
theres no validation whether the nominee belongs to the `circle_id` currently

api/hasura/actions/vouch.ts
```diff
@@ -0,0 +1,181 @@
...
+    circleId: nominee.circle_id,
+    sanitize: true,
+    channels: {
+      // TODO: figure out if these need to be conditionalized?
+      discord: true,
+      telegram: true,
+    },
+  });
+  return updatedNominee;
+}
```
topocount, line 170: this should be an event trigger, one for discord, one for telegram. This is important because we don't want the notification logic to cause the action/mutation to fail in the frontend.

See other notifications event_triggers folder to see how to set this up.

api/hasura/actions/vouch.ts
```diff
@@ -0,0 +1,181 @@
...
+async function getNominee(nomineeId: number) {
+  const nom = await gql.getNominee(nomineeId);
+  if (!nom.nominees_by_pk) {
+    throw `nominee ${nomineeId} not found`;
+  }
+  return nom.nominees_by_pk;
+}
+
+type voucher = Awaited<ReturnType<typeof getUserFromProfileId>>;
+type nominee = Awaited<ReturnType<typeof getNominee>>;
```
topocount, line 181: Nit: put these towards the top of the file, after the imports.

topocount, line 181: Types should be capitalized
```suggestion
type Voucher = Awaited<ReturnType<typeof getUserFromProfileId>>;
type Nominee = Awaited<ReturnType<typeof getNominee>>;
```

api/hasura/actions/vouch.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (vouches.pop()) {
+    throw new ForbiddenError('voucher has already vouched for this nominee');
+  }
+
+  return {
+    voucher,
+  };
+}
+
+async function vouch(nomineeId: number, voucher: voucher) {
```
topocount, line 92:
```suggestion
async function vouch(nomineeId: number, voucher: Voucher) {
```

api/hasura/actions/vouch.ts
```diff
@@ -0,0 +1,181 @@
...
+  // if there are enough nominations, go ahead and add the user to the circle
+  const nomCount = nominee.nominations_aggregate.aggregate?.count || 0;
+  if (nominee.vouches_required - 1 <= nomCount) {
+    return await convertNomineeToUser(nominee);
+  }
+
+  return nominee;
+}
+
+async function convertNomineeToUser(nominee: nominee) {
```
topocount, line 124:
```suggestion
async function convertNomineeToUser(nominee: Nominee) {
```

api-lib/HttpError.ts
```diff
@@ -5,13 +5,26 @@ export interface HttpError extends Error {
 }
 
 export class InternalServerError extends Error implements HttpError {
+  // TODO: it seems like hasura doesn't like 500s so we should pick a 4xx as the default error
   httpStatus = 500;
```
topocount, line 9: I humbly suggest [418](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418)

api/hasura/actions/vouch.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (profiles.length == 0) {
+    const addedProfiles = await gql.insertProfiles([
+      {
+        address: nominee.address,
+      },
+    ]);
+    if (!addedProfiles) {
+      throw new InternalServerError('unable to add profile for new user');
+    }
+  }
```
topocount, line 150: This is already handled by an event trigger and is not needed here.

src/lib/gql/index.ts
```diff
@@ -99,12 +99,43 @@ export function getGql(url: string, getToken: () => string) {
       }
     );
 
+  const getUserFromProfileIdWithCircle = async (
```
topocount, line 102: this is pretty slick ngl

topocount (changes requested):
Looks sound! The notifications should be converted to event triggers, otherwise the logic here looks good, save for the comments below.

api/hasura/actions/vouch.ts
CryptoGraffe: implemented these event triggers, not sure how to test yet

api/hasura/actions/vouch.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (!addedUser) {
+      throw new InternalServerError('unable to add user');
+    }
+    userId = addedUser.id;
+  }
+
+  // The profile is automatically created by the createProfile event trigger, if needed
+
+  // attach the user id to the nominee, and mark the nomination ended
+  const updatedNominee = await gql.updateNomineeUser(nominee.id, userId);
```
topocount, line 135: It's too bad we can't make this atomic. Maybe we could add a FK relationship in the other direction (a column in `users` that backlinks to a `nominee` row) instead of the current relationship. Then a nominee's success would be determined by a join in the gql query and we'd be able to do this atomically.

hasura/metadata/databases/default/tables/public_nominees.yaml
```diff
@@ -108,3 +108,35 @@ event_triggers:
...
+      columns:
+      - id
+      - name
+      - user_id
+  headers:
+  - name: verification_key
+    value_from_env: HASURA_EVENT_SECRET
+  name: nomineeVouchedInTelegram
+  retry_conf:
+    interval_sec: 3600
```
topocount, line 139: I think an hour is okay, especially if there is an outage on the other end (like there is on discord right now).

topocount (approved):
Good work dude!

When are you formally opening this up for review?

topocount (commented):
For some reason my nominee row was removed while completing the test sequence.

api-lib/okResponse.ts
```diff
@@ -0,0 +1,7 @@
+import { VercelResponse } from '@vercel/node';
+
+const okResponse = function (res: VercelResponse, json: unknown) {
+  return res.status(200).json(json);
+};
```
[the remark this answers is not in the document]
CryptoGraffe: Yeah fine ok. I like this stuff but its fine to remove. Allows single point for debug/analytics bla bla and i hate seeing 200 everywhere but its personal pref. Not important to me. BTW this is "formally open for review" there is no draft status anymore.

hasura/metadata/databases/default/tables/public_nominees.yaml
CryptoGraffe: i think it makes sense for these notis

api/hasura/actions/vouch.ts
CryptoGraffe: Can we TODO this and add a ticket for later? This is a general concern I have for TX stuff that mixes DB+bizlogic

api-lib/okResponse.ts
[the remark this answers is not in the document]
CryptoGraffe: this is deleted now

api/hasura/actions/vouch.ts
CryptoGraffe: I'm also unsure if its true that user ALWAYS has nominee. anyway i think this is a problem for later?

src/lib/gql/zeusHasuraAdmin/const.ts
```diff
@@ -481,6 +481,20 @@ export const AllTypesProps: Record<string, any> = {
       required: true,
     },
   },
+  VouchInput: {
+    circle_id: {
```
zashton, line 485: missed running yarn generate

api/hasura/actions/vouch.ts
zashton: i don't think `user_id` in `nominees` table is used currently anywhere , i added it just incase someone wanted to do analytics with this so it might not be that critical

Yes a user can be added directly by admin bypassing the vouching process so it wouldn't be linked to any nominee

zashton (changes requested):
main issue is the dupe notifications caused by similar event triggers, otherwise the rest is good for me.

api/hasura/event_triggers/utils/handleNomineeVouchedInMsg.ts
```diff
@@ -0,0 +1,23 @@
...
+  channels: { discord?: boolean; telegram?: boolean }
+) {
+  const {
+    event: { data },
+  } = payload;
+
+  // If user_id was not set, and now it is, that means it is a nominee that just became vouched in
+  if (!data.old.user_id && data.new.user_id) {
+    await sendSocialMessage({
+      message: `${data.new.name} has received enough vouches and is now in the circle!`,
```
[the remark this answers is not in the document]
CryptoGraffe: yeah haha we added them at same time, i rebased after your merge. will delete mine/merge if needed.

src/lib/gql/zeusHasuraAdmin/const.ts
CryptoGraffe: fixed

api/hasura/event_triggers/utils/handleNomineeVouchedInMsg.ts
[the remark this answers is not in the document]
CryptoGraffe: Ok i deleted mine they were truly redundant

zashton (approved):

PR #575, opened 2022-02-26
Jest + testchain

### LINKED ISSUE #527: Unified environment for automated tests
At the moment, there are two separate test suites in the front-end app: one for hardhat, which tests contracts, and one for the React app, which tests components and other UI code. We need to set up a test environment that will allow a single test to do both.

levity:
closes #527

## TL;DR

Does `yarn test:ci` work for you locally on this branch? How about `yarn test`?

## Change log / notes for reviewers

- Code in `hardhat/dist` is auto-generated, so there's no need to read it too closely
- I reduced some of the unnecessary prefixing of names with `Ape`
- There are some changes to eslint, tsconfig, craco config files: these are for configuring `ts-auto-mock`, which can make mock objects from type definitions. I didn't end up using it in this PR but it'll come in useful in the future.

- This makes all tests run against ganache by default. When you run `yarn test` in development, even in interactive mode, a ganache instance will start up before Jest and shut down after you quit.
  - If you want to save a few seconds, you can run `./hardhat/scripts/start-ganache.sh --deploy` in a separate terminal. Then when you run `yarn test` it'll reuse that instance.
  - set `TEST_ON_HARDHAT_NODE` before running `yarn test` if you want to switch to running tests against hardhat's node instead. The script will still start up / look for ganache; it just won't use it (FIXME?).

- We can now store deployment data for hardhat's built-in node and ganache separately. Ganache's default chain ID and port are set to 1338 and 8546, so you can add it as a separate network in Metamask.
  - Use `HARDHAT_CHAIN_ID`, `HARDHAT_PORT`, `HARDHAT_GANACHE_CHAIN_ID`, `HARDHAT_GANACHE_PORT` to configure the chain IDs and ports of the two testchains.
  - Mainnet addresses can be added to `hardhat/deploymentInfo.json` under the mainnet chain id 1. When using a testchain, we'll fall back on using these if the testchain doesn't have its own defined. This makes the code play nice with mainnet forks.
  - When a new contract is deployed on the testchain, we can pick up the new address and put it into `hardhat/deploymentInfo.json` by running `yarn hardhat:codegen`. But then the whole lib needs to be rebuilt & re-yarn-linked into the main project. 
    - `./scripts/rebuild-hardhat.sh` automates this process. It's been modified so it doesn't try to start up its own node by default anymore. So the basic workflow is:
      - make a change to a contract
      - restart your testchain (`FORK_MAINNET=1 yarn hardhat:dev` or `./hardhat/scripts/start-ganache.sh --deploy`)
      - run `rebuild-hardhat`
      - restart `vercel dev` and/or `yarn test`


src/hooks/useVaultFactory.ts
```diff
@@ -1,60 +1,54 @@
...
 import { useApeSnackbar } from 'hooks';
-import { useCurrentOrg } from 'hooks/gql';
 import { useFakeVaultApi } from 'recoilState/vaults';
 
 import { useContracts } from './useContracts';
 
 import { IVault } from 'types';
 
-export function useVaultFactory() {
+export function useVaultFactory(orgId?: number) {
```
levity, line 12: Adding `orgId` as an argument instead of using `useCurrentOrg` was a key step in making this hook testable. Otherwise it depends upon not only the GraphQL API but a long chain of Recoil atoms.

src/hooks/useContracts.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <TestWrapper>
+      <Harness />
+    </TestWrapper>
+  );
+});
+
+test('set up contracts', async () => {
+  let balance: BigNumber | undefined;
+
+  const Harness = () => {
```
levity, line 28: this is my first draft of a decent pattern for tests that depend upon hook &/or component interaction: 
- write a wrapper component (here `Harness` in the sense of "test harness")
- wrap it in `TestWrapper`
- set the `withWeb3` prop to set up the `useWeb3React` hook with a testchain connection
- set variables that were declared in test-local scope, so they can be checked with `await waitFor(...)` later. `waitFor` will retry if anything in its callback throws an exception, until it times out

src/services/contracts.ts
```diff
@@ -4,60 +4,57 @@ import {
...
+  // TODO this might not be quite the right way to do this, as the
+  // signer/provider used to create the contracts also has a network associated
+  // with it
+  chainId: number;
 
   signerOrProvider: SignerOrProvider;
 
   constructor(
     contracts: {
-      usdc: ERC20;
```
levity, line 37: this was just given its own prop for convenience while this file was being developed--now we can use `getToken` below instead

levity: `apeToken` below was removed entirely because we don't have a need for it yet

hardhat/scripts/start-ganache.sh
```diff
@@ -11,14 +11,17 @@ if [ -f "$DOTENV_FILE" ]; then
   export $(cat $DOTENV_FILE | sed 's/^#.*$//' | xargs)
```
crabsinger, line 11: I had some unfilled in ENV_vars in my .env like
```
TELEGRAM_BOT_TOKEN=<your Telegram testing bot token>
```

and the `<` broke the script with error:
```
 ./hardhat/scripts/start-ganache.sh --deploy --exec ./scripts/ci.sh
./hardhat/scripts/start-ganache.sh: line 11: export: `token>': not a valid identifier
error Command failed with exit code 1.
```

So maybe we should change the default ENV_VAR for that one to not have `<` or anything that breaks this script when unused.

.env.example
```diff
@@ -18,6 +15,15 @@ LOCAL_POSTGRES_PASSWORD=postgrespassword
 LOCAL_POSTGRES_DATABASE=postgres
 LOCAL_HASURA_ADMIN_SECRET=admin-secret
 
+# Hardhat
+HARDHAT_CHAIN_ID=1337
+HARDHAT_PORT=8545
+HARDHAT_GANACHE_CHAIN_ID=1338
+HARDHAT_GANACHE_PORT=8546
+HARDHAT_FORK_BLOCK=13500000
+HARDHAT_OWNER_ADDRESS=0xYOUR_DEV_ADDRESS
```
crabsinger, line 24: I'm so lazy, I wish this would use the LOCAL_SEED_ADDRESS here too by default

hardhat/dist/hardhat.config.js
```diff
@@ -62,9 +62,16 @@ exports.unlockSigner = unlockSigner;
...
-    url: constants_1.ETHEREUM_RPC_URL,
-    blockNumber: constants_1.FORKED_BLOCK,
+const sharedNetworkSettings = {
+    live: false,
+    allowUnlimitedContractSize: true,
+    gas: 'auto',
+    gasPrice: 'auto',
+    gasMultiplier: 1,
+    accounts: {
+        mnemonic: 'coordinape',
```
crabsinger, line 72: are these account ever used? I think they get some ETH prefilled when the network spins up. Seems like this will be a different address than LOCAL_SEED_ADDRESS or HARDHAT_OWNER_ADDRESS ...

It would be nice if the whatever is the devs primary eth address has some ETH for txs automatically

crabsinger (approved):
Awesome to see all the the pieces of our app come together.

`❯ ./hardhat/scripts/start-ganache.sh --deploy` and then in seprate terminal
`❯ yarn test:ci` succeeds 11 passing, 2 pending.

`yarn test` also works for me
deploys and then sets up the "watcher".

crabsinger (approved):

hardhat/scripts/start-ganache.sh
levity: Ah, good call

hardhat/dist/hardhat.config.js
levity: yeah, that's what HARDHAT_OWNER_ADDRESS does

.env.example
levity: yeah i should probably combine those

src/hooks/useContracts.test.tsx
levity: @teeolendo even though i've already merged this, it'd be good for you to take a look at the comment above and this test file to help you write your own frontend<->contract integration tests

PR #576, opened 2022-02-28
User Logout hasura

### LINKED ISSUE #563: Migrate Logout v2/logout
## Summary:
Migrate Logout Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `<POST v2/logout>` to Hasura.

**Middleware:** middleware event secret

-

**Logic:**
Hard delete token in  `personal_access_tokens` that is associated with the user's `profile`

**DB Tables:**
profiles, personal_access_tokens

#### Notes:
FE will need to purge the auth token from the cache as well.

zashton:
### Test Plan

1. Find out the `id` of the profile you are using to authenticate with hasura
2. Go to `personal_access_tokens` and search the `id` against the `tokenable_id` field
3. This mutation will purge your access token for hasura so you might want to copy down the data of the row first to insert back later
4. Execute the the mutation 
```gql
mutation MyMutation {
  logout_user {
    profile {
      address
      id
    }
  }
}
```
5.  Verify you cannot query hasura with your existing token anymore, you can access your DB to check the row is indeed deleted.
6. Add back your token and make sure you can access hasura again.

Closes #563


api/hasura/actions/logout_user.ts
```diff
@@ -0,0 +1,49 @@
...
+        message: 'Invalid input',
+        code: '422',
+      });
+    } else if (err instanceof GraphQLError) {
+      return response.status(422).json({
+        extensions: err.response.errors,
+        message: 'GQL Query Error',
+        code: '422',
+      });
+    }
```
topocount, line 41: in what cases would this be thrown? Wouldn't this only be an issue if the query is poorly structured?

topocount (approved):
this looks good to me!

api/hasura/actions/logout_user.ts
```diff
@@ -0,0 +1,49 @@
...
+  try {
+    const { hasuraProfileId } = HasuraUserSessionVariables.parse(
+      request.body.session_variables
+    );
+    await gql.q('mutation')({
+      delete_personal_access_tokens: [
+        {
+          where: {
+            tokenable_id: { _eq: hasuraProfileId },
+            tokenable_type: { _eq: 'App\\Models\\Profile' },
```
CryptoGraffe, line 19: yikes is this some laravel legacy syntax

api/hasura/actions/logout_user.ts
```diff
@@ -0,0 +1,49 @@
...
+          },
+        },
+        {
+          affected_rows: true,
+        },
+      ],
+    });
+    return response.status(200).json({ id: hasuraProfileId });
```
CryptoGraffe, line 27: in the future some of these type of calls might be better served by a 204 no content status, but i'm not sure how hasura handles that

api/hasura/actions/logout_user.ts
CryptoGraffe: Curious about this too. would be nice to use the newer HttpError stuff for these

hasura/metadata/actions.graphql
```diff
@@ -65,3 +69,7 @@ type update_profile_response {
   id: Int!
 }
 
+type logout_response {
```
CryptoGraffe, line 72: instead of the dupe responses we have now for various x_response that just return id, maybe it would be nice to have an id_response type

CryptoGraffe (approved):
l2gm just some minor comments

api/hasura/actions/logout_user.ts
zashton: removed this , agreed this is redundant

api/hasura/actions/logout_user.ts
zashton: yeah in future we might move towards JWT tokens for auth

hasura/metadata/actions.graphql
zashton: i think thats a good idea

api/hasura/actions/logout_user.ts
zashton: i think hasura accepts any 2xx status

PR #582, opened 2022-03-02
Standardize backend naming

topocount:
# Description
Rules are now enforced by the new eslint configs but I'll attempt to lay them out here

## GraphQL Schema
- functions/actions are camelCase
- Types are PascalCase
- arguments are snake_case so they match the postgres db columns

## Files
- Serverless function file names are all camelCase, to match the function names
- event-trigger directory kebab-case is maintained, but could be changed with enough sentiment. This is how hasura cases it in the metadata, so i thought grepping around would be easier if kept this way.
- 

# Todos

- [x] add linting rule for path names and graphql schema
- [x] make path names camelCase
- [x] migrate from `object` to `payload` for the mutation param

this is the regex i used to convert files from snake to camelcase:
```sh
❯ perl-rename 's/_([a-z])/\U$1/g' *.ts
```


zashton:
realised the current linting doesn't include api-lib folder , maybe we can add that as well.

hasura/metadata/actions.graphql
```diff
@@ -83,19 +67,35 @@ input AdminUpdateUserInput {
   role: Int
 }
 
-type update_profile_response {
+type CreateCircleResponse {
+  id: Int!
```
zashton, line 71: Would just need `id` for this `CreateCircleResponse`

zashton (changes requested):
Requires a `yarn generate` to be able to run the hasura console.

hasura/metadata/actions.graphql
```diff
@@ -1,44 +1,32 @@
...
-  upload_profile_background(
-    object: upload_image_input!
-  ): update_profile_response
+  uploadProfileBackground(payload: UploadImageInput!): UpdateProfileResponse
 }
 
-input create_circle_input {
+input CreateCircleInput {
   user_name: String!
   address: String!
```
zashton, line 31: not sure how the `address` came back, we removed it during the address spoofing fix

CryptoGraffe (approved):
Looks good to me. Thanks for doing all this.

zashton (approved):
All good, thanks for doing this ! , lgt hasura-next

Issue #314, opened 2021-11-23
Disconnecting wallet failing to fully disconnect


reeserj:
When using coordinape with multiple wallets, clicking the disconnect button disconnects the wallet, but then when you try to connect with another wallet, the prior wallet automatically connects without giving you an option to connect with a different wallet.

1. ![image](https://user-images.githubusercontent.com/82838264/142958257-f62f3706-661d-4483-b2de-294095cf70d7.png)
2. ![image](https://user-images.githubusercontent.com/82838264/142958278-32771739-3ec7-4f55-8a50-1dbb85e49c1b.png)
3. ![image](https://user-images.githubusercontent.com/82838264/142958304-bbefb59f-70de-45fc-a53d-1758fcd7fe9e.png)
4. ![image](https://user-images.githubusercontent.com/82838264/142958325-5820ee09-bba8-4473-a58b-a1a0d8c76f7f.png)

The same behavior exists in Walletconnect and Metamask

PR #586, opened 2022-03-04
Second wallet spike

### LINKED ISSUE #553: Spike: Nested Wallet VS. Second App to handle other wallet funding
**Specification:**
Goal: Enable smooth funding process for vaults using Safe Waller/HW Wallet or other wallet beside hot login wallet. Ensure ability to handle vault deposit and withdraw very smoothly for the human user. 

1. Attempt/Investigate opening a nested wallet connection in the app with as little disruption to the app state and current login. 
2. Make Vercel open a second app (micro-frontend) - A form to allow a different wallet (other than logged in wallet) to deposit or withdraw. 

Possibly talk wiht Gnosis people or developer DAO along the way :)

levity:
Closes #553: This is a proof-of-concept for allowing the user to connect a second wallet via WalletConnect, while viewing the modal for depositing funds into a vault. Unfortunately it can't be tested end-to-end until we have a _public_ testnet deployment, due to the limitations of WalletConnect. Even though I've written code to get your second wallet's balance & use it to send the deposit tx, all you can do is see the second wallet's address. But I think this suffices for a spike.

It also contains a grab-bag of minor improvements:

- fix the `yarn hardhat mint` task to work with Ganache, & send DAI as well as USDC
- add `yarn hardhat balance` to check an address's balance on the testnet
- upgrade `@web3-react/walletconnect-connector` (should test WalletConnect in the preview app)
- use stitches a bit more
- fix a React warning message
- make the default ui/Button style not be white-on-white 😅


levity:
The standard WalletConnect behavior still seems to work. The only difference is that choosing "Disconnect" also unlinks the wallet from the app, so that the next time you click "Connect" you have to scan the QR code again. I think this is actually good, because prior to this, there was no way to switch between different WalletConnect addresses unless you went into the wallet interface and unlinked from Coordinape there. See e.g. #314.

teeolendo (approved):
Difficult to test this WalletConnect but the core logic seems solid. We'll have to come back to this when we go live.

PR #587, opened 2022-03-04
Update balance after deposit

levity:
Update the vault's balance immediately after a deposit succeeds, and also on every block.


teeolendo (approved):
LGTM

PR #590, opened 2022-03-05
Refactor refundPendingGift Event Trigger

topocount:
I was really not a fan of my initial work on this and refactored it
in preparation for the non-admin `updateUser` action.
so:
- the separation of concerns is cleaner
- all accounting updates are atomic and clean
- it's still an nice idiomatic business logic event handler

Now all changes occur in a single transaction when a user is
updated to be non_giver or non_receiver. This includes refunds to or
from counterparties who have either given or received tokens from the
modifier user.

Additionally, no pending_gift or give_token accounting logic leaks
from this event handler. Anything related to GIVE token accounting
during and user updates is handled here.

test plan:
0. purge all seeded pending gifts for a chosen epoch from the pending_token
   gifts table. If you don't do this, the accounting won't balance with your
   allocations.
1. allocate GIVE to four user during an epoch in the UI. add a note for
   two of them.
2. run adminUpdateUser `non_receiver: true` on 1 user with and 1 user without
   a note. Verify that the pending_token_gift with the note is not
   delete, but set to 0, while the other entry is deleted.
3. run adminUpdateUser `non_giver: true` on your user in the circle and
   verify the same behavior with the notes.
4. check the balances of all involved users and ensure they have
   returned to their original values.

### DIFF
--- a/api/hasura/actions/adminUpdateUser.ts
+++ b/api/hasura/actions/adminUpdateUser.ts
@@ -65,10 +65,6 @@ async function handler(request: VercelRequest, response: VercelResponse) {
       });
     }
 
-    // invariant: there can only be one active epoch in a circle
-    // at a given time.
-    const currentEpoch = user.circle.epochs.pop();
-
     // Update the state after all external validations have passed
 
     const mutationResult = await gql.q('mutation')({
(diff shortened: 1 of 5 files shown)


zashton (approved):
Everything works correctly, had the issue you described where seeding(importing) data into the DB makes the event triggers go crazy and hangs the db.

topocount:
> Everything works correctly, had the issue you described where seeding(importing) data into the DB makes the event triggers go crazy and hangs the db.

@zashton good to know! I have a very low-lift fix that I'll push shortly.

PR #591, opened 2022-03-05
Small Simplication of the adminUpdateUser mutation

topocount:
The logic was unnecessarily complex so I cut those pieces out.
It is functionally identical, except it now doesn't touch the
give_token_remaining column if there is no change to make.

merge plan: merge to hasura-next after #590 is merged


zashton (approved):
looks good

PR #596, opened 2022-03-07
Quickfix for Seeding the DB

topocount:
Seeding the db with the vercel dev stateless functions causes a ton
of database events to be thrown, which eats up the processor and a ton
of memory. Typical options when this happens include either (a)
  restarting the machine or (b) waiting for the everything to pass.

This is obviously not a great solution, and a more resilient solution
should/could be implemented by refactoring the db-seed scripts to yield
sql-based hasura seeds, but this requires more discussion around what a
desirable seed script would be.

Test plan:

Try out this sequence for yourself and ensure your machine doesn't go
haywire.


crabsinger (approved):
ran it and it worked for me 👍

zashton (approved):

PR #597, opened 2022-03-07
don't re-render all ProfileCards on every allocation change

levity:
1. wrap ProfileCard in React.memo

2. prevent ProfileCard's props from changing unnecessarily
    - we can't pass updateGift from the useAllocation hook to ProfileCard, because it's redefined every time any user's allocation or note is changed, causing all ProfileCards to re-render
    - instead, I created a new selectorFamily in order to give each ProfileCard a setter that's scoped to its user, so it only re-renders when its own user changes

### DIFF
--- a/src/recoilState/allocation.ts
+++ b/src/recoilState/allocation.ts
@@ -110,6 +110,56 @@ export const rLocalGifts = atomFamily<ISimpleGift[], number>({
   default: [],
 });
 
+// WTFLOL: selectorFamily only allows a single number parameter, so we use this
+// and the wrapper below to combine multiple params into a single one
+const PARAM_OFFSET = 1000000;
+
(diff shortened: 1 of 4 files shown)


levity:
@teeolendo i just tested this by putting myself into the Bankless circle (600+ users) on staging, & i can type & change allocations without noticeable delay now

src/recoilState/allocation.ts
```diff
@@ -110,6 +110,56 @@ export const rLocalGifts = atomFamily<ISimpleGift[], number>({
   default: [],
 });
 
+// WTFLOL: selectorFamily only allows a single number parameter, so we use this
```
teeolendo, line 113: Lol

teeolendo (approved):

Issue #480, opened 2022-01-27
Build Submit Distribution: Part 3 (Submit and Save)


spydermonkee:
### Specification

- after the form input is valid and the new `MerkleDistributorInfo` has been generated ([see here](https://www.notion.so/On-Chain-Vaults-27a8a1da30f04967a240df9dd4ddae76)), submitting this form sends a tx, `ApeDistributor.uploadEpochRoot`
- the distribution data needs to be stored in the backend ([see here](https://www.notion.so/On-Chain-Vaults-27a8a1da30f04967a240df9dd4ddae76))
    - maybe store it before submission with a link to the tx and a status property (e.g. `pending` or `mined`) — reduce the risk of lost data
       
https://www.figma.com/file/wXF7xGVv1j2SqwWfgbamS8/Coordinape-Design-v8?node-id=2952%3A4783

Also see: #468, #469

PR #604, opened 2022-03-09
Save Vault Distributions from an Epoch.

### LINKED ISSUE #507: Build (Vaults) Submit Distribution: Part 2.3 Store Distribution Data in Backend
### Specification

- when submitting an epoch root, we need to store all the info generated--the "balance map" that contains amounts & proofs, the total amount, etc--in the backend, so we can use it again later. claimers will need the proofs to claim their distributions, and the circle admin will need the full list of amounts for generating the root for the next epoch

- adding this to hasura
  - new table, e.g. `distributions`. suggested columns below:
    - foreign key to epoch
    - JSON column for the entire merkle distributor info object ([e.g.](https://github.com/coordinape/coordinape/blob/main/src/lib/merkle-distributor/fixtures.json#L9))
      - feel free to break this out into more columns if appropriate
    - standard timestamp column (created_at)
    - foreign key to user who submitted it
  - create migration to create this new table inside local hasura console, which should generate migration files that will get run for the PR against a temporary hasura preview app
  - write mutation that stores the data into this new table

teeolendo:
Key changes in this PR are:
- Added vaults, vaults_transactions, distributions, claims
- The mutations for these tables aren't written inside of an action. There was no reasonable use-case that I could come up with that would justify the use of action at this time. Happy to review this based on feedback.
- Added a mutation hook called `useSaveDistribution` which is contained in the `useGetAllocations`. Not the best location but I'm happy to change it.
- The `useSaveDistribution` hook also saves claims that are passed from the `createDistribution` hook which generates the merkle tree for an epoch

This PR closes #507 and #480.


hasura/migrations/default/1646251832285_create_table_public_distributions/up.sql
```diff
@@ -1,13 +1,13 @@
 CREATE TABLE "public"."distributions"
 (
-  "id" bigint NOT NULL,
+  "id" bigserial NOT NULL,
```
zashton, line 3: The migrations are incremental, you will need to do a `ALTER` table if you are modifying an existing table otherwise the change won't be applied to the DB. You can use the hasura console to edit the tables, it should auto generate the migration sql files for you. Then you can do a `migration squash --from <timestamp>` if needed

levity: oh right, good point. the original versions of these have already been run in staging

levity:
Replaced by #627

PR #608, opened 2022-03-10
Cleanup implementation of graphql queries/mutations

META-DREAMER:
Improved naming and code organization of how queries and mutations were being done on the client side + moved Admin client stuff to api-lib


src/pages/DistributePage/DistributePage.tsx
```diff
@@ -37,16 +37,14 @@ function DistributePage() {
...
+  const vaults = useVaults(currentOrg.data?.id);
   const { uploadEpochRoot } = useDistributor();
   const [selectedVault, setSelectedVault] = useState<IVault | undefined>();
   const { getYVault } = useVaultWrapper(selectedVault as IVault);
   const { apeError } = useApeSnackbar();
 
   const { isLoading, isError, data } = useGetAllocations(Number(epochId));
-  const { myUser: currentUser } = useCircle(
-    data?.epochs_by_pk?.circle?.id as number
-  );
```
META-DREAMER, line 49: @teeolendo FYI we shouldnt ever cast types like this because it could cause runtime errors. `id` here would be actually be `undefined` here until the query resolves, so instead of casting it we should update the downstream code to handle the undefined value. You can see I made that update here:
https://github.com/coordinape/coordinape/pull/608/files#diff-fe13ed674a354105a9e96354ec08b53830b3a8f3653f90c4eb1311f6a464e2a1R182

CryptoGraffe (approved):
lg2m

PR #584, opened 2022-03-03
Dework integration

0xchomper:
Video demo: https://www.loom.com/share/2d4ae57f251d4b069ca67914c95086d1

Demo deployment to test against Coordinape staging environment: https://coordinape-virid.vercel.app

image

PR #609, opened 2022-03-10
Org View: Change nav

### LINKED ISSUE #555: Org View: Change nav
**Specification**
- Add Circles Link to left part of nav bar
- Add History to nav bar (link to existing History page)
- Right Justify existing nav 
- Add Breadcrumb when in circle ("ORG NAME">"CIRCLE NAME")
- Remove redundant links from the Avatar Dropdown menu ("My History", "My Team", "Add Circle", "My Epoch", Circles - Note that the circle selection is replaced by https://github.com/coordinape/coordinape/issues/554)
 

**Note that the "Funding" nav item will not be present in the first version of ORG VIEW that gets released before VAULTS. (Even though the figma design shows the "Funding" nav item currently.)

https://www.figma.com/file/wXF7xGVv1j2SqwWfgbamS8/App-1.0?node-id=3987%3A15460

**Note that the the whole milestone of [ORG VIEW  ](https://github.com/coordinape/coordinape/milestone/5) should be released together on the app. These changes need to be released as one new feature :)

MarcogodoyV:
Made the changes to de nav, adding the CirclesPage to make the Circles Nav item work, redundant items removed from the Avatar menu (rigth now added to the CirclesPage in order to have a way to switch circles, knowing that after everything is done it's going to work different), I also changed the mobile view. This PR can fix #555


levity:
hi @MarcogodoyV, thanks for the PR. can you please merge changes from the base `org-view` branch back into your branch? we recently made a change to the Vercel deployment process that requires a change from there.

levity:
> hi @MarcogodoyV, thanks for the PR. can you please merge changes from the base `org-view` branch back into your branch? we recently made a change to the Vercel deployment process that requires a change from there.

Never mind, looks like I was able to do it :)

src/routes/paths.ts
```diff
@@ -143,12 +149,24 @@ export const getMainNavigation = ({
   return mainItems;
 };
 
+export const getAdminNavigation = (): INavItem[] => [
```
levity, line 152: I know you're probably just following the existing pattern, but these data structures are small & simple enough that they can just be embedded directly in the components that use them.

src/routes/paths.ts
```diff
@@ -45,6 +45,7 @@ export const paths = {
   vaultTxs: (id: string) => `${paths.vaults}/${id}/txs`,
   adminCircles: '/admin/circles',
   connectIntegration: '/connect-integration',
+  circles: '/circles',
```
levity, line 48: feel free to start using the paths here instead of the getter methods

levity (changes requested):
Looks pretty good so far! Just a few requests below. The only thing that seemed off visually was the vertical alignment of the breadcrumb:

Coordinape

The content of the CirclesPage is unstyled too, but I assume you're well aware of that and it'll be cleaned up in a future PR.

By the way, we've been switching over to Stitches for styling on new feature development. See [this comment](https://github.com/coordinape/coordinape/pull/584#issuecomment-1061269158) for more context. If you're just making minor tweaks to existing components, feel free to keep using Material-UI, but any more significant changes should be done with Stitches going forward. My experience with Stitches so far is that it's more pleasant to use; I'd love to hear your feedback on it if you try it out too.

MarcogodoyV:
@levity thanks for the review!, I did 2 new commits  that I hope gets all done, and also added the breadcrumb in the mobile view. About the CirclesPage I have that issue assign so it's going to be make all new in a new PR ( I added the page in order to have a temporaly way to move across circles)

src/routes/paths.ts
levity: @MarcogodoyV this^ is the last thing to clean up, & then it looks good to merge

levity (approved):
Looks good. Make sure to remove the unused `get*Navigation` methods in routes/paths soon in later PRs

PR #610, opened 2022-03-10
Refactor backend Zeus setup

META-DREAMER:
Simplified backend setup with better code organize and closer alignment with updated client side implementation. Queries and mutations are now explicitly separated. Backend generated code is now in `api-lib` instead of `src/lib`


scripts/db-seed.ts
```diff
@@ -368,8 +380,8 @@ function fakeCircleGifts(
     }
     const getEpoch = createSampler(availableEpochs, epochSkew);
 
-    const getSender = createSampler(users, giveSkew);
-    const getRecipient = createSampler(users, recieveSkew);
+    const getSender = createSampler(users || [], giveSkew);
+    const getRecipient = createSampler(users || [], recieveSkew);
```
CryptoGraffe, line 384: just noticed typo here in `recieveSkew` but its not new here

CryptoGraffe (approved):
lg2m just one comment that isn't new

zashton (approved):
thanks for the zeus cleanup

topocount (approved):
nice, clean separation of concerns

PR #611, opened 2022-03-10
Sentry for Hasura Serverless funcs

CryptoGraffe:
This will throw an error to sentry if our actions/events throw a 500. We can get more clever with it later. This relies on the SENTRY_DSN env var which is already set on staging/prod. It also upgrades sentry for FE cuz it was incompatible without doing that. It intentionally does not setup sentry on dev.


zashton:
@CryptoGraffe tried setting the `SENTRY_DSN` env var , doesn't seem to log.

api-lib/validate.ts
```diff
@@ -18,6 +34,7 @@ export const verifyHasuraRequestMiddleware = (handler: VercelApiHandler) => {
     try {
       await handler(req, res);
     } catch (error: any) {
+      Sentry.captureException(error);
```
topocount, line 37: We're just going to only capture unhandled exceptions that throw all the way up? I think that makes a lot of sense, since it'll have a really high signal-to-noise ratio, but I'm just a tad concerned about false negatives falling through the cracks.

topocount (approved):
Thanks for setting this up :giraffe: ! I just have one comment that could use a little assuaging but this is worth merging regardless.

api-lib/validate.ts
levity: IMO, since this is all pretty new and untested in prod, it's worth starting out noisier and dialing it down afterward.

topocount (commented):
I can't get an error showing up in our web portal even though the `captureException` method is returning an event id. Even manually querying that event id isn't showing up in prod.

zashton (approved):
Logs correctly now

api-lib/validate.ts
CryptoGraffe: I changed it to capture all error responses

api-lib/HttpError.ts
```diff
@@ -14,45 +31,53 @@ export class PayloadTooLargeError extends Error implements HttpError {
...
-): VercelResponse {
-  return res.status(statusCode).json({
-    message: error.message || 'Unexpected error',
-    // included at this level for backwards compat w/ older hasura, perhaps can remove
-    code: `${statusCode}`,
-    extensions: {
+): void {
+  Sentry.captureException(error);
+  // We have to await this flush otherwise we exit before reporting to sentry
+  Sentry.flush(awaitSentryFlushMs).then(() => {
```
topocount, line 63: I think `Sentry.close` is probably more applicable given the lifecycle of serverless function invocations, but `flush` seems like a good conservative choice given that we don't have a total understanding of the inner-workings of vercel under the hood.

topocount (approved):
I like this refactor a lot.

PR #616, opened 2022-03-14
Warning component when opting in or opting out

### LINKED ISSUE #381: Warning when opting in or opting out.
A warning is needed when a user or admin are adjusting the opt-in/opt-out status to alert them that all allocated GIVE will be removed from this user after the setting is changed. 

# Current process
## Case 1
Admin changes a user with give allocated to be opted out.
image
No warning, give removed, and status changed

## Case 2
User opts themselves out of receiving GIVE
No warning, give removed and status changed

# Recommended process
## Case 1
Admin changes a user with give allocated to be opted out.
image
Recommended warning box here: https://www.figma.com/file/eAq3jcr4epDSof4mn255SR/opt-out-warning?node-id=2%3A6
image
Click "Okay, got it" button 
GIVE Removed and status changed.

## Case 2
User opts themselves out of receiving GIVE
Warning 
image
Click "Okay, got it" button 
give removed and status changed

@AlexDistill please provide guidance here (Close if you don't think we need it)

cashd:
image

fixes #381 

Used the `ActionDialog` component to render a warning when the optedOut value is changed **and** the user has allocated GIVE.

Copy for the dialog varies depending if the user being edited is the connected account or not.

### DIFF
--- a/src/pages/AdminPage/AdminUserModal.tsx
+++ b/src/pages/AdminPage/AdminUserModal.tsx
@@ -51,11 +51,6 @@ export const AdminUserModal = ({
   const { circle: selectedCircle, circleId } = useSelectedCircle();
   const { updateUser, createUser } = useApiAdminCircle(circleId);
 
-  const [cachedOptOutStatus] = useState(
-    !!user?.fixed_non_receiver ||
-      !!user?.non_receiver ||
-      !selectedCircle.default_opt_in
-  );
   const [showOptOutChangeWarning, setShowOptOutChangeWarning] = useState(false);
   const [hasAcceptedOptOutWarning, setHasAcceptedOptOutWarning] =
     useState(false);
@@ -64,6 +59,10 @@ export const AdminUserModal = ({
     setHasAcceptedOptOutWarning(false);
   }, [user?.address]);
 
+  const isOptedOut =
+    !!user?.fixed_non_receiver ||
+    !!user?.non_receiver ||
+    !selectedCircle.default_opt_in;
   const hasGiveAllocated = !!user?.give_token_received;
   const userIsAccount =
     account?.toLocaleLowerCase() === user?.address.toLocaleLowerCase();
@@ -83,8 +82,8 @@ export const AdminUserModal = ({
         hideFieldErrors
         submit={params => {
           const showWarning =
-            cachedOptOutStatus !== params.non_receiver && hasGiveAllocated;
-          if (showWarning && !hasAcceptedOptOutWarning) {
+            isOptedOut && hasGiveAllocated && !hasAcceptedOptOutWarning;
+          if (showWarning) {
             setShowOptOutChangeWarning(true);
           } else {
             setShowOptOutChangeWarning(false);
--- a/src/pages/AdminPage/AdminUserModal.tsx
+++ b/src/pages/AdminPage/AdminUserModal.tsx
@@ -1,5 +1,3 @@
-/* eslint-disable no-console */
-
 import { useEffect, useMemo, useState } from 'react';
 
 import { useWeb3React } from '@web3-react/core';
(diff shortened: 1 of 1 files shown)


reeserj:
@cashd heads up.

cashd:
Thanks for the testing @reeserj! I just fixed my bug. Along the way I found an existing bug that if `default-opt-in` was **disabled** for the circle then the `opt-out` status for every account would show up as **enabled** on the user forum. Fixed this also.

Before:
before

After:
after

reeserj:
I'm still not seeing the warning messages.

Does the vercel bot need to send a new link? @cashd

levity:
Just kicked off another deployment; pls check back in 30m or so (we have a few other deployments queued up ahead of it)

src/pages/AdminPage/AdminUserModal.tsx
```diff
@@ -43,11 +45,25 @@ export const AdminUserModal = ({
...
   const classes = useStyles();
 
   const { circle: selectedCircle, circleId } = useSelectedCircle();
   const { updateUser, createUser } = useApiAdminCircle(circleId);
 
+  const [showOptOutChangeWarning, setShowOptOutChangeWarning] = useState(false);
+  const [hasAcceptedOptOutWarning, setHasAcceptedOptOutWarning] =
+    useState(false);
+
+  useEffect(() => {
```
levity, line 58: If you apply [this change](https://github.com/coordinape/coordinape/commit/ffe6f35cf1e0ed42ec92fd5431abd9c981736879) to your PR then this `useEffect` becomes unnecessary. I'd do it myself but I can't push changes back to your fork

levity (changes requested):

levity:
The dialog is appearing for me, btw, so nice work there 👍

src/forms/AdminUserForm.tsx
```diff
@@ -33,10 +33,7 @@ const AdminUserForm = createForm({
     address: v.user?.address ?? '',
     non_giver: v.user?.non_giver ?? false,
     fixed_non_receiver: !!v.user?.fixed_non_receiver ?? false,
-    non_receiver:
-      !!v.user?.fixed_non_receiver ||
-      !!v.user?.non_receiver ||
-      !v.circle.default_opt_in,
+    non_receiver: !!v.user?.fixed_non_receiver || !!v.user?.non_receiver,
```
crabsinger, line 36: Ok, I've been thinking through this line and I agree that this change makes sense.

If circle.default_opt_in is false, but the user has opted-in, it should show them as "opt-in", and this makes it show as opted-out, which is a bug.

great catch!

reeserj:
This is working when I opt the user out but not when I use the "Force Opted Out" button

cashd:
Nice catches @levity, just addressed them!

src/forms/AdminUserForm.tsx
cashd: > This is working when I opt the user out but not when I use the "Force Opted Out" button

forceOptouttest

This is the flow I was testing and works for me. Let me know if you did something different :)

cashd:
> This is working when I opt the user out but not when I use the "Force Opted Out" button

forceOptouttest

This is the flow I was testing and it works. Let me know if you did something different and I can find out why :)

levity:
@crabsinger @cashd please check out [this comment](https://github.com/coordinape/coordinape/issues/381#issuecomment-1068407838) on the parent issue #381. This PR is good for what it is, but it only addresses half of the parent issue. If you want to do those other changes in a new PR, that's fine with me, just make sure that the parent issue isn't closed until everything is done.

cashd:
@levity @crabsinger Just saw the comment. I will add this dialog to the allocation flow also & circle back!

crabsinger:
I'm a bit confused, on this preview branch I **do** see the dialog pop when I have GIVE allocated to me, and I try to opt-myself out.

But don't see any code diff that explains this...

levity:
> I'm a bit confused, on this preview branch I **do** see the dialog pop when I have GIVE allocated to me, and I try to opt-myself out.
> 
> But don't see any code diff that explains this...

Oh, you're right, it already exists 😅 it was done in #363.
So we can just merge this and close the parent issue.

levity (approved):

crabsinger:
Sweet! Nice work @cashd - thanks for your first contribution!

PR #624, opened 2022-03-15
Create codeql-analysis.yml

crabsinger:
Enable https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning


crabsinger:
@levity might be more noise than signal, but thought we could try turnign this on?

levity (approved):
Let's try it :)

PR #626, opened 2022-03-15
Circle ID generator

### LINKED ISSUE #486: Write Shared Code: Circle ID Generator
### Specification

Since only the vaults will be on-chain at first, the `circle` arguments in the contracts will not be addresses, but arbitrary `bytes32` values. So we need to define a convention that allows us to map the IDs of circles in our database to `bytes32` values & vice versa. These are used when setting allowances, uploading epoch roots, & claiming.

Suggestion: use circle ID from database, prefixed with some version tag for future-proofing, separated by some special character. e.g. “v1-17”. Then convert this string to a `bytes32` using e.g. the ethers.js [Bytes32String](https://docs.ethers.io/v5/api/utils/strings/#Bytes32String) utility functions.

levity:
Closes #486 

Also lightly exploring how to organize code related to vault logic, as in, stuff that isn't directly related to UX or the details of interacting with the chain. This could reasonably go somewhere in `src/utils` instead, but I think a more "feature-oriented organization" will be easier to grok & also easier to factor into a separate module if/when that makes sense to do.


teeolendo (approved):
LGTM

PR #627, opened 2022-03-16
Save Vault Distributions from an Epoch (fixed from #604)

META-DREAMER:
Rebased #604 on hasura-next and updated code to follow new patterns + redid some migrations

paired with @teeolendo 

Original Description:
---
Key changes in this PR are:

Added vaults, vaults_transactions, distributions, claims
The mutations for these tables aren't written inside of an action. There was no reasonable use-case that I could come up with that would justify the use of action at this time. Happy to review this based on feedback.
Added a mutation hook called useSaveDistribution which is contained in the useGetAllocations. Not the best location but I'm happy to change it.
The useSaveDistribution hook also saves claims that are passed from the createDistribution hook which generates the merkle tree for an epoch
This PR closes https://github.com/coordinape/coordinape/issues/507 and https://github.com/coordinape/coordinape/issues/480.


src/hooks/gql/useVaults.ts
```diff
@@ -0,0 +1,31 @@
+import { client } from 'lib/gql/client';
+import { useQuery } from 'react-query';
+
+export function useVaults(orgId: number) {
```
levity, line 4: so... one of the things that `useVaults` in `src/recoilState/vaults` does is look up the actual token symbol for a simple token and set `vault.symbol` to that value. for parity, we either have to do that here as well, or look up the token symbol during creation and save it instead of "OTHER".

src/pages/DistributePage/DistributePage.tsx
```diff
@@ -38,15 +47,32 @@ function DistributePage() {
...
+
   const { uploadEpochRoot } = useDistributor();
-  const [selectedVault, setSelectedVault] = useState<IVault | undefined>();
+  const [selectedVault, setSelectedVault] = useState<
+    GraphQLTypes['vaults'] | undefined
+  >();
   const { apeError } = useApeSnackbar();
+  const [distributionDTO, setDistributionDTO] =
+    useState<ValueTypes['distributions_insert_input']>();
+  const { mutateAsync } = useSaveEpochDistribution(distributionDTO);
```
levity, line 64: now that we're using `useMutation` from `react-query`, we can avoid this weirdness and just pass `distribution` into the mutateAsync as an ordinary argument, right?

src/pages/VaultsPage/DepositModal.tsx
```diff
@@ -52,8 +51,10 @@ export default function DepositModal({
 
     (async () => {
       const tokenAddress =
-        vault.type === 'OTHER' ? vault.simpleTokenAddress : vault.tokenAddress;
-      const token = selectedContracts.getERC20(tokenAddress);
+        vault.symbol === 'OTHER'
```
levity, line 54: in the previous implementation, type & symbol were separate fields... if we're combining them (see my comment about useVaults), this check might have to become e.g. `Object.values(Asset).includes(vault.symbol)`

src/pages/VaultsPage/VaultsPage.tsx
```diff
@@ -13,7 +15,7 @@ const VaultsPage = () => {
   const closeModal = () => setModal('');
 
   const currentOrg = useCurrentOrg();
-  const vaults = useVaults(currentOrg.data?.id);
+  const { isLoading, data: vaults } = useVaults(Number(currentOrg.data?.id));
```
levity, line 18: This data refreshed correctly for me after creating the first vault, but not the second one

src/pages/DistributePage/DistributePage.tsx
```diff
@@ -101,8 +127,34 @@ function DistributePage() {
...
+          ),
+          epoch_id: Number(epochId),
+          merkle_root: distribution.merkleRoot,
+          claims: {
+            data: claims,
+          },
+          vault_id: Number(selectedVault.id),
+        };
+        setDistributionDTO(updateDistribution);
+        await mutateAsync();
```
levity, line 156: i feel nervous about this gap in time between the transaction being submitted & the distribution info being saved to the db... i kind of want to suggest that we save the info to the db first, marked as pending, then submit the tx, & mark as committed afterward. so that the worst case scenario is orphaned data in the db, not an orphaned merkle tree

src/pages/DistributePage/DistributePage.tsx
```diff
@@ -38,15 +47,32 @@ function DistributePage() {
...
   const contracts = useContracts();
   const currentOrg = useCurrentOrg();
-  const vaults = useVaults(currentOrg.data?.id);
+  const { isLoading: vaultLoading, data: vaults } = useVaults(
+    currentOrg.data?.id as number
+  );
+
   const { uploadEpochRoot } = useDistributor();
-  const [selectedVault, setSelectedVault] = useState<IVault | undefined>();
+  const [selectedVault, setSelectedVault] = useState<
```
levity, line 58: this is a bit redundant with selectedVaultId... maybe just do this instead?
```js
const selectedVault = vaults?.find(v => v.id === selectedVaultId)
```

levity (changes requested):

hasura/migrations/default/1647395497748_update_claims_table/up.sql
```diff
@@ -0,0 +1,46 @@
+
+alter table "public"."claims" drop column "flag" cascade;
+
+alter table "public"."claims" add column "claimed" boolean
```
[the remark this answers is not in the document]
teeolendo: I may not get what you mean but the claims table already has timestamp `created_at` and `updated_at` fields. Ideally, when the `updated_at` gets updated when a claim is made. Does this suffice or are you getting at something else?

src/lib/gql/queries.ts
```diff
@@ -19,3 +19,21 @@ export const getCurrentEpoch = async (
   });
   return currentEpoch;
 };
+
+export async function getVaults(orgId: number) {
```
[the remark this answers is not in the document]
teeolendo: Removed.

src/hooks/gql/useVaults.ts
teeolendo: I've changed introduced a symbol that's populated by the type or erc20 symbol

src/pages/DistributePage/DistributePage.tsx
teeolendo: Done.

src/hooks/gql/useVaults.ts
teeolendo: Done.

src/pages/VaultsPage/DepositModal.tsx
teeolendo: Done as well.

src/pages/DistributePage/DistributePage.tsx
teeolendo: I was uneasy about this as well. I implemented your suggestion and I think I'll sleep better tonight. Gonna create the migration and the update mutation.

src/lib/gql/queries.ts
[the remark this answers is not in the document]
teeolendo: Removed

src/utils/userHelpers.ts
```diff
@@ -0,0 +1,3 @@
+export const isUserAdmin = (user: { role: number }) => {
```
levity, line 1: it might be worth considering putting this somewhere in src/lib, e.g. a src/lib/users.ts file, as a start to gathering the "model logic" in one place rather than scattering it across the frontend app.

levity: i'm going to do this now on the org-view branch, actually -- feel free to leave this as-is and we can clean up later

src/pages/DistributePage/DistributePage.tsx
teeolendo: Added a boolean column to the distributions tables called `saved_on_chain` which is updated after the distribution is saved on chain. Updated table permissions to allow user updates to `distributions` and `claims` table.

src/utils/userHelpers.ts
teeolendo: Done.

src/pages/VaultsPage/VaultsPage.tsx
teeolendo: I'm still review how best to do this. What happens is that the mutations is called directed on the backend when a new vault is created without using react-query. So I have to invalidate the cached query to trigger a re-render. Almost there.

levity: Works for me!

FWIW, seems like QueryClient.invalidateQueries is relatively [straightforward](https://react-query.tanstack.com/reference/QueryClient#queryclientinvalidatequeries) too, so we can look into it later on. I can imagine that in other cases it'd be good to just signal the need and let the refetching get triggered at the right time by the component that owns the query.

src/hooks/useVaultFactory.test.tsx
```diff
@@ -47,15 +63,15 @@ test('create a vault', async () => {
     () => {
       expect(vault).toBeTruthy();
       expect(vault.id).toMatch(/0x[a-fA-F0-9]{40}/);
```
levity, line 65:
```suggestion
      expect(vault.vault_address).toMatch(/0x[a-fA-F0-9]{40}/);
```

src/hooks/useVaultFactory.test.tsx
```diff
@@ -81,7 +97,7 @@ test('create a vault with a custom asset', async () => {
...
```
levity, line 99:
```suggestion
      expect(vault.vault_address).toMatch(/0x[a-fA-F0-9]{40}/);
```

levity (approved):

levity (approved):

levity:
Ah, that’s a fair point. A bit more indirect but perfectly workable.

PR #649, opened 2022-03-17
Replace all usages of Prisma with Hasura adminClient

### LINKED ISSUE #605: Remove Prisma from codebase
The only place we are using Prisma is the auth handler here: `https://github.com/coordinape/coordinape/blob/main/api/hasura/auth.ts`

We can replace this with using the Zeus admin client instead to talk to the database (like all the other api handlers).

Will allow us to remove all the prisma code generation logic as well in our bootstrapping steps.

META-DREAMER:
Simplifies our setup a bunch and consolidates everything to flow through Hasura

closes #605


hasura/metadata/databases/default/tables/tables.yaml
```diff
@@ -6,6 +6,7 @@
 - "!include public_claims.yaml"
 - "!include public_distributions.yaml"
 - "!include public_epoches.yaml"
+- "!include public_feedbacks.yaml"
```
zashton, line 9: we could untrack this table since its not used in the app at all now

zashton (approved):
tested the reseed and it no longer hangs 👍

hasura/metadata/databases/default/tables/public_feedbacks.yaml
```diff
@@ -0,0 +1,3 @@
+table:
```
crabsinger, line 1: we can remove this too, right?

levity (approved):

hasura/metadata/databases/default/tables/tables.yaml
META-DREAMER: Its being used in the scripts though, thats why I needed to add it (so that adminClient can use it)

hasura/metadata/databases/default/tables/public_feedbacks.yaml
META-DREAMER: No, it was being used by the Prisma code in the scripts, so we need to track it in Hasura so we can interact with it via graphql instead

PR #650, opened 2022-03-17
Update DB seeding scripts to disable Hasura Event Triggers

META-DREAMER:
This fixes the issue where Hasura starts running thousands of event triggers when seeding the DB, causing peoples computers to grind to a halt for 15+ minutes.

This was accomplished by commenting out the event triggers in Hasura metadata and applying it before running the seeding scripts (and re-enabling them afterwards)


zashton (approved):
tested the reseed and it no longer hangs 👍

crabsinger:
nice simple solution! sed for the win @META-DREAMER!

Thanks for fixing this.

PR #652, opened 2022-03-17
New history page

### LINKED ISSUE #556: Org View: Update Circle History page
**Specification**

- Implement the new design for the circle history page seen [here in Figma](https://www.figma.com/file/wXF7xGVv1j2SqwWfgbamS8/App-1.0?node-id=3987%3A15460)
- See if it's possible to reuse the data-fetching that's already in place for the legacy history page
  - If not, pls comment; we'll probably create another task for redoing the data-fetching with GraphQL
- No need to add the claim button in the upper right; that will be covered by #470 & #481

levity:
Applies to #556 but doesn't close it (finishing touches still needed)


levity:
hey @MarcogodoyV, I'm going to leave some feedback here instead of on #666.

First off, here's an image with some visual changes needed:

Screen_Shot_2022-03-22_at_10_58_16_AM

src/pages/HistoryPage/HistoryPage.tsx
```diff
@@ -1,260 +1,451 @@
...
+                      <Text css={{ mb: '$xs', lineHeight: 'normal' }}>
+                        {gift.note}
+                      </Text>
+                    )}
+                    <Box css={{ fontSize: '$3', color: '$green' }}>
+                      {gift.tokens} {tokenName} received from {gift.sender.name}
+                    </Box>
+                  </Box>
+                </Box>
+              ))
```
levity, line 295: @MarcogodoyV Please extract this into its own component so it can be reused for sent gifts/notes as well. Change the text to `{gift.tokens} {tokenName} sent to {gift.sender.name}` in that case.

To test this, it's probably easier to put in some dummy data for `received` and `sent` in this component rather than having to go through the whole allocation process.

PR #653, opened 2022-03-17
Update profile/skills tags

### LINKED ISSUE #644: Specific Profile Tags for Yearn
Yearn needs four specific Profile Tags, which they are going to ask their contributors to select from next epoch.  These would form the basis of experimenting with weighted GIVE values for various "hubs" within the DAO.  

Tags will need to be available in **6 days** for Yearn to use them for their next epoch.

Tags should be:

*People & Governance
*Development & Engineering
*Growth & Marketing
*DAO General

Please **do include the asterisks**, as this will allow the tags to be easily found by Yearn (we use an alphabetical list I believe) for their experiment this month.

spydermonkee:
This adds a few tags that we want to add to the app for Yearn's sake. We worked to make them generic enough to be useful to others and not be weird or out of place. Resolves #644

### DIFF
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -68,4 +68,8 @@ export const SKILLS = [
   'Risk',
   'Tokenomics',
   'Contributor Experience',
+  'People & Governance',
+  'Development & Engineering',
+  'Growth & Marketing',
+  'DAO Teams',
 ];
--- a/src/config/constants.ts
+++ b/src/config/constants.ts
@@ -68,8 +68,8 @@ export const SKILLS = [
   'Risk',
   'Tokenomics',
   'Contributor Experience',
-  'People & Governance',
-  'Development & Engineering',
-  'Growth & Marketing',
-  'DAO Teams',
+  '*People & Governance',
+  '*Development & Engineering',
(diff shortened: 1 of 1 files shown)


levity (dismissed):

src/config/constants.ts
```diff
@@ -68,4 +68,8 @@ export const SKILLS = [
   'Risk',
   'Tokenomics',
   'Contributor Experience',
+  '*People & Governance',
```
crabsinger, line 71: We want the asterisk at the start of these? Seems different than the other tags? @spydermonkee

spydermonkee: Yep, @crabsinger - The request specifically included that so that these tags would be easy to stand out. I feel mixed about it, but am more focused on getting this done to empower the Yearn team. If you would like, I could check with the Yearn team and I think the fact that I added them at the end means that they will be easy to find by virtue of being all at the end of the list might be acceptable for "discoverability". Want me to check on that?

crabsinger (approved):

crabsinger (approved):
Rebased main onto this and force pushed to fix the build issues. Looks good without the asterisks.

PR #655, opened 2022-03-19
Draft: beginning of deleteUser mutation

CryptoGraffe:
Todo on this:

- [ ] actually implement the mutation, can it just be the native delete?
- [ ] connect the userDeleted event_trigger
- [ ] consider the permissions
- [ ] look at the other uses of resetUser in the php code and try to unify the impl here
- [ ] consider how much of this should be event triggers
- [ ] test plan


topocount:
closing since #701  was merged

PR #656, opened 2022-03-20
Issue/#561

j-yw:
added the tooltip info to the allocate give page

### DIFF
--- a/src/components/ProfileCard/ProfileCard.tsx
+++ b/src/components/ProfileCard/ProfileCard.tsx
@@ -70,6 +70,9 @@ const useStyles = makeStyles(theme => ({
     fontSize: 24,
     fontWeight: 600,
     color: theme.colors.text,
+    display: 'flex',
+    alignItems: 'center',
+    justifyContent: 'center',
   },
   tooltipLink: {
     display: 'block',
@@ -77,9 +80,15 @@ const useStyles = makeStyles(theme => ({
     textAlign: 'center',
     color: theme.colors.linkBlue,
   },
+  tooltipButton: {
+    marginTop: '16px',
+  },
   tooltip: {
     fontWeight: 400,
     color: theme.colors.text,
+    padding: '28px',
+    minWidth: '296px',
+    minHeight: '263px',
   },
   skillContainer: {
     gridColumn: '1 / 4',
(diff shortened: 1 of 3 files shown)


crabsinger:
@j-yw seems you're combining two fixes into one PR. 

first is the tooltip addition
second is adding `stream` as a package, can you separate these apart so we can discuss and move ahead with them individually?

j-yw:
> @j-yw seems you're combining two fixes into one PR.
> 
> first is the tooltip addition second is adding `stream` as a package, can you separate these apart so we can discuss and move ahead with them individually?

my apologies that this was overlooked. It's fixed now.

j-yw:
> https://capture.dropbox.com/3J4kbcNrqAlJlSLZ

I think making the button fullwidth address the button centering issue better as we don't need to change the tooltip display property on the tooltip component itself.

crabsinger:
@j-yw please add a screenshot or video of the change to PR description.

j-yw:
> @j-yw please add a screenshot or video of the change to PR description.

done!

crabsinger:
@j-yw title still doesn't seem bold.

j-yw:
> 
Screen Shot 2022-04-01 at 2 41 37 PM

It is bold on my system, I'm using a <b> tag. Could it be your system not rendering correctly, I noticed a slight color difference between the title and the text from your screen shots.

j-yw:
I reworked it and it's more straight forward now after the latest commit.

added a screenshot to the PR desc. as well

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -176,27 +190,52 @@ const ProfileCardInner = ({
...
-                Let us know what you think
-              </a>
-            </ApeInfoTooltip>
+            <span
+              style={{
+                marginTop: '8px',
+              }}
+            >
+              <ApeInfoTooltip classes={{ tooltip: classes.tooltip }}>
+                {user.role !== 2 ? <></> : <></>}
```
levity, line 202: remove

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -176,27 +190,52 @@ const ProfileCardInner = ({
...
+                </p>
+                <a
+                  href={EXTERNAL_URL_FEEDBACK}
+                  rel="noreferrer"
+                  target="_blank"
+                  className={classes.tooltipLink}
+                >
+                  Let us know what you think
+                </a>
+                <div className={classes.tooltipButton}>
```
levity, line 219: three things about this:

1. i think this div is unnecessary? there's already one two lines below.
2. the test on the next line is incorrect. you're supposed to be checking that the viewing user is an admin; this checks that the user whose card you're looking at is the Coordinape user.
3. when you want to hide or show a component based on a boolean value, the idiomatic way to do it is `test && <Component>`, not `test ? <Component> : ''`.

levity (changes requested):
See below. Also, the padding on the tooltip now doesn't match the screenshot or Figma

PR #660, opened 2022-03-21
create epoch

### LINKED ISSUE #567: Migrate Create Epoch v2/{circle_id}/admin/epoches
## Summary:
Migrate Create Epoch Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `<POST v2/{circle_id}/admin/epoches>` to Hasura.

**Middleware:** Hasura event secret + circle admin

**Additional Validations besides basic inputs:** (`repeat == 1 weekly` , `repeat == 2 monthly`)
Check if its more than 7 days length for a weekly repeating epoch. Check if its more than 28 days length for a monthly repeating epoch. Check if there is more than 1 repeating epoch for this circle if creating a repeating epoch. Check if there is any overlap of start/end datetimes with current epochs. Check whether epoch `ended` == 1.

**Logic:**
Input `start_date`(this is a UTC datetime), `grant`, `days`, `repeat`

Add days to `start_date` to compute `end_date` (`start_date` & `end_date` timing is the same)

**DB Tables:**
epochs

#### Notes:

zashton:
Closes #567
### Test Plan

1. Create an epoch via the hasura console using this mutation
```gql
mutation MyMutation {
  createEpoch(payload: {circle_id: <circle_id>, days: 7, repeat: 1, 
    start_date: "2021-08-27T10:00:00"}) {
    epoch{
      id
      start_date
      end_date
      repeat
    }
  }
}
```
2. Epoch should appear in your circle admin page http://localhost:3000/admin/circles
3. try creating a date overlapping epoch or more than 1 active epochs that are repeating > 0 it should throw an validation error

### DIFF
--- a/api/hasura/actions/createEpoch.ts
+++ b/api/hasura/actions/createEpoch.ts
@@ -39,16 +39,16 @@ async function handler(request: VercelRequest, response: VercelResponse) {
         ],
       });
       if (epochs.length) {
-        const overlappingEpoch = epochs.pop();
+        const repeatingEpoch = epochs.pop();
         errorResponseWithStatusCode(
           response,
           {
             message:
               `You cannot have more than one repeating active epoch. ` +
-              `Dates overlap with epoch id #${overlappingEpoch?.id} ` +
+              `Dates overlap with epoch id #${repeatingEpoch?.id} ` +
               `that occurs between ` +
-              `${formatShortDateTime(overlappingEpoch?.start_date)} and ` +
-              `${formatShortDateTime(overlappingEpoch?.end_date)}`,
+              `${formatShortDateTime(repeatingEpoch?.start_date)} and ` +
+              `${formatShortDateTime(repeatingEpoch?.end_date)}`,
           },
           422
         );
@@ -78,14 +78,14 @@ async function handler(request: VercelRequest, response: VercelResponse) {
       ],
     });
     if (epochs.length) {
-      const epoch = epochs.pop();
+      const overlappingEpoch = epochs.pop();
       errorResponseWithStatusCode(
         response,
         {
           message:
             `This epoch overlaps with an existing epoch that occurs between ` +
-            `${formatShortDateTime(epoch?.start_date)} and ` +
-            `${formatShortDateTime(epoch?.end_date)}. ` +
+            `${formatShortDateTime(overlappingEpoch?.start_date)} and ` +
+            `${formatShortDateTime(overlappingEpoch?.end_date)}. ` +
             `Please adjust epoch settings to avoid overlapping with existing epochs`,
         },
         422
--- a/api/hasura/actions/createEpoch.ts
+++ b/api/hasura/actions/createEpoch.ts
@@ -1,5 +1,4 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node';
-import { z } from 'zod';
 
 import { authCircleAdminMiddleware } from '../../../api-lib/circleAdmin';
 import { EPOCH_REPEAT } from '../../../api-lib/constants';
@@ -8,7 +7,6 @@ import { adminClient } from '../../../api-lib/gql/adminClient';
 import {
   errorResponse,
   errorResponseWithStatusCode,
-  zodParserErrorResponse,
 } from '../../../api-lib/HttpError';
 import {
   createEpochInput,
@@ -112,10 +110,6 @@ async function handler(request: VercelRequest, response: VercelResponse) {
 
     response.status(200).json(insert_epochs_one);
   } catch (err) {
-    if (err instanceof z.ZodError) {
-      zodParserErrorResponse(response, err.issues);
-      return;
-    }
     errorResponse(response, err);
     return;
   }
--- a/api/hasura/actions/createEpoch.ts
+++ b/api/hasura/actions/createEpoch.ts
@@ -1,9 +1,9 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node';
-import { DateTime } from 'luxon';
 import { z } from 'zod';
 
 import { authCircleAdminMiddleware } from '../../../api-lib/circleAdmin';
 import { EPOCH_REPEAT } from '../../../api-lib/constants';
+import { formatShortDateTime } from '../../../api-lib/dateTimeHelpers';
 import { adminClient } from '../../../api-lib/gql/adminClient';
 import {
   errorResponse,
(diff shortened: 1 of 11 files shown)


topocount:
@zashton I'm going to rebase the epoch cron job work on this and will review it as I borrow from it.

src/lib/zod/index.ts
```diff
@@ -94,6 +97,36 @@ export const deleteEpochInput = z
...
+  .strict()
+  .superRefine((val, ctx) => {
+    let message;
+    if (val.days > 7 && val.repeat === 1) {
+      message =
+        'You cannot have more than 7 days length for a weekly repeating epoch.';
+    } else if (val.days > 28 && val.repeat === 2) {
+      message =
+        'You cannot have more than 28 days length for a monthly repeating epoch.';
+    }
```
topocount, line 120: Is this a departure from the incumbent logic in laravel? I thought we could have epoch lengths that were greater than 28 days but would be truncated down to a shorter month's length if necessary.

api/hasura/actions/createEpoch.ts
```diff
@@ -0,0 +1,118 @@
...
+          },
+          {
+            id: true,
+          },
+        ],
+      });
+      if (epochs.length) {
+        errorResponseWithStatusCode(
+          response,
+          { message: 'You cannot have more than one repeating active epoch.' },
```
topocount, line 44: Consider adding the dates of the existing epoch too.
```suggestion
      if (epochs.length) {
      const overlappingEpoch = epochs.pop()
        errorResponseWithStatusCode(
          response,
          { message: `You cannot have more than one repeating active epoch. Dates overlap with epoch id #${overlappingEpoch.id}` },
```

topocount (changes requested):
This is looking pretty good. I'll call this from within the endEpoch cron task.

A couple requested changes, however.

src/lib/zod/index.ts
zashton: i copied the validation over from what i did in the past
https://github.com/coordinape/coordinape-backend/blob/main/app/Http/Requests/NewEpochRequest.php#L41
 
there is also a logic that trims down the start day for repeating epochs
e.g for monthly if start date is 30th when it comes to repeating in Feburary the start date will be trimmed to 28th

https://github.com/coordinape/coordinape-backend/blob/main/app/Repositories/EpochRepository.php#L90

api/hasura/actions/createEpoch.ts
zashton: updated with start/end datetime

api/hasura/actions/createEpoch.ts
```diff
@@ -0,0 +1,118 @@
...
+          },
+          {
+            id: true,
+            start_date: true,
+            end_date: true,
+          },
+        ],
+      });
+      if (epochs.length) {
+        const overlappingEpoch = epochs.pop();
```
META-DREAMER, line 42: This doesn't seem to check if it's actually overlapping, just if there's another repeating epoch set. Should this be named "repeatingEpoch" instead?

api/hasura/actions/createEpoch.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const end_date = start_date.plus({ days: days });
+    const { epochs } = await adminClient.query({
+      epochs: [
+        {
+          limit: 1,
+          where: {
+            start_date: { _lte: end_date },
+            circle_id: { _eq: circle_id },
+            end_date: { _gt: start_date },
```
META-DREAMER, line 70: I think this filter should have an OR condition on start and end dates, right now it would only error if both are true IIRC

api/hasura/actions/createEpoch.ts
```diff
@@ -0,0 +1,118 @@
...
+        },
+        {
+          id: true,
+          start_date: true,
+          end_date: true,
+        },
+      ],
+    });
+    if (epochs.length) {
+      const epoch = epochs.pop();
```
META-DREAMER, line 81: This can be called overlappingEpoch

api/hasura/actions/createEpoch.ts
zashton: updated this

api/hasura/actions/createEpoch.ts
zashton: updated this as well

api/hasura/actions/createEpoch.ts
zashton: pretty sure to overlap the 2 conditions must be true.
e.g end_date > the other epoch start_date doesn't mean it intersects
@META-DREAMER

topocount: I agree, both rules should be enforced in the filter.

src/lib/zod/index.ts
topocount: > there is also a logic that trims down the start day for repeating epochs
e.g for monthly if start date is 30th when it comes to repeating in Feburary the start date will be trimmed to 28th

I see. Is this legacy functionality then for existing circles? It'd be nice to remove it, or make it consistent with the current way we create epochs. We should probably just wait to create the new schema to sort this out. Thanks for the explanation!

topocount (approved):

src/lib/zod/index.ts
zashton: > > there is also a logic that trims down the start day for repeating epochs
> > e.g for monthly if start date is 30th when it comes to repeating in Feburary the start date will be trimmed to 28th
> 
> I see. Is this legacy functionality then for existing circles? It'd be nice to remove it, or make it consistent with the current way we create epochs. We should probably just wait to create the new schema to sort this out. Thanks for the explanation!

How would you truncate the end dates to make it not overlap with the subsequent repeating epochs ? Do you mean creating the repeating epochs in advance ?

api/hasura/actions/createEpoch.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const end_date = start_date.plus({ days: days });
+    const { epochs } = await adminClient.query({
+      epochs: [
+        {
+          limit: 1,
+          where: {
+            start_date: { _lte: end_date },
+            circle_id: { _eq: circle_id },
+            end_date: { _gt: start_date },
```
topocount, line 70: I think we're all wrong here. In order to check for any overlaps at all, something like this is required:
```js
        where: {
          _or: [
            {
              start_date: { _lt: nextEndDate },
              end_date: { _gt: nextEndDate },
            },
            {
              start_date: { _lt: nextStartDate },
              end_date: { _gt: nextStartDate },
            },
          ],
        },

```

PR #617, opened 2022-03-14
ADD Give Feedback button to the profile dropdown menu

amrro:
Closes #603

PR #606, opened 2022-03-09
Add Discord Support to Creat Circle Form

amrro:
fixes #601 

This adds the Coordinape Discord support to the Create Circle and follow up the new design described in the #601.

PR #661, opened 2022-03-21
Fix misalignment in Form after adding Discord support & blue underline bug under links

amrro:
### Summary
While I was working on #606, I didn't make an accurate review for form's fields alignment, the reason for that is using `</a>` tag around `<Button>` component. 

Using `</a>` tag around `<Button>` introduced two visual bugs: 
- **blue underline** for links in Discord Support Button in **Create Circle Page** & **Give Feedback** in the menu. 
- **Misalignment** among some fields in **Create Circle Page** Form.

I found a better solution that solves both issues at once by using the `href` that comes included with the [`<Button>` API](https://mui.com/api/button/). This allows MUI to decide how to include the `</a>` tag, in addition, this is more concise and encapsulated than the old way.

### Details
This PR solves two issues at once.
#### Blue underline
Blue underline is showing in couple of places: Discord button in create
new circle and in Give Feedback link in the menu.

Worth noting that the blue underline bug **only appears in Google Chrome** 
(inherently maybe all chrome-based browsers) but this underline doesn't show 
at all in Mozilla Firefox.

This issue was reported in this [comment](https://github.com/coordinape/coordinape/pull/617#issuecomment-1071899863), but also noticed in the Discord button. 

#### Misalignment
Finally, `</a>` component wasn't playing nicely with the form which caused Discord support button to not be perfectly aligned. 
Here are some examples: 
**Large Screen**
large screen screenshot
**Small Screen**
small screen screenshot

Actually, the behavior varies based on the breakpoints.

### Other fixes: 
- Center both texts in sub-label (text under Need More Help?) for matching and better experience. 
- Make **Join Discord** font size smaller.

### DIFF
--- a/src/components/MyAvatarMenu/MyAvatarMenu.tsx
+++ b/src/components/MyAvatarMenu/MyAvatarMenu.tsx
@@ -105,7 +105,6 @@ const useStyles = makeStyles(theme => ({
   },
   feedbackButton: {
     backgroundColor: '#2a849b',
-    width: '100%',
     margin: theme.spacing(1.5, 0, 0),
     padding: theme.spacing(1.5, 0, 1.5, 5),
     borderRadius: 0,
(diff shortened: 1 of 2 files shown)


teeolendo (approved):
LGTM. Thanks for catching these.

amrro:
@teeolendo, thanks for the fast review.
 
I worked on a better solution that fixes other bugs too.

levity (approved):
Looks great! Thanks for being thorough.

PR #662, opened 2022-03-21
investigating the hasura-preview CI workflow

wawhal:
Looking into the Hasura Cloud CI workflow. Please feel free to ignore this PR

### DIFF
--- a/.github/workflows/hasura-preview.yml
+++ b/.github/workflows/hasura-preview.yml
@@ -89,23 +89,3 @@ jobs:
             GraphQL Endpoint available at ${{steps.hasura_cloud_preview.outputs.graphQLEndpoint}}
             ${{ steps.hasura-change.outputs.change_html }}
 
-  remove-cloud-preview-on-failure:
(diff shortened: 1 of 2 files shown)


wawhal:
@crabsinger another approval on the deployment please :pray: 
The first one failed due to [an unrelated error](https://github.com/coordinape/coordinape/runs/5633827503?check_suite_focus=true) with adding an env to Vercel.

crabsinger:
@wawhal done.

wawhal:
Looks like the `VERCEL_TOKEN` secret is empty in the workflow. That's causing the setting and removing vercel env step to fail.

crabsinger:
> Looks like the `VERCEL_TOKEN` secret is empty in the workflow. That's causing the setting and removing vercel env step to fail.

@wawhal yeah, we don't allow secrets to be passed to Github Actions that run on forks. I've copied this branch to not be a fork here: https://github.com/coordinape/coordinape/pull/664 and it built successfully.

PR #664, opened 2022-03-21
Fix hasura preview ci workflow

crabsinger:
Internal version of https://github.com/coordinape/coordinape/pull/662 for testing Hasura Previews

### DIFF
--- a/.github/workflows/hasura-preview.yml
+++ b/.github/workflows/hasura-preview.yml
@@ -89,23 +89,3 @@ jobs:
             GraphQL Endpoint available at ${{steps.hasura_cloud_preview.outputs.graphQLEndpoint}}
             ${{ steps.hasura-change.outputs.change_html }}
 
-  remove-cloud-preview-on-failure:
-    runs-on: ubuntu-latest
-    needs: [ hasura-cloud-preview ]
-    if: always() && needs.hasura-cloud-preview.result == 'failure'
-    steps:
-      - name: Delete failed preview app
-        continue-on-error: true
-        uses: hasura/hasura-cloud-preview-apps@v0.1.5
-        with:
-          name: "coordinape-${{github.env.GITHUB_HEAD_REF}}${{github.event.number}}"
-          delete: true
-          postgresDBConfig: |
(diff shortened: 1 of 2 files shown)


wawhal:
@crabsinger wouldn't this work for you? Ideally, removing the app on failure is not recommended because it stops you from seeing what's wrong with the metadata/migrations, which beats the point of preview apps. You could delete it manually after seeing what's went wrong with the deployment, because if a deployment fails, you'd almost always want to see what went wrong.

What do you think?

META-DREAMER (approved):

META-DREAMER:
@wawhal The reason we added this was so it would properly re-deploy when new stuff was pushed, otherwise it would give an error that the app already exists instead of updating it, or the preview app would have some migrations applied that then got squashed or removed in the PR when someone pushes new commits and it would get stuck in a bad state.

I'm wondering if it would be possible to have more control over these preview deployments where we could manually trigger a redeploy by clicking a checkbox in the GitHub comment or pushing a commit with a certain string like `[redeploy]` etc.

Another issue we have when deleting the preview apps once the PR is merged or redeploying is the error `Could not create/delete ephemeral database(s). database "coordinape_627" is being accessed by other users`

wawhal:
@META-DREAMER apologies I missed this comment.

> The reason we added this was so it would properly re-deploy when new stuff was pushed, otherwise it would give an error that the app already exists instead of updating it, or the preview app would have some migrations applied that then got squashed or removed in the PR when someone pushes new commits and it would get stuck in a bad state.

Very valid concern and this is elegantly handled by Hasura Cloud preview apps system. It destroys the existing preview app with the given name before recreating a new one. So you need not remove an existing app yourself.

> Another issue we have when deleting the preview apps once the PR is merged or redeploying is the error Could not create/delete ephemeral database(s). database "coordinape_627" is being accessed by other users

This is a work in progress and it should be done soon. It's a known bug that we're working on that happens due to Postgres being used by Hasura when we try to drop the ephemeral database. As a workaround however, you can trigger the preview app workflow again from the CI by rerunning the failed job -- try to make sure you are not using the preview app when the preview workflow starts. Agree it's a little convenient, but it should be fixed in one of the upcoming releases.

PR #666, opened 2022-03-22
New History page

MarcogodoyV:
Removed the getAdminNavigation from paths.

Created the new design of the History Page based on the figma design (I couldn't made the SVG icon in stitches works so I added temporaly icons from MUI, this is going to be changed when the reviewed code comes). Thanks @levity for helping me with the branch `556-HistoryPage` (it was a really clear path to follow with stitches)


levity:
Merged into #652

PR #667, opened 2022-03-22
createCircle Allows both protocol id and name inputs

topocount:
Laravel currently allows both parameters in the createCircle REST
endpoint and ignores the protocol name in the case where both are
present.

Our frontend currently supplies all parameters, even if no change has
been made. This change allows for compat with frontend behavior while
behaving more soundly if the name does not match the name of the
protocol id supplied. I've also updated the error message to be more
contextually helpful in this case.

test plan:
test cases where
1. no name is suppled: the endpoint should behave as before, and create
   the circle if the protocol id points to and existing protocol
2. both are supplied and the name matches the existing protocol name:
   the new circle should point to the existing protocol
3. both are supplied and the protocol name does not match the existing
   name: The error message should inform the user that the user is not
   an admin of the protocol with id _and_ name instead of just id. It's
   not great ergonomics but it's better than just name.

### DIFF
--- a/api-lib/gql/queries.ts
+++ b/api-lib/gql/queries.ts
@@ -229,7 +229,8 @@ export async function getExpiredNominees() {
 
 export async function checkAddressAdminInOrg(
   address: string,
-  protocol_id: number
+  protocol_id: number,
+  protocol_name = '%%'
 ) {
   const { profiles } = await adminClient.query({
     profiles: [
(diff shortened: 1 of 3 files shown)


CryptoGraffe:
I ended up fixing part of this in my PR but not some of the details you added here

zashton (approved):
Thanks, works as described

PR #676, opened 2022-03-22
make remaining tokens uneditable during createUser

zashton:
### Test Plan

Go to hasura console run the createUser mutation with a `starting_token` value and ensure the `remaining_token` values follows the same


CryptoGraffe (approved):
tested locally and looks good to me

PR #677, opened 2022-03-22
[Long-running branch] Org view

levity:
Contains changes for the "Org view" [milestone](https://github.com/orgs/coordinape/projects/4?card_filter_query=milestone%3A%22org+view%22).

### n.b.

- `api/hasura/auth` and `lib/gql/client` have changes so that you have to explicitly query as a superadmin rather than having that be done by default. (otherwise superadmin users, like me, end up fetching all circles in the app when they just want the circles they're actually members of.


levity:
Screen Shot 2022-03-31 at 10 45 49 AM

Screen Shot 2022-03-31 at 10 50 29 AM

src/components/LoadingModal/LoadingModal.tsx
```diff
@@ -29,7 +29,7 @@ const useStyles = makeStyles(theme => ({
 export const LoadingModal = (props: {
   visible: boolean;
   onClose?: () => void;
-  text: string;
+  text?: string;
 }) => {
   const classes = useStyles();
```
teeolendo, line 34: Sorry, possible candidate for a stitches rewrite? If it's not too much trouble ofc. Entirely up to you.

teeolendo:
@levity I know the design is different but the menu items not aligning left is causing me a great deal of anxiety.

image

teeolendo:
> @levity I know the design is different but the menu items not aligning right is causing me a great deal of anxiety.
> 
> image

Actually, I think the designs has them aligning right, thankfully :D

levity:
> I know the design doesn't explicitly state this but I don't think the Current and Next sections should be lost when a user is paginating through past epochs

@teeolendo that's fair; i can make that change

src/components/LoadingModal/LoadingModal.tsx
levity: I'm going to hold off on this. The circular progress bar is a Material-UI component

levity:
> > @levity I know the design is different but the menu items not aligning right is causing me a great deal of anxiety.
> > image
> 
> Actually, I think the designs has them aligning right, thankfully :D

thanks for making the change!

src/stitches.config.ts
```diff
@@ -73,29 +73,34 @@ export const {
...
+      lightText: '#859296',
+      placeholder: '#99A2A5',
 
       background: '#DFE7E8',
-      pacificBlue: '#00ABBF',
       selected: '#31A5AC',
       lightBlue: '#84C7CA',
       darkBlue: '#5D9C9F',
-      green: '#29D07E',
+      blue: '#2c91ee', // Blue/12
```
teeolendo, line 98: What's with the comments?

levity: they correspond to color names in `src/ui/colors`, which are imported from Figma. i don't think they should all be included in the theme necessarily, so i'm just adding the comments to note when i've copied one over manually

src/pages/CirclesPage/getOrgData.ts
```diff
@@ -0,0 +1,37 @@
+import { order_by } from 'lib/gql/__generated__/zeus';
```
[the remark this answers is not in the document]
levity: hmm... i dunno, this query is really specific to the needs of this page. i feel like it's misleading to move it somewhere more general, because it suggests it should be reused. i liked the idea of organizing files by feature that someone posted in chat the other day (maybe it was @META-DREAMER) and this is in line with that.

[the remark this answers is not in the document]
levity: "ease of reference" is an important goal, but i think we get that well enough by searching for uses of client.query, e.g.:

Screen Shot 2022-04-01 at 12 47 09 PM

src/stitches.config.ts
teeolendo: Oh, I thanks for clarifying.

src/pages/CirclesPage/getOrgData.ts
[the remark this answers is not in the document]
teeolendo: Makes sense.

teeolendo (dismissed):

PR #681, opened 2022-03-22
Skip hasura previews on forks

crabsinger:
@META-DREAMER 

these will always fail on forks, so let's just always skip them.

Not sure if syntax is correct...

### DIFF
--- a/.github/workflows/hasura-preview.yml
+++ b/.github/workflows/hasura-preview.yml
@@ -2,6 +2,7 @@ name: 'hasura-cloud-preview'
 on:
   pull_request:
     types: [opened , reopened, synchronize]
+    if: ${{ !github.event.pull_request.head.repo.fork }}
     paths:
       - 'hasura/**' # path to Hasura config, prevents creating deployments if no changes
     branches:
--- a/.github/workflows/hasura-preview.yml
+++ b/.github/workflows/hasura-preview.yml
@@ -2,7 +2,7 @@ name: 'hasura-cloud-preview'
 on:
   pull_request:
     types: [opened , reopened, synchronize]
-    if: ${{ !github.event.pull_request.head.repo.fork }}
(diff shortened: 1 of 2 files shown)


.github/workflows/hasura-preview.yml
```diff
@@ -2,6 +2,7 @@ name: 'hasura-cloud-preview'
 on:
   pull_request:
     types: [opened , reopened, synchronize]
+    if: ${{ !github.event.pull_request.head.repo.fork }}
```
META-DREAMER, line 5: would be better to check if the head repo org/name matches the current repo with `github.event.pull_request.head.repo.full_name == github.repository`

META-DREAMER (approved):
feel free to merge after addressing comments

.github/workflows/hasura-preview.yml
META-DREAMER: Mostly because its more explicit in what its checking (that the repo the PR branch is from exactly matches our repo) vs this just checking if the incoming branch is a fork. There could be scenarios where someone creates a copy of our repo without forking it and make PRs from that.

crabsinger:
@META-DREAMER makes sense, thanks for adding your insight, pushed the fix, please take a look again.

hasura/metadata/actions.graphql
```diff
@@ -2,8 +2,6 @@ type Mutation {
   adminUpdateUser(payload: AdminUpdateUserInput!): UserResponse
 }
 
-# ignore; fake commit for triggering workflow
```
crabsinger, line 5: @META-DREAMER this got merged from previous PR, so removing now. https://github.com/coordinape/coordinape/pull/664/commits/bf3eb2a9101e348c3c04d9bab243c04dcd72ca57

META-DREAMER (approved):

PR #683, opened 2022-03-23
Fix: topocount reviews on api and hasura PRs

topocount:
In the prior config, zashton's assignments to the same folders overwrote my assignments
on prior lines. This allows both of us to be assigned on all changes to the applicable
folders.


crabsinger (approved):
makes sense! thanks!

PR #689, opened 2022-03-24
check constraint on profiles.website

### LINKED ISSUE #441: Migrate POST v2/profile
## Summary:
Migrate v2/profile Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `v2/profile` to Hasura.

**Middleware:**
auth token & **Current profile**

validate website url to be a valid website to prevent XSS injection

**Logic:**

**DB Tables:**
profiles

#### Notes:

 - Will the permission model for this be simple enough for it to be a mutate, or an action

CryptoGraffe:
This fixes #441 by obviating the need for a custom validation action. 

This includes migrations that fix existing rows that don't match this pattern. One side effect is the y.at emoji urls don't work. There were only 3 of them that weren't properly URL encoded, so they are cleaned up in this change explicitly. 

I tested this by importing a facsimile of the production data and running this migration. 

To test, run some mutations in the console that play around with variations on valid and invalid websites, e.g. 
```
mutation MyMutation {
  update_profiles_by_pk(
  pk_columns: {id: 2218},
 _set: {website: "http://slurp.next.mock.google.com?moo#jack"}) {
    id
    website
  }
}
```

### DIFF
--- a/hasura/migrations/default/1648096542742_alter_table_public_profiles_add_check_constraint_valid_website/up.sql
+++ b/hasura/migrations/default/1648096542742_alter_table_public_profiles_add_check_constraint_valid_website/up.sql
@@ -1,1 +1,3 @@
+update "public"."profiles" set website = CONCAT('https://',website) where website not ilike 'https://%' and website not ilike 'http://%' and website is not null;
+
 alter table "public"."profiles" add constraint "valid_website" check (website::text ~* 'https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{2,255}\.[a-z]{2,9}\y([-a-zA-Z0-9@:%_\+.,~#?!&>//=]*)$'::text);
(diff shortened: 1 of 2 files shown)


CryptoGraffe:
@zashton I tweaked this and I think it will work on staging/prod now.

zashton (dismissed):

hasura/migrations/default/1648096542742_alter_table_public_profiles_add_check_constraint_valid_website/up.sql
```diff
@@ -0,0 +1,8 @@
+update "public"."profiles" set website = CONCAT('https://',website) where website not ilike 'https://%' and website not ilike 'http://%' and website is not null;
+-- update these three specific users, making sure they have these y.at addresses
+update "public"."profiles" set website = RTRIM(website) where website is not null;
```
zashton, line 3: hmm why not use `TRIM` instead

zashton (approved):
looks good to me

hasura/migrations/default/1648096542742_alter_table_public_profiles_add_check_constraint_valid_website/up.sql
CryptoGraffe: because the only example I found was right trim 🤷‍♀️

PR #694, opened 2022-03-26
Setup Dev Portal page to document GraphQL API access

META-DREAMER:
Page is not accessible to general users from the main site navigation, you can only reach it by going to the /devPortal endpoint.

image

Clicking the link will take the users to a live GraphQL console where they can get their auth tokens and browse the schema / build queries and mutations + comes prefilled with an example query.

### DIFF
--- a/src/pages/DevPortalPage/DevPortalPage.tsx
+++ b/src/pages/DevPortalPage/DevPortalPage.tsx
@@ -46,10 +46,11 @@ export const DevPortalPage: React.FC = () => {
           use Coordinape&#39;s GraphQL API. Click the button below to visit our
           API Explorer where you will find the GraphQL API endpoint, the
           token/headers you can use to authenticate your request, and a live
-          console to construct queries.
+          console to construct queries. Note that your auth token will be
+          invalidated upon logout from the Coordinape website.
         </Box>
         <Box css={{ pt: '$md', color: '$text', fontStyle: 'italic' }}>
-          NOTE: This API is subject to change as we improve our data model.
+          This API is subject to change as we improve our data model.
         </Box>
         <Box
           css={{
--- a/src/pages/DevPortalPage/DevPortalPage.tsx
+++ b/src/pages/DevPortalPage/DevPortalPage.tsx
@@ -4,10 +4,13 @@ import { REACT_APP_HASURA_URL } from '../../config/env';
 import { useCurrentUserAuthToken } from '../../hooks/useAuthToken';
 import { Box, Panel, Button } from '../../ui';
 
+const EXAMPLE_QUERY =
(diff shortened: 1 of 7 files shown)


src/pages/DevPortalPage/DevPortalPage.tsx
```diff
@@ -0,0 +1,73 @@
...
+          }}
+        >
+          Developer Portal
+        </Box>
+        <Box css={{ pt: '$md', color: '$text' }}>
+          If you would like to access your circle data programmatically, you can
+          use Coordinape&#39;s GraphQL API. Click the button below to visit our
+          API Explorer where you will find the GraphQL API endpoint, the
+          token/headers you can use to authenticate your request, and a live
+          console to construct queries.
```
topocount, line 49: It might be worth adding a note that the token will be invalidated upon logout from the coordinape website.

src/routes/paths.ts
```diff
@@ -40,6 +40,7 @@ export const paths = {
   epoch: '/epoch',
   give: '/give',
   vouching: '/vouching',
+  devPortal: '/devPortal',
```
crabsinger, line 43: I prefer case insensitive URLs, so maybe `/devs`,  `/developers` or  `/api` as URLs to use instead?

crabsinger (approved):
@META-DREAMER this is awesome. love how we just prepopulate the auth token rather than having to show it.

I made one comment about the URL pattern.

src/pages/DevPortalPage/DevPortalPage.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { Box, Panel, Button } from '../../ui';
+
+const EXAMPLE_QUERY =
+  'query+MyCircles+%7B%0A++circles%28limit%3A+3%29+%7B%0A++++name%0A++++organization+%7B%0A++++++name%0A++++%7D%0A++++epochs%28limit%3A+3%2C+where%3A+%7Bended%3A+%7B_eq%3A+true%7D%7D%29+%7B%0A++++++id%0A++++++end_date%0A++++++start_date%0A++++%7D%0A++++users%28limit%3A+5%2C+order_by%3A+%7Bcreated_at%3A+asc%7D%29+%7B%0A++++++id%0A++++++name%0A++++++give_token_received%0A++++++give_token_remaining%0A++++++profile+%7B%0A++++++++twitter_username%0A++++++%7D%0A++++%7D%0A++%7D%0A%7D%0A';
+
+const getConsoleUrl = (authToken: string) => {
+  return `https://cloud.hasura.io/public/graphiql?endpoint=${REACT_APP_HASURA_URL}&header=Authorization:${encodeURIComponent(
+    `Bearer ${authToken}`
+  )}&query=${EXAMPLE_QUERY}`;
+};
```
topocount, line 14: kind of a silly nit: would it make sense to direct the console url towards `http://localhost:8080` if the uri contains localhost? That way the process for community devs to test apis using an app-level auth token could utilize this interface.

If this is out of scope, I understand.

topocount (approved):
looks good to me!, save for the nit below

src/pages/DevPortalPage/DevPortalPage.tsx
META-DREAMER: The localhost:8080 Hasura instance console should be disabled / should not be used, for local dev people should use `yarn hasura console` so that any changes they make in the tables etc get persisted as yaml / sql in their local filesystem. 

If you want to use app level auth, you can just add `X-Hasura-Role` and even `X-Hasura-UserId` headers in the console and you can "act as" any specific user.

levity: wait, can you say more about this? i thought localhost:8080 and yarn hasura console were the same. when i make changes in the former, they show up in `yarn hasura metadata export`.

levity: ah, got it, thanks! will switch to that.

PR #695, opened 2022-03-26
Implement End Epoch Cron Task

### LINKED ISSUE #524: Migrate End Epoch Cron Trigger
## Summary:
Migrate Check EPOCH (Start/Ending) Laravel Cron Trigger functionality to Hasura.

### Details
This task is to migrate the functionality of Check EPOCH (Start/Ending) to Hasura.

**Middleware:** Hasura event secret middleware

Nil

**Logic:**
*/5 * * * *
Iterate through `epochs` that has  `end` = false but has past its `end_date`.  Duplicate over all records of `pending_gift_tokens` to `token_gifts` and purge away the existing `pending_gift_tokens` of this circle. Check if circle's `auto_opt_out` = true , if true set `users` inside the circle that didn't allocate this epoch (`starting_tokens` == `give_tokens_remaining`) to opt out(`non_receiver`= 1). 

Create a snapshot of the user's `bio` record and save in the `histories` table. Update all users in the circle and reset their `give_tokens_received` to zero , and `give_token_remaining` = `starting_tokens`, `epoch_first_visit` to 1 and `bio` to `null`. Set `number` to epoch if not yet by counting through past ended `epochs` and set it to `end` = true. 

Check if the circle of the epoch has telegram_id or discord_webhook set and ended = true & check protocols as well for telegram_id for notifications

an e.g of discord/telegram notification 
```
protocol_name / circle_name epoch has just ended!
Users that did not allocate any GIVE Tokens:
xxx, yyy, xxxuser_name

```
(`repeat` = 0 = 1-time, `repeat` = 1 = weekly, `repeat` = 2 = monthly )
Check if epoch is a `repeat` epoch, if its a weekly epoch (for the new repeated epoch) add 7 days to the `start_date`, if monthly add 1 month to it ( use the last day of the month or the specific end day of the existing epoch whichever is earlier.). Do validation checks if the new repeated epoch `start_date` doesn't intercept the `end date` of the current epoch, check if the repeated epoch intercepts with any other existing epochs in the circle. (Take Note that the `end_date` of the ending epoch is allowed to be exactly the same value as the newly created epoch's `start_date` time also need to be factored in. )

If no issues create the repeated epoch and send notification via  to alert the start of a new epoch.
e.g of an discord/telegram notification
`A new repeating epoch has just been created 2022-02-24 22:11:00 to 2022-03-03 22:11:00
`

**DB Tables:**
`circles`,`epochs`

#### Notes:
@jsur

topocount:
closes #524

test plan:

1. set up a an epoch with ended=false and an end_date < `now()` in the
   database. Add some pending gifts to the db with this epoch id if none
   already exist. Check that some users in the relevant circle have less
   than the starting amount of give remaining. in the relevant circle
   set auto_opt_out = true.
2. run `vercel dev`
3. observe the bios get transferred to histories with the relevant
   epoch_ids. Observe pending_token_gifts for the epoch get transferred
   to token_gifts. observe users who didn't spend share any give get
   opted out. observe all users in the circle get reset with
   epoch_first_visit = true and all token memos reset.

### DIFF
--- a/api/hasura/cron/epochs.ts
+++ b/api/hasura/cron/epochs.ts
@@ -4,6 +4,7 @@ import type { VercelRequest, VercelResponse } from '@vercel/node';
 import dedent from 'dedent';
 import { DateTime, Settings } from 'luxon';
 
+import { ValueTypes } from '../../../api-lib/gql/__generated__/zeus';
 import { adminClient } from '../../../api-lib/gql/adminClient';
 import { errorLog } from '../../../api-lib/HttpError';
 import { sendSocialMessage } from '../../../api-lib/sendSocialMessage';
@@ -20,6 +21,7 @@ async function handler(req: VercelRequest, res: VercelResponse) {
   const results = await Promise.all([
     notifyEpochStart(epochResult),
     notifyEpochEnd(epochResult),
+    endEpoch(epochResult),
   ]);
 
   res.status(200).json({ message: results });
@@ -104,16 +106,37 @@ async function getEpochsToNotify() {
           {
             where: {
               end_date: { _lt: 'now()' },
-              notified_end: { _is_null: true },
+              ended: { _eq: false },
             },
           },
           {
+            id: true,
+            circle_id: true,
             repeat: true,
+            number: true,
             repeat_day_of_month: true,
+            start_date: true,
+            end_date: true,
             circle: {
+              id: true,
+              name: true,
+              token_name: true,
+              auto_opt_out: true,
               telegram_id: true,
               discord_webhook: true,
-              users: [{}, { circle_id: true, bio: true }],
+              organization: { name: true, telegram_id: true },
+              users: [
+                {},
+                {
+                  id: true,
+                  name: true,
+                  non_receiver: true,
+                  circle_id: true,
+                  bio: true,
+                  starting_tokens: true,
+                  give_token_remaining: true,
+                },
+              ],
             },
             epoch_pending_token_gifts: [
               {},
@@ -122,6 +145,7 @@ async function getEpochsToNotify() {
                 epoch_id: true,
                 note: true,
                 sender_id: true,
+                tokens: true,
                 sender_address: true,
                 recipient_id: true,
                 recipient_address: true,
@@ -208,7 +232,7 @@ export async function notifyEpochEnd({
           message,
           { discord: true },
           epoch,
-          updateEpochEndNotification
+          updateEpochEndSoonNotification
         );
 
       if (circle.telegram_id)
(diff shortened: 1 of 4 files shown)


api/hasura/cron/epochs.ts
```diff
@@ -230,13 +254,246 @@ export async function notifyEpochEnd({
...
+  end_date: string;
+  circle_id: number;
+  repeat: number;
+  repeat_day_of_month: number;
+  circle?: { telegram_id?: string; discord_webhook?: string };
+}) {
+  const { start_date, end_date, repeat, repeat_day_of_month, circle } = epoch;
+  const start = DateTime.fromISO(start_date);
+  const end = DateTime.fromISO(end_date);
+  const days = end.diff(start, 'days');
```
zashton, line 399: we should use the `days` field from the previous epoch so its more consistent

api/hasura/cron/epochs.ts
```diff
@@ -230,13 +254,246 @@ export async function notifyEpochEnd({
...
+    const usersWithStartingGive = [] as Array<string>;
+    // if auto_opt_out is true, set users where `starting_tokens == give_tokens_remaining to non_receiver = true
+    // copy user bios to histories
+    // reset give_tokens_received = 0, give_token_remaining = starting tokens, epoch_first_visit = 1, bio = null
+    const userUpdateMutations = circle.users.reduce((ops, user) => {
+      const userUserHasAllGive =
+        user.give_token_remaining === user.starting_tokens;
+      if (userUserHasAllGive) usersWithStartingGive.push(user.name);
+      const optOutMutation =
+        !user.non_receiver && circle.auto_opt_out && userUserHasAllGive
```
zashton, line 290: this should be `!user.non_giver` instead of `!user.non_receiver`

api/hasura/cron/epochs.ts
```diff
@@ -230,13 +254,246 @@ export async function notifyEpochEnd({
...
+        {
+          pk_columns: { id: epoch.id },
+          _set: {
+            ended: true,
+          },
+        },
+        { __typename: true },
+      ],
+      __alias: {
+        ...userUpdateMutations,
```
zashton, line 336: nice that the bulk mutations can be chained together like this

zashton (changes requested):
tested the repeating and end epoch notifications works , some minor changes.
And the user mutations seem to trigger off 1 event trigger for each user, not sure will it be an issue just that the server gets hit with a sudden spike of queries

api/hasura/cron/epochs.ts
topocount: This is actually a small optimization to ensure that users who are already opted out don't create an un-needed event trigger request, so I'll keep it and add the non_giver condition on top of it.

zashton (approved):

PR #697, opened 2022-03-27
enable event triggers as FE already ported over

zashton:


api/hasura/event_triggers/eventManager.ts
```diff
@@ -35,6 +30,12 @@ const STAGING_HANDLERS: HandlerDict = {
   refundGiveTelegram,
```
CryptoGraffe, line 30: I'm not 100% sure on the dependencies here but do we need deleteUser to be impld ?

CryptoGraffe (approved):
Looks good just one comment

api/hasura/event_triggers/eventManager.ts
zashton: theres an issue because if admin updates user to opt out the refund wouldn't happen

PR #698, opened 2022-03-27
Compare contributors names by English locale

### LINKED ISSUE #673: BUG: Admin member list view should ignore casing when sorting
**Summary:**

The user view on a circle admin page does not ignore case when sorting

uppercase names show up before lowercase names

case should be ignored when sorting

4. **Steps to recreate:**

1. visit the admin page for the Conn - Friesen circle 
2. create a user with a lowercase name (e.g. 'aaa').
3. and see that all the capitalized users are first. If you scroll to page 8, the aaa name will be after the Z names

5. **Screenshots/GIF/Video recording**
image

sshmm:
fixes #673
**Summary**
use  Intl.Collator('en-u-kf-upper') to compare the contributors names strings by English locale so that only strings that differ in base letters compare as unequal.
kf-upper: to sort upper case letters first

### DIFF
--- a/src/pages/AdminPage/AdminPage.tsx
+++ b/src/pages/AdminPage/AdminPage.tsx
@@ -202,6 +202,8 @@ const epochDetail = (e: IEpoch) => {
       }`;
 };
 
+const englishCollator = new Intl.Collator('en-u-kf-upper');
+
 const AdminPage = ({ legacy }: { legacy?: boolean }) => {
   const classes = useStyles();
   const [keyword, setKeyword] = useState<string>('');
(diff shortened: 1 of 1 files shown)


levity (commented):
Neat. Any idea how this performs on long lists? Is it worth instantiating the collator once rather than repeatedly inside the comparison function?

sshmm:
> Neat. Any idea how this performs on long lists? Is it worth instantiating the collator once rather than repeatedly inside the comparison function?
thanks for the info. I updated it to be instantiated once then passed to sort function

levity (approved):

levity:
thanks @sshmm!

PR #699, opened 2022-03-28
Update teammate Hasura Migration

### LINKED ISSUE #544: Migrate Update Teammates v2/{circle_id}/teammates
## Summary:
Migrate Update Teammates Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `<POST v2/{circle_id}/teammates>` to Hasura.

**Middleware:** Hasura event secret middleware

- check if the two users are in the same circle
- check whether any of the user_ids is the own user

**Logic:**
Endpoint accepts an array of `user_ids`, resyncs the `teammates` table with the new `user_ids` . Any existing teammates not in the `user_ids` and has received tokens from the user will be refunded. Purge all pending_gifts rows that are associated with those removed teammates and the current user as sender. Recalculate their `give_tokens_received` and `give_tokens_remaining` accordingly.

**DB Tables:**
`users`, `pending_token_gifts`, `teammates`

#### Notes:

zashton:
fixes #544 

### Test plan
1. Test using a circle that has more than 3 users.
2. Create an epoch in the FE http://localhost:3000/admin/circles, set the start date/time so it begins immediately.
3. Update your teammates via the mutation, note the teammates you add must be in the same circle 
```gql
mutation MyMutation {
  updateTeammates(payload: {circle_id:<circle_id>, teammates: [<user_id>, <user_id>]}) {
    user {
      teammates {
        teammate {
          id
          name
        }
      }
    }
  }
}
```
4.  Go to http://localhost:3000/team to check if your teammates is updated 
5. Go to the allocation page and allocate some tokens to all your teammates and save your allocation
6. run the mutation again and but this time remove a few of your teammates.
7. check the `pending_token_gifts` table to check if your tokens are correctly refunded.  Check the `users` table and check if `give_token_received`, `give_token_remaining` are correctly updated. 
8. go back to http://localhost:3000/team to check if your teammates are correctly updated.

Theres a new unique constraint added in this PR
```sql
alter table "public"."teammates" add constraint "teammates_user_id_team_mate_id_key" unique ("user_id", "team_mate_id");
```
To check if there are any existing records that violate this constraint run this sql query
```sql
SELECT t1.* FROM public.teammates t1
JOIN (
    SELECT user_id, team_mate_id
    FROM public.teammates
    GROUP BY user_id, team_mate_id
    HAVING COUNT(*) > 1
) t2 ON t1.user_id = t2.user_id AND t1.team_mate_id = t2.team_mate_id
```

### DIFF
--- a/api-lib/event_triggers/removeTeammate.ts
+++ b/api-lib/event_triggers/removeTeammate.ts
@@ -1,0 +1,124 @@
+import assert from 'assert';
+
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { ValueTypes } from '../gql/__generated__/zeus';
+import { adminClient } from '../gql/adminClient';
+import { errorResponse } from '../HttpError';
+import { EventTriggerPayload } from '../types';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  // no parsing should be needed here since this data comes straight from
+  // the database and zeus keeps this consistent for us
+  const {
+    event: { data },
+  }: EventTriggerPayload<'teammates', 'DELETE'> = req.body;
+
+  const { user_id, team_mate_id } = data.old;
+
+  const results = [];
+  try {
+    const { users_by_pk } = await adminClient.query({
+      users_by_pk: [
+        {
+          id: user_id,
+        },
+        {
+          pending_sent_gifts: [
+            {
+              where: {
+                recipient_id: {
+                  _eq: team_mate_id,
+                },
+              },
+            },
+            {
+              id: true,
+              recipient_id: true,
+              tokens: true,
+            },
+          ],
+          circle: {
+            epochs: [
+              {
+                where: {
+                  end_date: { _gt: 'now()' },
+                  start_date: { _lt: 'now()' },
+                },
+              },
+              {
+                id: true,
+              },
+            ],
+          },
+        },
+      ],
+    });
+    assert(users_by_pk, 'panic: user must exist');
+
+    const { pending_sent_gifts, circle } = users_by_pk;
+
+    const currentEpoch = circle.epochs.pop();
+    if (!currentEpoch || pending_sent_gifts.length === 0) {
+      res.status(200).json({
+        message: `Not a refund event.`,
+      });
+      return;
+    }
+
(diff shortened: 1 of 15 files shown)


api-lib/event_triggers/removeTeammate.ts
```diff
@@ -0,0 +1,124 @@
...
+                pk_columns: { id: gift.recipient_id },
+                _inc: { give_token_received: -gift.tokens },
+              },
+              { give_token_received: true },
+            ],
+          };
+        return ops;
+      },
+      {} as { [aliasKey: number]: ValueTypes['mutation_root'] }
+    );
```
topocount, line 88: looks like we might want to create a library function suite for this pattern, haha

api-lib/findUser.ts
```diff
@@ -72,3 +72,34 @@ export const getUserFromAddress = async (address: string, circleId: number) => {
...
+export const getUsersFromUserIds = async (
+  userIds: number[],
+  circleId: number
+) => {
+  const { users } = await adminClient.query({
+    users: [
+      {
+        where: {
+          id: { _in: userIds },
+          circle_id: { _eq: circleId },
```
topocount, line 85: nit: the circle_id shouldn't be needed, since users are all tied to a single circle. I don't think adding it should even optimize the query at all, given that the index still needs to be traversed to resolve the user ids regardless.

api/hasura/actions/updateTeammates.ts
```diff
@@ -0,0 +1,79 @@
...
+        affected_rows: true,
+      },
+    ],
+    insert_teammates: [
+      {
+        objects: insertInput,
+        on_conflict: {
+          constraint: teammates_constraint.teammates_user_id_team_mate_id_key,
+          update_columns: [],
+        },
```
topocount, line 65: nice

topocount (dismissed):
looks, good pending an optional nit. Acceptance tests all pass.

api-lib/findUser.ts
zashton: I added the `circle_id` as a check to validate the `teammate_ids` sent by the client are indeed same circle as the user and so invalid users (from other circles) could be filtered out.

PR #703, opened 2022-03-29
Enable Previous Distributions

### LINKED ISSUE #506: Build (Vaults) Submit Distribution: Part 2.2 Create Merkle Distribution
### Specification

- you may want to call this without a previous distribution arg just to make sure that it matches the values that were shown in the table
- use `createDistribution` in `lib/merkle-distributor` 
    - for values to be submitted to the chain, call this with `previousDistribution` set to the stored distribution data for the previous epoch, if one exists ([why?](https://www.notion.so/Vault-Contracts-Walkthrough-705a0b4abfa04adb9db7dbc025d95da6))
- actually submit to blockchain (`ApeDistributor.uploadEpochRoot`) 
- make sure the number of decimal places matches token's decimal places

teeolendo:
- Move the submit distribution function from the Distribute page to its own hook
- Migrated a few of the mutations to the global user files

TODO:
- Add columns to claims table to enable previous distribution tracking
- Enable Previous Distributions

closes #506 and #507


hasura/migrations/default/1648654670645_new_amount_and_distribution_id/down.sql
```diff
@@ -0,0 +1,15 @@
+
+-- Could not auto-generate a down migration.
```
META-DREAMER, line 2: Wondering if this would cause issues if we wanted to rollback. Not an SQL pro so not sure, but if we can would be good to write manual down migrations here

src/pages/DistributePage/DistributePage.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    .mockReturnValue({
+      data: mockEpoch,
+    })
+    .mockReturnValueOnce({
+      data: mockEpoch,
+    })
+    .mockReturnValueOnce({
+      data: null,
+    }),
+}));
```
META-DREAMER, line 57: Would be nice if this was moved to a dedicated fixtures files so it could be shared between multiple page tests.

src/pages/DistributePage/queries.ts
```diff
@@ -85,3 +90,38 @@ export function useGetAllocations(epochId: number) {
...
+  });
+  return distributions;
+};
+
+export type PreviousDistribution = Exclude<
+  Awaited<ReturnType<typeof getPreviousDistribution>>,
+  undefined
+>[0];
+
+export function usePreviousDistributions(circleId: number) {
```
META-DREAMER, line 121: Should allow nullish values here since it handles them with `enabled`

```suggestion
export function usePreviousDistributions(circleId: number | null | undefined) {
```

src/pages/DistributePage/DistributePage.tsx
```diff
@@ -48,29 +48,30 @@ function DistributePage() {
...
     isLoading: isAllocationsLoading,
     isError,
     data,
   } = useGetAllocations(Number(epochId));
 
-  const currentUser = useCurrentUserForEpoch(Number(epochId));
+  const {
+    data: previousDistribution,
+    isLoading: loadingPreviousDistributions,
+  } = usePreviousDistributions(Number(data?.epochs_by_pk?.circle?.id));
```
META-DREAMER, line 68: Wrapping the data result in Number casts the potentially undefined type to be a number. `usePreviousDistributions` should be updated to handle undefined type instead of wrapping this in Number

src/pages/DistributePage/queries.ts
levity: good catch

src/pages/DistributePage/DistributePage.test.tsx
levity: these are pretty specific to this page; i think it's unlikely to be very reusable. maybe just the mock user & epoch? i have some similar mocks in `utils/testing/recoil` that might be reusable for this purpose

hasura/migrations/default/1648654670645_new_amount_and_distribution_id/down.sql
teeolendo: Yeah, I saw this and at the time I felt that it wasn't that big a deal cause worst case, you just have to write alter table statements, remove constraints and drop tables/columns. not a huge lift but open to hearing if this should be included now.

levity (approved):

Issue #558, opened 2022-02-22
"Coordinape In Your Circle": ENABLE/DISABLE in members list in admin view


spydermonkee:
**Problem**
We've added a Coordinape user in all circles to enable people to give to Coordinape if they feel inspired as a token of thanks or just to support Coordinape. The problem is that we have not given our Coordinape Humans enough context as to what this is all about and many admins have simply deleted it. Others have ignored. A few have given to Coordinape! 

**Specification**
- Add the ability to "Enable" or "Disable" the Coordinape member if you are an admin. 
- Add tooltip next to ENABLE/DISABLE explaining the option. 

This issue is a little more complex since it may involve some deeper changes to implement the state of "Coordinape as a Circle Member". 

For the enable, this would require a new hasura mutation that restores the user, mutation would set`deleted_at` back to a `null` value. This mutation can be for generic use in restoring normal users.
(validate that the address is still unique within the circle if restored).

For the disable you can use the hasura mutation that deletes a normal user.

Design: 
https://www.figma.com/file/Db8KthuZRjT5CfMvZsjTT2/Coordinape-Design-v7?node-id=2692%3A9277

topocount:
@spydermonkee how about enabling/disabling the coordinape user with the `non_receiver` field in the backend?

spydermonkee:
Would we then create F/E logic that says IF "coordinape user" AND `non_receiver` THEN don't display in map and show disabled in member list... Or something to that effect? 
 I think it doesn't matter much how the backend implements if the experience result is what we are trying to get.

topocount:
I just think `non_receiver` is a more discrete way to track this

topocount:
@amrro if you comment on this PR requesting it be assigned to you we can now do so.

amrro:
Thanks @topocount, 
Sure, I would like to hack this one.

spydermonkee:
@amrro - Heyo! I see you've been doing some good stuff lately! Thanks a lot! Are you still working on this issue?

amrro:
You are welcome. 

I haven't started working on it. I was Waiting to #739 to merge first.

reeserj:
Looks like it's merged now! @amrro!

amrro:
Hey @reeserj, Not sure what you are talking about. 
I guess #739 requires some changes and updating and I think it's ready.

reeserj:
@amrro are you unblocked on this now?

amrro:
[...]

PR #708, opened 2022-03-29
Refactor createUser to use existing softdeleted users

topocount:
In order to prevent duplicate user entries with the same circle_id and
address from showing up in the database, createUser should recover
soft-deleted users instead of creating a new entry with the same address
alongside the soft-deleted entry.

This is relevant for the coordinape user in #558

test plan:
1. create a user in an existing circle:
    ```gql
    mutation MyMutation {
      createUser(payload: {
        address: "0x02675dfa4d4d6b3fb7d04e0c8ad000eff752bd89", 
        circle_id: 1, 
        name: "dopeocount"
      }) {
        id
        UserResponse {
          address
        }
      }
   }
   ```
2. soft-delete the user:
    ```sql
    UPDATE users
    SET deleted_at = 'now()'
    WHERE name = 'dopeocount'
    ;
    ```
3. recreate the user:
   ```gql
    mutation MyMutation {
      createUser(payload: {
        address: "0x02675dfa4d4d6b3fb7d04e0c8ad000eff752bd89", 
        circle_id: 1, 
        name: "dopeocount"
    }) {
        id
        UserResponse {
          address
        }
      }
    }
   ```
4. Verify that the same row in the database is returned and no duplicate
   rows exist.

### DIFF
--- a/api/hasura/actions/createUser.ts
+++ b/api/hasura/actions/createUser.ts
@@ -1,6 +1,7 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node';
 
 import { authCircleAdminMiddleware } from '../../../api-lib/circleAdmin';
+import { ValueTypes } from '../../../api-lib/gql/__generated__/zeus';
 import { adminClient } from '../../../api-lib/gql/adminClient';
 import {
   createUserSchemaInput,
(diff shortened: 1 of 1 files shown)


zashton (approved):
Works great, great idea to reuse soft deleted users

PR #711, opened 2022-03-30
prototyping moving mutations into their own files

CryptoGraffe:
Trying this out... if there is a better model, let me know. I didn't want to import/export these in an index.ts because then that file would have frequent conflicts. Let me know if there is a better idiom here.


levity:
LGTM 👍

PR #716, opened 2022-03-30
Add sending daily updates to organisation level telegram

zashton:
Adding the functionality of sending telegram daily updates to organisation level telegram so its in line with how it works previously (affects Yearn circles only)


topocount (approved):
approved. apologies for missing this in the spec.

Also, I don't think technically this is yearn-only functionality (gated on a specific circle or org id), but I see how it is practically.

PR #717, opened 2022-03-30
add circles-nominees hasura relation & add missing foreign keys

levity:
Needed to add the circles-nominees relation to support a page I'm building for #677. 

I verified in production that no existing data violates these new constraints, i.e. the following queries all return 0:

```
select count(*) from nominees where circle_id not in (select id from circles);
select count(*) from vouches where voucher_id not in (select id from users);
select count(*) from vouches where nominee_id not in (select id from nominees);
```


hasura/metadata/databases/default/tables/public_circles.yaml
```diff
@@ -51,6 +51,15 @@ array_relationships:
       table:
         name: circle_integrations
         schema: public
+- name: nominees
+  using:
+    manual_configuration:
+      column_mapping:
+        id: circle_id
+      insertion_order: null
+      remote_table:
```
topocount, line 60:
```suggestion
    foreign_key_constraint_on:
      column: circle_id
      table:
```

topocount (approved):
Since you're adding the relationship in at the DB level, allow hasura to just utilize that to populate the api interface.

Otherwise looks good.

Issue #671, opened 2022-03-22
BUG: Show error in nominate form when ETH address invalid


CryptoGraffe:
**Summary:**
When nominating a user there is no appropriate error text when the eth address is invalid 

1. **What is actually happening:**
No error is presented, but Nominate Member button is disabled

3. **What should be happening:**
There should be an error inline in the form

4. **Steps to recreate:**
Go to nominate a user, input an invalid ETH address

5. **Screenshots/GIF/Video recording**
<img width="729" alt="image" src="https://user-images.githubusercontent.com/95149165/159428039-6c899280-afb5-4d99-bfce-1d6a9b132681.png">

sshmm:
is anyone working on this?

sshmm:
@spydermonkee  when nominating a new user. the description should be at least three characters. that is the only condition
is this ok?
shouldn't it be a sentence at least
or may be it can be left empty?

spydermonkee:
@sshmm - It's fine to just have validation that "something is in there" for now. Validating at least 3 characters is a decent enough proxy for "something". We are planning more changes with more specifics in the coming months, so it's fine for now. Thanks for working on this!

sshmm:
@spydermonkee  I changed it in the PR commit to 40 characters as suggested by Lawris

PR #718, opened 2022-03-30
bugFix: Show error in nominate form

sshmm:
Summary:
Fixing BUG: Show error in nominate form when ETH address invalid #671
by Reusing createForm for NewNominateModal instead of custom checks.

Name, address and description will be checked on change and error text
will appear untill the values are accepted and submit button is disabled
untill then.

### DIFF
--- a/src/forms/NewNominateForm.tsx
+++ b/src/forms/NewNominateForm.tsx
@@ -1,0 +1,28 @@
+import { z } from 'zod';
+
+import { createForm } from './createForm';
+import { zEthAddress } from './formHelpers';
+
+const schema = z
+  .object({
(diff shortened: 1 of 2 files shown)


sshmm:
@levity  would you have a look at this?

levity:
hi @sshmm, thanks for the PR. the code looks pretty good at first glance. i should have mentioned though, when we first talked about this, that we're trying to move away from using our custom `createForm` and `FormModal` code, because it's opaque & coupled with Recoil, and use `react-hook-form` directly instead.

for the modal part, you can use `src/ui/Modal`, and for using react-hook-form directly, you can check out how it's done in `src/pages/DistributePage`.

i started a thread in Discord about this too, so feel free to ask more questions there.

PR #721, opened 2022-03-31
fixes #633 by integrating updateProfile graphql into FE

### LINKED ISSUE #633: Integrate Hasura Update Profile into FE
## Summary:
Integrate Hasura Update Profile into FE.

### Details
This task is to integrate the frontend of Update profile to Hasura.
https://app.coordinape.com/profile/me

**Screenshot of Page**
image

**Logic:**
-

**Related Backend Issue:**
#441

CryptoGraffe:
Fixes #633 ~~This is blocked by #689  and shouldn't be tested until then.~~

To test, 
1. visit profiles page and set/unset a bunch of fields and change settings
2. ensure that network traffic is going to /graphql instead of laravel
~~3. this will let in bogus website values until #689 is qualified and merged~~
4. Try to set website to valid url, invalid url, and then also after setting it, try clearing it out

### DIFF
--- a/hasura/metadata/databases/default/tables/public_profiles.yaml
+++ b/hasura/metadata/databases/default/tables/public_profiles.yaml
@@ -78,5 +78,24 @@ update_permissions:
       id:
         _eq: X-Hasura-User-Id
     set:
-      id: x-hasura-User-Id
+      id: X-Hasura-User-Id
   role: user
+- permission:
+    check: null
+    columns:
+      - avatar
+      - background
+      - bio
+      - discord_username
+      - github_username
+      - medium_username
+      - skills
+      - telegram_username
+      - twitter_username
+      - website
+    filter:
+      id:
+        _eq: X-Hasura-User-Id
+    set:
+      id: X-Hasura-User-Id
+  role: superadmin
--- a/hasura/metadata/databases/default/tables/public_profiles.yaml
+++ b/hasura/metadata/databases/default/tables/public_profiles.yaml
@@ -77,4 +77,6 @@ update_permissions:
     filter:
       id:
         _eq: X-Hasura-User-Id
+    set:
+      id: x-hasura-User-Id
   role: user
--- a/src/components/EditProfileModal/EditProfileModal.tsx
+++ b/src/components/EditProfileModal/EditProfileModal.tsx
@@ -133,7 +133,7 @@ export const EditProfileModal = ({
             />
             <FormTextField
               {...fields.website}
-              placeholder="Enter link"
+              placeholder="https://website.com"
               label="Website"
             />
           </div>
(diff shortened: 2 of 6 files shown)


src/components/EditProfileModal/EditProfileModal.tsx
```diff
@@ -56,9 +56,13 @@ export const EditProfileModal = ({
...
-      submit={params =>
-        updateMyProfile(params).then(onClose).catch(console.warn)
-      }
+      submit={params => {
+        const fixedParams: any = { ...params };
+        const skills = params.skills;
+        delete fixedParams.skills;
+        fixedParams.skills = JSON.stringify(skills);
+        updateMyProfile(fixedParams).then(onClose).catch(console.warn);
+      }}
```
topocount, line 65: @CryptoGraffe I hope this refactor isn't too over the top:
```suggestion
            submit={async params => {
        const fixedParams: Omit<typeof params, 'skills'> & { skills: string } =
          { ...params, skills: JSON.stringify(params.skills) };
        try {
          await updateMyProfile(fixedParams);
          onClose();
        } catch (e: unknown) {
          console.warn(e);
        }
      }}

```

CryptoGraffe: ok I applied this, plus a change that allows nulling of website via empty string

hasura/metadata/databases/default/tables/public_profiles.yaml
```diff
@@ -77,4 +77,6 @@ update_permissions:
     filter:
       id:
         _eq: X-Hasura-User-Id
+    set:
+      id: x-hasura-User-Id
```
zashton, line 81: lets make it consistent `X-Hasura-User-Id`,

think its also missing an update permission for `superadmin`here, permissions can be the same as `user` role

zashton (changes requested):
Some UX suggestions, the website placeholder can have an example of a valid website.
e.g `https://google.com` so the user understands the https part is required.

The error message might be too technical for the user to understand when the website violates the regexp. (not sure if this is possible to customize)

The other parts are working great, just need to update the permissions.

hasura/metadata/databases/default/tables/public_profiles.yaml
CryptoGraffe: ok i didn't change these but I can sync them up

CryptoGraffe: We can only customize the error message in the client side javascript. There is a LOT of this kind of stuff... every zod error is pretty tough too. Not sure I want to take this on right now

zashton: > ok i didn't change these but I can sync them up


yeah np, all the permission settings was done by me during the start, probably was me that left that out

zashton (approved):
all concerns addressed, tested & working 👍

PR #729, opened 2022-03-31
set superadmin role only if requested

levity:


crabsinger (approved):

api/hasura/auth.ts
```diff
@@ -47,9 +47,15 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
     });
 
     assert(profile, 'Profile cannot be found');
+
+    const role =
+      profile.admin_view && req.headers['X-Preferred-Role'] === 'superadmin'
```
crabsinger, line 52: clients can't just set this as a header can they?

levity: they could, but it wouldn't have any effect if they don't have profile.admin_view = true

PR #730, opened 2022-04-01
Implement Login endpoint on top of hasura

### LINKED ISSUE #562: Migrate Login v2/login
## Summary:
Migrate Login Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `<POST v2/login>` to Hasura.

**Middleware:** wallet signature (we could upgrade to using EIP-712 typed signatures for better security)

-

**Logic:**
Check if user address exists in `profiles` table. if profile exists revoke(delete) the token  in `personal_access_tokens` (that is linked to this profile) else create a new `profile` record with the address. 
Generate a random 40 character stringToken with different cases and hash it with sha256 algo.
`const hashedToken = crypto.createHash('sha256').update(stringToken).digest('hex');`

insert a `personal_access_tokens` record
column values
`tokenable_type: App\Models\Profile`
`token_id: {profile_id}`
`token: {hashedToken}`

return rawtoken value `{access_token_id}|{stringToken}` so the FE can cache it for future auth with hasura.
it should look something like `123|abcdefg...`

**DB Tables:**
profiles, personal_access_tokens

#### Notes:
In future we could update to using JWT so tokens can be generated directly in FE.

topocount:
resolves #562

The location of this endpoint might not belong in the hasura actions
subpath.

The message validation ensures the message contains "Login to Coordinape"
and also validates the the timestamp is in the past.

test plan:

1. generate a signed login payload from within the frontend and save it
   as a json file. It's preferrable to use a throwaway address for this.
2. Run `curl -X POST http://localhost:3000/api/hasura/actions/loginUser
   -H "Content-Type: application/json" -d @<my file name>.json
3. You should receive a token in response formatted for use in the
   browser. Try multiple login attempts and ensure the old db entry is
   deleted and a new one is created with a new ID and a fresh hashed
   token.
4. try to crack the authentication:
    a. change the length of signature payload
    b. move chars around in the signature
    c. modify the message/data payload

### DIFF
--- a/src/lib/zod/index.ts
+++ b/src/lib/zod/index.ts
@@ -5,6 +5,26 @@ import {
   zStringISODateUTC,
 } from '../../../src/forms/formHelpers';
 
+const PERSONAL_SIGN_REGEX = /0x[0-9a-f]{130}/;
+export const loginInput = z.object({
+  address: zEthAddressOnly,
+  data: z.string().refine(
+    msg => {
+      const templateOk = msg.startsWith('Login to Coordinape');
+      const timestamp = msg.split(' ').pop();
+      const timestampInPast =
+        timestamp && Number.parseInt(timestamp) * 1000 < Date.now();
+
+      const validLength = msg.length === 30;
+
+      return templateOk && validLength && timestampInPast;
+    },
+    { message: 'Invalid message payload' }
+  ),
+  hash: z.string().length(0),
+  signature: z.string().regex(PERSONAL_SIGN_REGEX),
+});
+
 export const createCircleSchemaInput = z
   .object({
     user_name: z.string().min(3).max(255),
@@ -184,6 +204,16 @@ export function composeHasuraActionRequestBody<T extends z.ZodRawShape>(
       HasuraUserSessionVariables,
     ]),
     request_query: z.string().optional(),
+  });
+}
+
+export function composeCrossClientAuthRequestBody<T extends z.ZodRawShape>(
+  inputSchema:
+    | z.ZodObject<T, 'strict' | 'strip'>
+    | z.ZodEffects<z.ZodObject<T, 'strict' | 'strip'>>
+) {
+  return z.object({
+    input: z.object({ payload: inputSchema }),
   });
 }
 
--- a/api/hasura/actions/loginUser.ts
+++ b/api/hasura/actions/loginUser.ts
@@ -1,0 +1,133 @@
+import assert from 'assert';
+import { randomBytes, createHash } from 'crypto';
+
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+import {
+  bufferToHex,
+  fromRpcSig,
+  fromAscii,
+  ecrecover,
+  hashPersonalMessage,
+  toBuffer,
+  pubToAddress,
+} from 'ethereumjs-util';
+import { DateTime, Settings } from 'luxon';
+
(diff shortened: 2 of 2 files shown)


levity:
I agree this doesn't belong under api/hasura. How about just api/login?

zashton (approved):
works for me, just that payload needs to be wrapped in an input object, 
fine with moving out of the hasura/actions since its not really an action handler

Would need to handle the hash too (a smart contract wallet) if its sent

this might help
https://docs.authereum.com/verifying-signatures

src/lib/zod/index.ts
```diff
@@ -5,6 +5,26 @@ import {
...
+      const timestampInPast =
+        timestamp && Number.parseInt(timestamp) * 1000 < Date.now();
+
+      const validLength = msg.length === 30;
+
+      return templateOk && validLength && timestampInPast;
+    },
+    { message: 'Invalid message payload' }
+  ),
+  hash: z.string().length(0),
```
zashton, line 24: the hash would have a value if its a smart contract wallet

topocount: Ah thanks, didn't know we supported this.

zashton: > Ah thanks, good point. Maybe we'll implement a handler for this when there's a use-case for smart contracts delegating others to obtain api keys?

Its currently in use though, initially we didn't have this, until the users that were using special wallets like [Argent](https://www.argent.xyz/) were complaining they couldn't log into our app using their wallet via walletconnect. So we added this part.

topocount: I'm adding this now, since it's so straightforward

topocount:
@CryptoGraffe the login payload here is wrapped so we can use this function to generate api keys from within a hasura authenticated context if we want.

During the frontend input, the login should be wrapped like:
```ts
input: {
  payload: {
    ...originalLoginPayload
  }
}
```

src/lib/zod/index.ts
zashton: > I'm adding this now, since it's so straightforward

@topocount 
theres a helper function i created that you might be able to reuse
https://github.com/coordinape/coordinape/blob/main/src/utils/provider.ts#L40

topocount:
@zashton @levity I deployed some test contracts to rinkeby to test https://github.com/coordinape/coordinape/pull/730/commits/f77b4fabf4f3fe04d14975a75b7e489f89a2fe53

https://rinkeby.etherscan.io/address/0xc92be0eb67f6932a3f4a236b3c3d8c0962625741#code
https://rinkeby.etherscan.io/address/0xf0e0e0d1a8d2747ecb4ce8710ec44fa71d1ea28f#code

I was a bit confused because of  the recent updates to EIP 1271 so I just figured we should support the original working spec and the finalized spec. Feel free to use them to test.

src/lib/zod/index.ts
topocount: @zashton sorry I missed this. I reimplemented this in the api-lib/sigatures.ts file a bit differently.

zashton (dismissed):
Thanks for the quick update, i've tested it on a mainnet Argent wallet, the legacy interface always fails but the final abi works. 
@topocount

topocount:
> Thanks for the quick update, i've tested it on a mainnet Argent wallet, the legacy interface always fails but the final abi works. @topocount

Maybe I should log a PR on argent's side with the updates to their docs...

PR #732, opened 2022-04-01
make pending_gifts.epoch_id non-nullable

CryptoGraffe:
Checked on prod that this will be ok by running this mutation:

```
query MyQuery {
  pending_token_gifts_aggregate(where: {epoch_id: {_is_null: true}}) {
    aggregate {
      count
    }
  }
}
```

and got this result:

```
{
  "data": {
    "pending_token_gifts_aggregate": {
      "aggregate": {
        "count": 0
      }
    }
  }
}
```


zashton (approved):
lg2m

PR #735, opened 2022-04-04
Bug/671/show errors nominate modal

### LINKED ISSUE #671: BUG: Show error in nominate form when ETH address invalid

sshmm:
update NomianteModal to show errors to fix  #671  and use ui library with stitches

Update NewNominationModal to use ui components like Modal and Form. Change the styling from make Styles to stitches. Add FormLabel in ui library to use as labels for input fields.

### DIFF
--- a/src/pages/VouchingPage/NewNominationModal.tsx
+++ b/src/pages/VouchingPage/NewNominationModal.tsx
@@ -83,7 +83,9 @@ export const NewNominationModal = ({
         }}
         onSubmit={handleSubmit(onSubmit)}
       >
-        <Text css={{ lineHeight: 'normal' }}>{nominateDescription}</Text>
+        <Text css={{ lineHeight: 'normal', mb: '$lg' }}>
+          {nominateDescription}
+        </Text>
         <Box
           css={{
             display: 'grid',
--- a/src/pages/VouchingPage/NewNominationModal.tsx
+++ b/src/pages/VouchingPage/NewNominationModal.tsx
@@ -1,95 +1,19 @@
-import clsx from 'clsx';
 import { ethers } from 'ethers';
 import isEmpty from 'lodash/isEmpty';
 import { useForm, SubmitHandler, Controller } from 'react-hook-form';
 
-import { makeStyles } from '@material-ui/core';
-
-import { FormTextField } from 'components';
 import { useApiWithSelectedCircle } from 'hooks';
 import { useSelectedCircle } from 'recoilState/app';
-import { Form, Button, Modal } from 'ui';
-
-const useStyles = makeStyles(theme => ({
-  description: {
-    marginTop: theme.spacing(1),
-    fontSize: 16,
-    fontWeight: 400,
-    color: theme.colors.primary,
-    textAlign: 'center',
-  },
-  quadGrid: {
-    width: '100%',
-    display: 'grid',
-    gridTemplateColumns: '1fr 1fr',
-    gridTemplateRows: 'auto auto',
-    columnGap: theme.spacing(3),
-    rowGap: theme.spacing(4),
-  },
-  gridAllColumns: {
-    gridColumn: '1/-1',
-  },
-  modal: {
-    display: 'flex',
-    alignItems: 'center',
-    justifyContent: 'center',
-  },
-  closeButton: {
-    color: theme.colors.mediumGray,
-    top: 0,
-    right: 0,
-    position: 'absolute',
-  },
-  body: {
-    position: 'relative',
-    display: 'flex',
-    flexDirection: 'column',
-    alignItems: 'center',
-    backgroundColor: 'white',
-    width: '100%',
-    borderRadius: 8,
-    padding: theme.spacing(0, 8, 3),
-    overflowY: 'auto',
-    maxHeight: '100vh',
-  },
-  large: {
-    maxWidth: 1140,
-    padding: theme.spacing(0, 12, 3),
-    [theme.breakpoints.down('sm')]: {
-      padding: theme.spacing(0, 6, 3),
-    },
-  },
-  medium: {
-    maxWidth: 820,
-  },
-  small: {
-    maxWidth: 648,
-  },
-  title: {
-    margin: theme.spacing(3, 0, 2),
-    fontSize: 30,
-    fontWeight: 700,
-    lineHeight: 1.2,
-    color: theme.colors.text,
-    textAlign: 'center',
-  },
-  subtitle: {
-    margin: theme.spacing(0, 0, 2),
-    fontSize: 16,
-    fontWeight: 300,
-    lineHeight: 1.2,
-    color: theme.colors.text,
-    textAlign: 'center',
-  },
-  errors: {
-    display: 'flex',
-    flexDirection: 'column',
-    justifyContent: 'flex-end',
-    margin: 0,
-    minHeight: 45,
-    color: theme.colors.red,
-  },
-}));
+import {
+  Form,
+  Button,
+  Modal,
+  Text,
+  TextField,
+  FormLabel,
+  Box,
+  TextArea,
+} from 'ui';
 
 export const NewNominationModal = ({
   onClose,
@@ -98,7 +22,6 @@ export const NewNominationModal = ({
   visible: boolean;
   onClose: () => void;
 }) => {
-  const classes = useStyles();
   const { circle } = useSelectedCircle();
   const { nominateUser } = useApiWithSelectedCircle();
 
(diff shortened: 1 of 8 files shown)


src/ui/index.ts
```diff
@@ -10,6 +10,8 @@ export * from './Panel/Panel';
 export * from './SvgIcon/SvgIcon';
 export * from './Text/Text';
 export * from './TextField/TextField';
+export * from './FormLabel/FormLabel';
```
levity, line 13: please keep this file in alphabetical order

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -58,70 +40,116 @@ export const NewNominationModal = ({
           : ''
       }`
     : '';
+  type NominateFormValues = {
```
levity, line 43: Types should be defined at the top-level of a file, not inside a component

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -58,70 +40,116 @@ export const NewNominationModal = ({
...
+    name: string;
+    address: string;
+    description: string;
+  };
+  const intialValues = {
```
levity, line 48:
```suggestion
  const initialValues = {
```

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -58,70 +40,116 @@ export const NewNominationModal = ({
...
-          inputProps={{
-            maxLength: 280,
-          }}
-          fullWidth
-        />
-      </div>
-    </FormModal>
+    <Modal title="Nominate New Member" open={visible} onClose={onClose}>
+      <Form nominate onSubmit={handleSubmit(onSubmit)}>
+        <Text>{nominateDescription}</Text>
```
levity, line 72: Please add `lineHeight: 'normal'` to the styles here so that the text does not look so crowded

src/ui/Form.tsx
```diff
@@ -1,2 +1,20 @@
 import { styled } from 'stitches.config';
-export const Form = styled('form', {});
+export const Form = styled('form', {
+  variants: {
+    nominate: {
```
levity, line 4: In this case, rather than creating a new `Form` component, I would suggest using a regular `<form>` tag without any styles, and just adding styles to a `<Box>` that's either a parent or child of `<form>`. There's no advantage to coupling these styles to the `<form>` tag.

src/ui/Form.tsx
```diff
@@ -1,2 +1,20 @@
...
+        display: 'flex',
+        flexDirection: 'column',
+        alignItems: 'center',
+        backgroundColor: 'white',
+        width: '100%',
+        borderRadius: 8,
+        padding: '0 0 $lg',
+        overflowY: 'auto',
+        maxHeight: '100vh',
+        maxWidth: '648px',
```
levity, line 16: you should be able to avoid hardcoding a value like this, and instead rely on the fact that `Modal` already has a maxWidth set.

src/ui/Box/Box.tsx
```diff
@@ -1,5 +1,36 @@
...
+        width: '100%',
+      },
+    },
+    errors: {
+      true: {
+        display: 'flex',
+        flexDirection: 'column',
+        justifyContent: 'center',
+        margin: 0,
+        minHeight: 45,
```
levity, line 19: not sure what this accomplishes

levity (changes requested):
Thanks for submitting this!

In general, my feedback for this is that many of the styles in variants should instead be added directly to their components in `NewNominationModal` using the `css` prop. Variants should only be created when a group of styles is meant to be reused in different contexts. 

So the variants named `nominate` should definitely just be added directly in NewNominationModal, since they're specific to that component's needs. Many of the other ones should be as well, like `input` and `errors` in Box. They happen to be useful for wrapping inputs and errors in NewNominationModal, but it's unlikely that they'll be used in exactly the same way elsewhere. Even if it seems a little repetitive, it's better to err on the side of defining styles inline, at least while our design system is still in flux.

src/ui/Form.tsx
sshmm: @levity I am not using a new form component. I am using ui/Form that already exists should I change that?

src/ui/Box/Box.tsx
sshmm: it's meant to center then input and the label, keep them in a column..

src/ui/Form.tsx
levity: ah, i see. yes, i think it's ok to keep using this, but putting the styles on the tag in the NewNominateModal rather than creating the variant here.

src/ui/Box/Box.tsx
levity: i mean minHeight specifically. were you trying to prevent the form from resizing when errors appear/disappear?

sshmm: I tried to replicate the old styling.But you are right, this should be removed because it make the error text and textarea above when resizing the page to smaller heights

sshmm:
> Thanks for submitting this!
> 
> In general, my feedback for this is that many of the styles in variants should instead be added directly to their components in `NewNominationModal` using the `css` prop. Variants should only be created when a group of styles is meant to be reused in different contexts.
> 
> So the variants named `nominate` should definitely just be added directly in NewNominationModal, since they're specific to that component's needs. Many of the other ones should be as well, like `input` and `errors` in Box. They happen to be useful for wrapping inputs and errors in NewNominationModal, but it's unlikely that they'll be used in exactly the same way elsewhere. Even if it seems a little repetitive, it's better to err on the side of defining styles inline, at least while our design system is still in flux.

@levity thanks for the through review. I applied these suggestion in 5a5e0c98b07a3fcd1d344a26c1983d8ad6101567

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -59,69 +52,209 @@ export const NewNominationModal = ({
...
+                required: 'Name must be at least 3 characters long.',
+                minLength: {
+                  value: 3,
+                  message: 'Name must be at least 3 characters long.',
+                },
+              }}
+              control={control}
+              render={({ field }) => (
+                <TextField
+                  css={{ height: '48px', width: '100%' }}
```
levity: instead of `48px` you can use the `$2xl` size token here

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -59,69 +52,209 @@ export const NewNominationModal = ({
...
+                fontSize: '$4',
+                fontFamily: 'Inter',
+                fontWeight: '$bold',
+                textAlign: 'center',
+                m: '0 0 $sm',
+              }}
+            >
+              Name
+            </FormLabel>
+            <Controller
```
levity: Can you use the `useController` hook instead? It's a lot nicer IMO because the form logic is more separated from presentation, and also there's less nesting of components. Check out what we did in VaultsPage/CreateForm for an example.

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -59,69 +52,209 @@ export const NewNominationModal = ({
...
+              'grid-column': '1 / -1',
+              display: 'flex',
+              flexDirection: 'column',
+              alignItems: 'center',
+              width: '100%',
+            }}
+          >
+            <FormLabel
+              htmlFor="description"
+              css={{
```
levity, line 175: Even though this isn't a variant anymore, it is still the same the 3 times it's used in this file, so it would be good to pull it out into a variable to reduce duplication.

sshmm:
@levity  I added some margin between the Nomination description and the inputs and above the description input to match the previous design.
Separated the labels and inputs of name and address into different rows so that the address input box won't be pushed down on resizing to very narrow width
image

levity (changes requested):
Thanks for the latest updates! I have a few more suggestions.

src/pages/VouchingPage/NewNominationModal.tsx
sshmm: I updated it to use useController with zodresolver instead of passing "rules" to useController

src/pages/VouchingPage/NewNominationModal.tsx
sshmm: added labelStyles const

src/pages/VouchingPage/NewNominationModal.tsx
```diff
@@ -1,33 +1,30 @@
...
-  gridAllColumns: {
-    gridColumn: '1/-1',
-  },
-}));
+type NominateFormValues = {
+  name: string;
+  address: string;
+  description: string;
+};
+const intialValues = {
```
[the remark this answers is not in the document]
sshmm: I removed this cont and passed default values to useController directly

levity (approved):

levity:
nice work @sshmm!

PR #739, opened 2022-04-05
Update Admin Page Attribute & support Mobile

### LINKED ISSUE #442: Admin Contributor Attributes Updates
### Summary 
We are changing the phrasing of contributor attributes to add more context and make managing users easier for admins. 

### Specification
The contributor's permissions are confusing to first-time users and the cognitive load to read the chart is high because of the excessive text. Therefore, I propose changing the Admin Table attributes, Add/Edit Contributor Modal, and Circle Settings. 

**Admin Page Contributor Table Changes** 
Non Giver ----> GIVER
Opted Out ---->  GIVE Recipient
Are they Admin ----> Admin

Instead of having "Yes" or "No." 
Yes ----> ✓
No  ----> ✖

The table's header will be sticky, so the chart header is always in view as you scroll. 

**Admin Page Tooltips**
Tooltips should appear on the following headers on the following headers: Giver, Give Recipient, and Admin. 

- GIVER - Circle Member allocating GIVE
- GIVE Recipient - Circle Member receiving GIVE
- Admin - As a Circle Admin, you will be able to edit Circle Settings, Edit Epoch settings, edit your users, and create new circles. Learn More - https://docs.coordinape.com/welcome/admin_info

**Add Contributor Modal** 
The following inputs will change from "Yes/No" into checkboxes. The following titles will also change. 

- Starting Tokens ---->  GIVE Allotment 
- Are They Admin?  ---->  Grant Administrative Permissions  / Connected to Admin on chart 
- Can Give? ----> Allow contributor to give tokens / Connected to GIVER
- Opted out ----> Allow contributor to receive tokens / Connected to GIVE Recipient 
- Force Opted out ---->  Removed

**Add Contributor Modal Tool Tips**
- Grant Administrative Permissions  - As a Circle Admin, you will be able to edit Circle Settings, Edit Epoch settings, edit your users, and create new circles. Learn More - https://docs.coordinape.com/welcome/admin_info
-  Allow contributor to give tokens - Allows the Contributor to get paid based on the amount of giving allocated by circle members  Learn More- https://docs.coordinape.com/welcome/gift_circle#the-gift-circle
- Allow contributor to receive tokens - Gives the member the ability to reward circle members with giving. Learn More - https://docs.coordinape.com/welcome/gift_circle#the-gift-circle
- GIVE Allotment - The maximum amount of giving a user can allocate in an epoch Learn More -https://docs.coordinape.com/welcome/gift_circle#the-gift-circle

**Circle Settings Changes** 
The following settings will change in vocabulary. 
Default Opt-In ----> Default Contributors GIVE

**Add Contributor Modal Tool Tips**
Default Contributors GIVE - Allows all new contributors to get paid based on the amount of giving allocated by circle members. Learn More https://docs.coordinape.com/welcome/admin_info#editing-circle-settings

[Desktop Prototype](https://www.figma.com/proto/KciBaNjP5Q9idchsxMp9ek/Coordinape-Tool-Tip-Additions?page-id=0%3A1&node-id=28%3A2967&viewport=241%2C48%2C1.23&scaling=min-zoom&starting-point-node-id=28%3A2967&show-proto-sidebar=1&hide-ui=1)

[Mobile Prototype](https://www.figma.com/proto/KciBaNjP5Q9idchsxMp9ek/Coordinape-Tool-Tip-Additions?page-id=0%3A1&node-id=49%3A1046&viewport=241%2C48%2C1.23&scaling=min-zoom&starting-point-node-id=49%3A1046&show-proto-sidebar=1&hide-ui=1)

[Add Contributor Modal](https://www.figma.com/proto/KciBaNjP5Q9idchsxMp9ek/Coordinape-Tool-Tip-Additions?page-id=0%3A1&node-id=7%3A794&viewport=241%2C48%2C1.23&scaling=min-zoom&starting-point-node-id=7%3A794&show-proto-sidebar=1&hide-ui=1)

[Circle Settings Modal](https://www.figma.com/proto/KciBaNjP5Q9idchsxMp9ek/Coordinape-Tool-Tip-Additions?page-id=0%3A1&node-id=125%3A912&viewport=241%2C48%2C1.23&scaling=min-zoom&starting-point-node-id=125%3A912&show-proto-sidebar=1&hide-ui=1)

### User Story
As an admin, I can easily see who can allocate, give and receive it. 

### Acceptance Criteria
1. Scenario: Admin wants to see if Leslie can GIVE 
Given: The user navigates to the circle admin page
And: On a desktop device 
When: The user scrolls to the Admin chart 
Then: The system shows a checkmark under GIVER
When: The user hovers over the "i" icon of GIVER
Then: The system shows the definition of GIVER 

2. Scenario: Admin wants to see if Leslie can receive GIVE
Given: The user navigates to the circle admin page
And: On a desktop device 
When: The user scrolls to the Admin chart 
Then: The system shows an X mark under GIVE Recipient 
When: The user hovers over the "i" icon of GIVE Recipient 
Then: The system shows the definition of GIVE Recipient 

3. Scenario: Admin wants to take away Leslie's GIVE ability 
Given: The user navigates to the circle admin page
When: The user scrolls to the Admin chart  
And: Clicks Edit 
Then: The opens the Edit/Add Contributor Modal 
When: The Admin Unchecked Allow contributor to give tokens 
And: Clicks Save
Then: The modal closes the chart shows an X next to GIVER

4. Scenario: Admin wants to see if Ivy has Admin permissions 
Given: The user navigates to the circle admin page
When: The user scrolls to the Admin chart  
Then: The header stays in the same place
When: The Admin gets to Ivy 
Then: The Admin can click a checkmark next to Ivy's name
When: The user hovers over the word Admin 
Then: The system shows the definition of Admin

4. Scenario: Admin wants to have all members not allocated on default 
Given: The user navigates to the circle admin page
When: The clicks Circle Settings 
Then: The Admin hovers over the word Default Contributors GIVE
When: The system shows the definition of Default Contributors GIVE
Then: The Admin clicks No 
And: Save

### Functional Requirement
Tooltips must appear long enough for users to navigate the tooltip link for approximately 2 seconds. In addition, the admin chart header must stay in the same place so users can see permission headers as they scroll. 

### Assumption
The developer can configure tooltip interactions in seconds. For example, developers can easily add the material design "i" icon to buttons that Front-End can change the vocabulary without significantly impacting the function of those permissions. 

### Stakeholders
@AlexDistill , the art director  &  Sasche, UI Designer

amrro:
resolves #442
____
### Here are a list of changes: 
- update `StaticTable` headers' names. 
- rename giver/receiver/admin columns' headers and use ✅ & ❌ instead of text in rows. I used emoji for simplicity's sake.
- Update `StaticTable` to accept tooltip messages for the specified columns. It will show before the order arrow.
- Update `StaticTable` to use single column only and hide headers for Mobile edition. 
- Implement the new Epoch & Contributor for Mobile edition. 
- update action column to use edit/delete text instead of icons buttons. In addition, a transparent variant has been added to `ui/Button`. 
- 🆕 Add new component that based on Stitches to `ui`: 
  - `<Tooltip>`: with support to mobile. 
  - `<Avatar>`
  - `<Flex>`: simple new `<div>` just like `<Box>` but this could increase readability
  -  Coherently with the new designs, some of components has been updated for the new vairant. 
  - 'ui/Checkbox' based on 
- Added a new, simple `<ui/Checkbox>`, with support for tooltips. 
- Updated the Add Contributor modal to use match the provided design and utiltise the new `ui/Checkbox` instead of toggle views.
- New dependencies has been added like: `@radix-ui/react-avatar`, `@radix-ui/react-hover-card`, and SVG icons library `@radix-ui/react-icons`.
- Finally, some changes to the global configs that needs closer review like disabled color value, or space around `<ApeTextField>`.

Things that didn't changed: 
- *Forced Opt-out* is not changed left as text. Simply, it doesn't fit as only yes or no.

### Add Contributor Modal
Screen Shot 2022-05-10 at 10 35 33 PM

Screen Shot 2022-05-10 at 6 16 33 PM

Screen Shot 2022-05-10 at 10 30 46 PM


levity:
Thanks @amrro! Will take a look at this soon. Btw, in general it's helpful to include a screenshot in the PR when you've made visual changes.

src/pages/AdminPage/CircleComponent.tsx
```diff
@@ -0,0 +1,235 @@
+import { GearIcon, InfoCircledIcon } from '@radix-ui/react-icons';
```
levity, line 1: Thanks for breaking things out into a separate file. Since this file doesn't contain a component named CircleComponent, please rename it to something like `components`, `helpers`, etc.

src/components/StaticTable/StaticTable.tsx
```diff
@@ -162,49 +165,83 @@ export const StaticTable = ({
...
+            <tr
+              className={clsx(classes.headerTr, {
+                [classes.emptyHeader]: !data.length,
+              })}
+            >
+              {columns.map(
+                (
+                  {
+                    label,
+                    tooltip,
```
levity, line 179: I wonder if, rather than heavily modifying StaticTable this way, it makes more sense to just stop using it and write your own table in AdminPage that's specific to that page's needs. When I see this ad-hoc "column data format" with tooltip fields, it makes me think we're trying to create too much abstraction/reusability, & the code would be much simpler to understand if we didn't.

src/ui/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const avatarPath = getAvatarPathWithFallback(
+    profile?.avatar || user?.profile?.avatar,
+    user?.name
+  );
+  const src = path ?? avatarPath;
+
+  return (
+    <AvatarRoot>
+      <AvatarImage src={src} alt={user?.name} />
+      <AvatarFallback>avatarPath</AvatarFallback>
```
levity, line 59: ?

src/components/ApeCheckbox/ApeCheckbox.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
```
levity, line 1: what was your reasoning for creating this with Material-UI rather than Stitches?

levity (commented):
I'm still reviewing this, but have some high-level questions & thoughts.

src/components/ApeCheckbox/ApeCheckbox.tsx
amrro: No Reasoning actually. When I started working on this issue, I wasn't totally aware of the migration to stitches. But I will change that to stitches too.

src/ui/Avatar/Avatar.tsx
amrro: If you asking about the component, this represents the fallback to failure situation to display just the initials.

If you asking about using `avatarPath` as a fallback, the `getAvatarPathWithFallback()` will return initials in case of path absence. 

https://github.com/coordinape/coordinape/blob/258fad5b4572ed35324f562ec75e086c7f483f70/src/utils/domain.ts#L56-L60

Besides, I'm not fan of this implementation; It knows too much about the logic behind determining path and user. I would change it to more generic function that accepts only path and initials and change `ApeAvatar` to utilize it instead. What do you think?

amrro:
Thanks @levity for your review. 
I added some screenshots to the PR. waiting for you insights.

levity:
@amrro i hope to have some feedback from the design team on this soon. in the meantime i see there are some conflicts that need to be resolved; can you take care of that when you have a moment?

amrro:
Yes. I'm going to fix it.

amrro:
Updated `ui/Avatar`: 
- red ring around avatar when hover over it. 
- Clicking on avatar should navigate to profile.

src/ui/Avatar/Avatar.tsx
amrro: @levity I updated `ui/Avatar` to only accept `path` & `name`.

levity: Thanks for updating the avatar. My "?" comment was about the presence of the string literal "avatarPath" here. I think you're just missing curly braces? But even then I don't understand why you would print the path. Maybe I'm missing something about how AvatarFallback works.

amrro: Yes, my mistake. It's fixed now.
In addition, `getAvatarPathWithFallback()` is kinda confusing. I should think of better naming.

src/components/StaticTable/StaticTable.tsx
levity: I think it'd really be better to deprecate StaticTable entirely rather than augmenting it here. It's pretty coupled with Material-UI, which we're moving away from, and honestly I find the separation of `columns` and `data` to be unnecessarily complex. If you can make your changes to AdminPage with a normal table or grid setup, we'll be pretty close to being able to remove StaticTable from the codebase entirely.

For mobile: In my experience, sometimes you want a table that changes its layout for mobile, and sometimes you just want a totally different thing to be rendered. I think it's ok to be a little repetitive, for the sake of readability & avoiding excess abstraction.

levity: Anyway, I appreciate you being willing to make bold changes to the code.

amrro:
Hey @levity , Please let me know if there are any thing missing to fix/update in order to merge this PR.

levity:
> Hey @levity , Please let me know if there are any thing missing to fix/update in order to merge this PR.

Hey @amrro, please check out the feedback here: https://github.com/coordinape/coordinape/issues/442#issuecomment-1095807648

For #2, you can get the token name from the circle with `selectedCircle.tokenName`. 

Speaking of token name, this wasn't addressed in the original mockup (our mistake), but circles can change the name of their token from GIVE, so the word "GIVE" shouldn't be in the table column headers. We can call them instead:
- Giver
- Receiver
- Sent
- Received

Please also see the comments above:
- using Stitches for ApeCheckbox
- changes to StaticTable

For the second one, I still think it would be ideal not to change StaticTable, and just have AdminPage render one thing on mobile and another one on desktop in a simple way. Ideally we'd stop using StaticTable entirely. I know this is more work, but it would really help us move away from legacy code. In addition to the issues I mentioned above, StaticTable is also problematic because its use of GenericObject breaks type-checking. So the sooner we stop using it the better.

levity:
> I still think it would be ideal not to change StaticTable, and just have AdminPage render one thing on mobile and another one on desktop in a simple way. Ideally we'd stop using StaticTable entirely. I know this is more work, but it would really help us move away from legacy code. In addition to the issues I mentioned above, StaticTable is also problematic because its use of GenericObject breaks type-checking. So the sooner we stop using it the better.

Though I would understand if you didn't want to take on that larger refactoring. One advantage of the code you've already written is... it's already written. 😉  So I'd like to hear what you think about this.

amrro:
Thanks for briefing this up for me. I would be happy to work on pushing away from the legacy code if this really helps.

I thought about deeply hacking some of these changes in distinct PR (I wasn't clear enough). However, I am going work on these changes. 

Totally agree with your views on `<StaticTable>`: My first impression wasn't great while interacting with it; Too much abstraction that requires too much knowledge in order to utilize and update it.

Finally, I am not fully aware of the review process of the Design Team. I took Zemm's suggestions, under consideration but waited for some direction (they are very valid actually).

levity:
I made an partial alternative to StaticTable (no search or pagination) in #800 (check out Table and AllocationsTable); what do you think about this approach?

amrro:
@levity Thanks for letting me know. 

I was going to ask you the same about your opinion on my implementation here coordinape@a9eee56b260d6168ed09bd028ee33704ed218716.

My motivation was to build a Lego-style table components with the minimum iteration of variants that are easier to customize and agnostic about its children views in order to include any views with different functionalities e.g sorting. 
 
Here is an WIP example for the epochs table: 
```ts
import * as Table from 'ui/Table/Table';

// ...

    <Table.Table>
      <Table.Root>
        <Table.Header>
          <Table.Row>
            <Table.HeaderCell align="left">Epoch Details</Table.HeaderCell>
            <Table.HeaderCell>Status</Table.HeaderCell>
            <Table.HeaderCell align="left">Dates</Table.HeaderCell>
            <Table.HeaderCell area="narrow">Actions</Table.HeaderCell>
          </Table.Row>
        </Table.Header>
        <Table.Body>
          {epochs.map(e => {
            return (
              <Table.Row key={e.id}>
                <Table.Cell key={`details-${e.id}`} align="left">
                  {RenderEpochDetails(e)}
                </Table.Cell>

                <Table.Cell key={`status-${e.id}`}>
                  {RenderEpochStatus(e)}
                </Table.Cell>

                <Table.Cell key={`date-${e.id}`} align="left">
                  {RenderEpochDates(e)}
                </Table.Cell>

                <Table.Cell key={`actions-${e.id}`}>
                  {RenderEpochActions(e)}
                </Table.Cell>
              </Table.Row>
            );
          })}
        </Table.Body>
      </Table.Root>
    </Table.Table>
```
For your table implementation: I like it; but there is one downside IMHO, we've different variants like area: narrow & wide, align: left & center variants that we need to implement. Epochs table is example with cells with different styles. 
 
**what do you reckon about that?**
___
For pagination, I implemented a new Paginator but later I discovered that there is already one of your implementation using stitches 🤡. 
However, I guess It is not located under the convenient structure: It's under 'components' folder not 'ui' directory. I am going to utilize it. It's used in the history page.

For sorting, this table doesn't need to know about it. because cell is the playground for any view we need to add, right? 

Finally, after working with the epoch table, my target is to create a completely isolated Epochs table. This table has like 3 actions: edit, delete and export with their dialog. The table's logic was hardly interwoven with other components inside '<AdminPage>' and double that because there is the Contributors Table also.

amrro:
Hey @levity 
I updated the checkbox and the tables 💆🏻‍♂️. Please take a look.

There is a bug with the new `ui/Tooltip` it doesn't play nicely with Modal views: It get's dismissed when clicked. can't figure it out.

levity:
> It's under 'components' folder not 'ui' directory

The rule I'm following here is that `ui` should contain only styles (i.e. the output of `styled(tagName, { ... })`, not React components with any logic or state.

> For your table implementation: I like it; but there is one downside IMHO, we've different variants like area: narrow & wide, align: left & center variants that we need to implement. Epochs table is example with cells with different styles.

Yes, the `Cell` component you added could be useful for this. In some cases this may also be accomplished by 

In general I'm trying to err on the side of not wrapping tags in `styled` until I have a need for them. I like how we can make use of descendant selectors to affect normal tags, i.e.:

```
styled('table', {
  'th:first-child': {
    // ...
  }
}```

...so that the `th` tag inside the table can just be a normal JSX tag.

We don't need to reconstruct the same level of abstraction that is in StaticTable using Stitches; instead, let's aim to improve the simplicity and readability of what's in AdminPage. If that means there will be a bit more repetition in styles or logic on different pages that have tables, that's fine.

**Comment by amrro** [comment]
> The rule I'm following here is that `ui` should contain only styles (i.e. the output of `styled(tagName, { ... })`, not React components with any logic or state.

This clears it out. Nice sharing. 

I was going to name them as original name ex: `<td>` would be a `<Td>`. But I think. But I think we can combine them both, you can use normal table's tags, or import a `Cell` for specific usage.

**Comment by amrro** [comment]
Hello @levity, 
I've added two sizes variant to `ui/Avatar`. 

Please reconsider merging this PR if it's functionally OK. In Addition, I can make the rest of changes in different, atomic PR; Keeping track of this PR is becoming really tiresome and requires a lot of effort, I have to track the changes manually and unexpected bugs maybe introduced. 

Here's plans list after merging this:
- Resolve #558. 
- Apply `ui/Avatar` everywhere and delete other versions. 
- Update `Tooltip` to be less-abstracted and solve some bug that links inside tooltips inside modal is not working just dismissing tooltip dialog. 
- Totally migrate `AdminPage` to stitches, few left actually. 

Also, `StaticTable` is only used on table, I can hack this too and totally delete.

File: src/forms/AdminUserForm.tsx [hunk 5]
```diff
@@ -33,7 +33,7 @@ const AdminUserForm = createForm({
     address: v.user?.address ?? '',
     non_giver: v.user?.non_giver ?? false,
     fixed_non_receiver: !!v.user?.fixed_non_receiver ?? false,
-    non_receiver: !!v.user?.fixed_non_receiver || !!v.user?.non_receiver,
+    non_receiver: !!v.user?.fixed_non_receiver || !!v.user?.non_receiver || !v.circle.default_opt_in,
```
levity, line 36: Won't this change all existing users to `non_receiver=true` if the circle has `default_opt_in=false`?

levity:
Hi @amrro, hope you enjoyed your time off.

As you've noticed, the codebase has been changing a lot in the past few weeks, as we prepare for some larger UX changes. I appreciate you going through the effort of resolving conflicts. However, #856 is about to be merged, and this is going to produce other conflicts. I think we need to wait for that.

Sorry that this has gone on for so long without merging. I wish you'd let me know earlier that you were going to be absent for 2 weeks; I would have tried to merge parts of your PR earlier.

Anyway, I will do another review of the code. You should probably look at #856 because the changes to styles are pretty major.

src/ui/Paginator/Paginator.tsx
```diff
@@ -0,0 +1,162 @@
+import React from 'react';
```
levity, line 1: Please move this to `pages/AdminPage`, or remove it and use `components/Paginator` instead

src/ui/Table/Table.tsx
```diff
@@ -0,0 +1,97 @@
+import { styled } from '../../stitches.config';
```
levity, line 1: We're going to use [the Table that I created for DistributionsPage](https://github.com/coordinape/coordinape/blob/main/src/pages/DistributionsPage/Table.tsx) instead. @teeolendo is reusing it in #883. If you want to keep using this one for AdminPage, can you move it to that folder?

src/ui/Checkbox/CheckBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+        onCheckedChange={onChange}
+        disabled={disabled}
+        id={label}
+      >
+        <CheckboxIndicator>
+          <CheckIcon />
+        </CheckboxIndicator>
+      </CheckboxRoot>
+      <Label htmlFor={label}>{label}</Label>
+      <Tooltip title="Create Epoch" content={infoTooltip}>
```
levity, line 91: fix hardcoded title here

src/ui/Checkbox/CheckBox.tsx
```diff
@@ -0,0 +1,103 @@
...
+        <InfoCircledIcon />
+      </Tooltip>
+    </Flex>
+    {error && errorText && (
+      <Text color="red" css={{ px: '$xl', fontSize: '$3' }}>
+        This is an error.
```
levity, line 97: fix or remove this placeholder text

levity (changes requested):

src/ui/Paginator/Paginator.tsx
amrro: The reason I created this one is to add support for ellipses for mobile version.

src/forms/AdminUserForm.tsx
amrro: I am using this to make the best guess about the initial value for the corresponding `Checkbox`. Consequently, if the circle's default value is to opt out it won't be checked, admin can decide then and vice versa. 

I can't get how it's going affect user(s) outside the Add/Edit User Modal.

src/ui/Link/Link.tsx
```diff
@@ -7,6 +7,12 @@ export const Link = styled('a', {
   textDecoration: 'none',
 });
 
+export const ExternalLink = styled(Link, {
+  // color: '$blue',
+  rel: 'noreferrer',
```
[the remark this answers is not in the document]
levity: You should remove `rel` and `_target` from here and add them on the tag... at that point, ExternalLink becomes identical to Link and should be removed

src/ui/Table/Table.tsx
amrro: I just copied it for now. Later, I will utilize your implementation and remove this.

src/ui/Button/Button.tsx
```diff
@@ -4,7 +4,14 @@ import { styled } from '../../stitches.config';
...
 
 export const Button = styled('button', {
+  '& img': {
+    paddingRight: '$sm',
+  },
+  '& svg': {
+    margin: '$sm',
+  },
   px: '$md',
+  py: '$sm',
```
levity, line 14:
```suggestion
```

I intentionally set zero vertical padding on buttons so that their contents don't change the button heights

src/ui/Checkbox/CheckBox.tsx
```diff
@@ -0,0 +1,102 @@
...
+  alignItems: 'center',
+  justifyContent: 'center',
+  '&:hover': { borderColor: '$secondaryText' },
+  variants: {
+    border: {
+      default: {
+        border: '2px solid $secondaryText',
+      },
+      error: {
+        border: '2px solid $red',
```
levity, line 25:
```suggestion
        border: '2px solid $alert',
```

src/ui/Checkbox/CheckBox.tsx
```diff
@@ -0,0 +1,102 @@
...
+  height: 20,
+  borderRadius: '2px',
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  '&:hover': { borderColor: '$secondaryText' },
+  variants: {
+    border: {
+      default: {
+        border: '2px solid $secondaryText',
```
levity, line 22:
```suggestion
        border: '2px solid $border',
```

src/ui/Checkbox/CheckBox.tsx
```diff
@@ -0,0 +1,102 @@
...
+  all: 'unset',
+  backgroundColor: 'transparent',
+  boxSizing: 'border-box',
+  width: 20,
+  height: 20,
+  borderRadius: '2px',
+  display: 'flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  '&:hover': { borderColor: '$secondaryText' },
```
levity, line 18:
```suggestion
  '&:hover': { borderColor: '$focusedBorder' },
```

src/ui/Button/Button.tsx
```diff
@@ -46,6 +53,14 @@ export const Button = styled('button', {
           backgroundColor: 'transparent',
         },
       },
+      textOnly: {
+        padding: '$xs',
+        backgroundColor: 'transparent',
+        color: '$primary',
```
levity, line 59:
```suggestion
        color: '$text',
```

levity:
Just a few more minor things and then I think we'll be ready to move on.

amrro:
It's ready. Thanks a lot for taking care of updating colors.
I'm going to update PR's screenshots.

amrro:
One last minor change for the separator color.

levity:
I'm seeing three issues:

1. When a user has a very long name, it messes up the size of the whole table
2. The tooltip next to the Coordinape user in the user table is broken
3. When you are editing a user, the "GIVE allotment" label should reflect the circle's custom token symbol, like the "Allow contributor to send/receive" checkboxes do

Screenshot for first two: https://capture.dropbox.com/xQ3vl4IKXVaVhKn7
Screenshot for 3rd: https://capture.dropbox.com/O3Ce6oBd9kPznXVS

amrro:
Hi @levity, 
I've tried many approaches but it doesn't work properly, especially `flex`. Here is the best I achieved 
for `<td>`, I used `grid`: 
```ts
css={{
          height: 48,
          alignItems: 'center',

          overflow: 'hidden',

          display: 'grid',
          gridTemplateColumns: 'auto minmax(0, 1fr) auto',
          width: '100%',
        }}
```

and for the `<Text>` that holds the user's name: 
```ts
  <Text
    css={{
      maxWidth: '100%',
      textOverflow: 'ellipsis',
      overflow: 'hidden',
      whiteSpace: 'nowrap',
      minWidth: 0,
    }}
  >
    {user.name}
  </Text>
```
But ellipsis doesn't show up: 


Any hints regard that?

levity:
@wingfeatherwave or @teeolendo, do you think you could spare some time to help get this across the finish line?

wingfeatherwave:
@amrro Could you trigger Vercel to deploy this as a preview so I can see the changes in browser?  Or perhaps I'm not seeing the link for it?

amrro:
I don't think I've the permission to do so. I guess @levity can help.

levity:
@wingfeatherwave for future reference, you can authorize a deployment by clicking here:

Update_Admin_Page_Attribute___support_Mobile_by_amrro_·_Pull_Request__739_·_coordinape_coordinape

wingfeatherwave:
That's ok.  I can look at it locally

wingfeatherwave:
@amrro add `display: block` to get the ellipses.  you were very close 👍

crabsinger:
This is good work, but now a large PR. I think if we could break this up into smaller chunks, it would be easier to merge it in sections and get these improvements shipped. Thanks for all your work and on-going collaboration on this @amrro - I know it's beena while and had lots of feedback.

amrro:
Thanks @wingfeatherwave for your help.

> Regarding other feedback, when I use the browser device simulation I get into modal window traps where I cannot click on any of the elements because of a couple issues, it seems, including multiple overlays and the body tag having: `pointer-events:none;` https://www.dropbox.com/s/58b9wpl6n78r5g1/Mobile-scrim-issue.mov?dl=0

It's silly mistake I made. 
I tried to make `Tooltip` (utilizing [@radix-ui/react-hover-card](https://www.radix-ui.com/docs/primitives/components/hover-card)) and supports Mobile view. It has never played nicely for many reasons, most notably: We're dealing with totally different events.

I suppose it's fixed now, please take a look.

wingfeatherwave:
@amrro modals are looking good now with respect to the clickable bug.  nice work

amrro:
@crabsinger Thanks a lot. 
Based on Levity's input, I think we are very close to merge. What a ride!

levity (dismissed):
Thanks for all your work on this!

levity:
@amrro Just one more thing, sorry for missing it earlier. The table headers don't currently reflect custom token names:

Screen Shot 2022-05-25 at 11 47 13 AM

Let's change these to use the custom token symbol (e.g. FOO):
- GIVER -> Can give FOO
- GIVE Recipient -> Can receive FOO
- GIVE sent -> FOO sent 
- GIVE received -> FOO received

After this we should be good to merge!

amrro:
> Let's change these to use the custom token symbol (e.g. FOO):
> 
>     * GIVER -> Can give FOO
> 
>     * GIVE Recipient -> Can receive FOO
> 
>     * GIVE sent -> FOO sent
> 
>     * GIVE received -> FOO received

Hey @levity. I updated table headers plus any incident of **GIVE** in Admin page like tooltips [a17e2e1](https://github.com/coordinape/coordinape/pull/739/commits/a17e2e179c915b986713bc2078a466174459cbdf).

src/pages/AdminPage/components.tsx
```diff
@@ -0,0 +1,861 @@
...
+                )}
+                {renderTooltip(
+                  <>{`Circle Member allocating ${circle.tokenName}`}</>
+                )}
+              </Table.HeaderCell>
+              <Table.HeaderCell
+                clickable
+                onClick={() => updateOrder('fixed_non_receiver')}
+              >
+                {renderLabel(`Can receive  `, order, 'fixed_non_receiver')}
```
levity, line 727:
```suggestion
                {renderLabel(`Can receive ${circle.tokenName}`, order, 'fixed_non_receiver')}
```

levity (approved):

levity:
🎉  🎉  🎉

crabsinger:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #742, opened 2022-04-06
making updated_at and start_date non-nullable on epoch

CryptoGraffe:
This makes the zeus typing a lot nicer in the FE. I verified this was ok in the prod data, except for 7 rows that are covered by the UPDATE in this migration.


zashton (approved):
looks good, hope hasura plays nice this time.

PR #743, opened 2022-04-06
make an FK from circles.protocol_id to protocols.id

CryptoGraffe:
I tested that this should work in prod by running this query:

```
select count(*) from circles c LEFT OUTER JOIN protocols p ON c.protocol_id=p.id where p.id is null;
```

and getting 0 results back

### DIFF
--- a/hasura/metadata/databases/default/tables/public_profiles.yaml
+++ b/hasura/metadata/databases/default/tables/public_profiles.yaml
@@ -79,23 +79,23 @@ update_permissions:
         _eq: X-Hasura-User-Id
     set:
       id: X-Hasura-User-Id
-  role: user
+  role: superadmin
 - permission:
     check: null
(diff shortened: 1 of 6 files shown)


hasura/metadata/databases/default/tables/public_profiles.yaml
```diff
@@ -79,23 +79,23 @@ update_permissions:
         _eq: X-Hasura-User-Id
     set:
       id: X-Hasura-User-Id
-  role: user
+  role: superadmin
```
CryptoGraffe, line 82: I don't really understand why this changed, especially the indent?

zashton: might be a previous merge that caused the indenting to go off

zashton (approved):

PR #745, opened 2022-04-06
bugfix: remove dom dependency for NodeJS ts code

topocount:
We were intiially relying on the frontend tsconfig.json file to
typecheck our api handlers. This is bad because we're expecting dom
libraries to be accessible in our backend ts code. This led to some
funky things like "shimming" node-fetch into our code when it should
work out of the box.

test plan:

1. vercel dev
2. ensure webpack is able to build
3. test some endpoints but this isn't technically necessary
4. `yarn db-seed-fresh` just to be safe.


api-lib/gql/__generated__/zeus/index.ts
```diff
@@ -22109,11 +22111,24 @@ const handleFetchResponse = (
 export const apiFetch =
   (options: fetchOptions) =>
   (query: string, variables: Record<string, any> = {}) => {
-    let fetchFunction = fetch;
```
levity, line 22112: the changes below won't get stomped by re-generation?

topocount: @levity the addition of the `--node` flag on the zeus command for server-side libgen is what generated this node-friendly code: 

https://github.com/coordinape/coordinape/blob/c688cb6452f38632f34d0e08fd4d3b78fbc3aa96/scripts/generate-zeus.sh#L47

scripts/generate-zeus.sh
```diff
@@ -44,9 +44,20 @@ function generate() {
   fi
 }
 
-generate $ADMIN_PATH -h x-hasura-admin-secret:$HASURA_GRAPHQL_ADMIN_SECRET
+generate $ADMIN_PATH --node -h x-hasura-admin-secret:$HASURA_GRAPHQL_ADMIN_SECRET
```
topocount, line 47: @levity here's the diff that generates the codegen

levity (approved):

PR #748, opened 2022-04-07
Create/update token & note allocations

### LINKED ISSUE #568: Migrate Update GIVES v2/{circle_id}/token-gifts
## Summary:
Migrate Update GIVES Laravel API Endpoint functionality to Hasura.

### Details
This task is to migrate the functionality of `<POST v2/{circle_id}/token-gifts>` to Hasura.

**Middleware:** Hasura event secret

Check sum of GIVEs allocation is less than user `starting_tokens`. Check if there is a current active epoch in the circle.
Check if the `recipient_ids` exist in the same circle as sender and not soft deleted

**Logic:**
Input array of `gifts: [{recipient_id, tokens, note}]`
Iterate through allocation If it exists check if the recipient is `non_receiver` = 0 , if either sender is a `non_giver` or recipient is a `non_receiver` set token value to 0. Search through `pending_token_gifts` and check if gift between sender recipient exists. If yes update the record if not create a new record. After processing each allocation recalculate and reset the `give_token_received` by the recipient by adding the new received token value. Recalculate `give_tokens_remaining` by sender and set to the new value.
Make sure give tokens sent does not exceed `starting_tokens`

**DB Tables:**
`users`, `pending_token_gifts`

#### Notes:

zashton:
fixes #568 
### Test Plan

1. run `yarn hasura console`
2. Choose a circle with multiple users. Make sure you have an active epoch or else create one. Purge all the `pending_gift_token` dummy data for the circle. (this may cause lag as event triggers may fire off)
3. Run this mutation to test, change the allocation array as desired
```gql
mutation MyMutation {
  updateAllocations(payload: {circle_id: <circle>, allocations: 
    [{note: "thank you", recipient_id: <id>, tokens: 42},
     {note: "no thanks", recipient_id: <id>, tokens: 15},
      {note: "", recipient_id: <id>, tokens: 43},
      {note: "hmmm", recipient_id: <id>, tokens: 11},
    ]
  
  }) {
    user {
      id
      give_token_remaining
      pending_sent_gifts{
        recipient_id
        tokens
        gift_private {
          note
        }
      }
    }
  }
}
``` 
4. After allocating check the `users` and `pending_token_gifts` tables to make sure the accounting logic is done right.
5. You should not be able to:
- Allocate to anyone outside of your circle
- Allocate to yourself
- Allocate to the same recipient more than once in the same request
- Allocate when there is no active epoch
- Allocate away more tokens than your `starting_tokens` value
6. Note that:
- Updating an allocation to empty `note` & 0 `tokens` should delete the record
- Allocating to a `non_receiver` will reduce the `token` value to 0
- Allocating as a `non_giver` will reduce the `token` values to 0
- Updating an allocation token amount should update the recipient's `give_token_received` correctly

A new DB unique constraint is added for `pending_token_gifts` for `sender_id`, `recipient_id`, `epoch_id`
to validate there is no constraint violation run
```sql
SELECT sender_id, recipient_id, epoch_id
FROM public.pending_token_gifts
group by sender_id, recipient_id, epoch_id having count(*) > 1;
```

### DIFF
--- a/api/hasura/actions/updateAllocations.ts
+++ b/api/hasura/actions/updateAllocations.ts
@@ -1,0 +1,193 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { getUsersFromUserIds } from '../../../api-lib/findUser';
+import {
+  ValueTypes,
+  pending_token_gifts_constraint,
+  pending_token_gifts_update_column,
+} from '../../../api-lib/gql/__generated__/zeus';
+import { adminClient } from '../../../api-lib/gql/adminClient';
+import { errorResponseWithStatusCode } from '../../../api-lib/HttpError';
+import { getUserFromProfileIdWithCircle } from '../../../api-lib/nominees';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+import {
+  updateAllocationsInput,
+  composeHasuraActionRequestBodyWithSession,
+  HasuraUserSessionVariables,
+} from '../../../src/lib/zod';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    input: { payload: input },
+    session_variables: sessionVariables,
+  } = composeHasuraActionRequestBodyWithSession(
+    updateAllocationsInput,
+    HasuraUserSessionVariables
+  ).parse(req.body);
+
+  const profileId = sessionVariables.hasuraProfileId;
+  const { circle_id, allocations } = input;
+
+  const user = await getUserFromProfileIdWithCircle(profileId, circle_id);
+  if (!user) {
+    return errorResponseWithStatusCode(
+      res,
+      { message: 'User does not belong to this circle' },
+      422
+    );
+  }
+  const currentEpoch = user.circle.epochs.pop();
+  if (!currentEpoch) {
+    return errorResponseWithStatusCode(
+      res,
+      { message: 'No Active Epoch' },
+      422
+    );
+  }
+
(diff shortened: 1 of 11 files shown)


topocount:
Sorry @zashton for the delay, I'll look at this first thing tomorrow.

hasura/migrations/default/1649169600501_alter_table_public_pending_token_gifts_add_unique_sender_id_recipient_id_epoch_id/down.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."pending_token_gifts" drop constraint "pending_token_gifts_sender_id_recipient_id_epoch_id_key";
```
topocount, line 1: can we add the `IF EXISTS`?
```suggestion
alter table "public"."pending_token_gifts" drop constraint if exists "pending_token_gifts_sender_id_recipient_id_epoch_id_key";
```

api/hasura/actions/updateAllocations.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  if (allocations.some(a => a.recipient_id === user.id)) {
+    return errorResponseWithStatusCode(
+      res,
+      { message: 'You cannot allocate to yourself' },
+      422
+    );
+  }
+
+  const uniqueValues = new Set(allocations.map(v => v.recipient_id));
```
topocount, line 56: clever

topocount (dismissed):
works for me

Issue #753, opened 2022-04-09
Unable to clear out / delete a note for a member


CryptoGraffe:
There is no way to delete a note for a member during allocation

1. **What is actually happening:**
I am trying to remove all the text from a note and set allocation to 0 so that a gift gets removed.

2. **What should be happening:**
I should be able to clear out a note. What If i didn't mean to leave that user a note?

3. **Steps to recreate:**
Visit allocation page, start typing a note. Try to clear it out - nope! Also happens if you reload page after saving a note, you can't remove it

4. **Screenshots/GIF/Video recording**
https://www.loom.com/share/b168671cb64c429398b52a469b26205b

PR #754, opened 2022-04-09
migrate postTokenGifts / updateAllocations to use hasura in frontend

### LINKED ISSUE #637: Integrate Hasura Update GIVES into FE
## Summary:
Integrate Hasura Update GIVES into FE.

### Details
This task is to integrate the frontend of `<Update GIVES >` to Hasura.
https://app.coordinape.com/give

**Screenshot of Page**
image

**Logic:**
-

**Related Backend Issue:**
#568

CryptoGraffe:
Fixes #637 

To test:
1. visit the allocation page for an open epoch
2. open the network tab up
3. make some allocations and write some notes
4. /graphql should be hit but not laravel
5. stuff should work ok and update ok

Note you can't clear out a note to test deletion of the gift entirely because of #753

### DIFF
--- a/src/lib/gql/mutations.ts
+++ b/src/lib/gql/mutations.ts
@@ -5,6 +5,7 @@ import {
   NominateUserParam,
   UpdateUsersParam,
   UpdateCreateEpochParam,
+  PostTokenGiftsParam,
 } from '../../types';
 
 import { $, ValueTypes } from './__generated__/zeus';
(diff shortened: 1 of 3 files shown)


zashton (approved):
Other than the existing weird note issue , works well.

src/lib/gql/mutations.ts
```diff
@@ -311,6 +312,32 @@ export async function updateTeammates(circleId: number, teammates: number[]) {
...
+) {
+  await client.mutate({
+    updateAllocations: [
+      {
+        payload: {
+          circle_id: circleId,
+          allocations: params.map(a => {
+            const withPatchedNote: Omit<typeof a, 'note'> & { note: string } = {
+              ...a,
+              note: a.note ? a.note : '',
```
META-DREAMER, line 327: If you do this, do you still need the Omit?
```suggestion
              note: a.note ? a.note.toString() : '',
```

zashton (dismissed):

src/lib/gql/mutations.ts
CryptoGraffe: @META-DREAMER I think its even simpler, like this

```
return {
  ...a,
  note: a.note ? a.note : '',
};
```

zashton (approved):

PR #762, opened 2022-04-11
migrate fetch manifest to graphql

### LINKED ISSUE #758: Migrate getManifest uber query to graphql
## Summary:
Migrate getManifest uber query to graphql

CryptoGraffe:
Fixes #758 

I broke this into multiple queries instead of one massive one. I think we can break it up further and ultimately eliminate it in some future work to come. 

I'd like some eyes on this, particularly from @zashton . I am nervous about the fetching of users in this way and the nuances between how the php code was using teammates. I don't follow it all 100%.  

Test plan:

This touches pretty much everything in the app so we just need really thorough acceptance testing. I'd like to merge it into staging for testing up there after a bit of review.

### DIFF
--- a/src/lib/gql/queries.ts
+++ b/src/lib/gql/queries.ts
@@ -306,7 +306,14 @@ export const getOtherUserProfile = async (
 export const getActiveEpochs = async (): Promise<IApiEpoch[]> => {
   const { epochs } = await client.query({
     epochs: [
-      {},
+      {
+        where: {
+          ended: { _eq: false },
+          end_date: {
+            _gt: 'now()',
+          },
+        },
+      },
       {
         id: true,
         number: true,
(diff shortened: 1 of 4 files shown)


src/lib/gql/queries.ts
```diff
@@ -299,3 +302,102 @@ export const getOtherUserProfile = async (
...
+  });
+
+  return adaptedCircles;
+};
+
+export const fetchManifest = async (
+  address: string,
+  circleId?: number
+): Promise<IApiManifest> => {
+  const profile = await getOtherUserProfile(address);
```
zashton, line 389: theres no way merge the queries into one request ?

zashton (changes requested):
The teammates doesn't seem to be displaying correctly at the moment

PR #769, opened 2022-04-12
Install Cypress

topocount:
- [x] auto-signing web3 provider working with JSON-RPC controlled
  accounts
- [x] write up summary of remaining work to enable E2E full stack tests

# local test plan
1. `docker compose --profile ci up -d --force-recreate`
2. `yarn hardhat:ganache`
3.  in a separate terminal: `./scripts/ci_local.sh`
    I added a scary user confirmation dialog so community devs can't complain if they run this
    without thinking about what they're doing to their environment first.
3. in a separate terminal: `export CI=true && yarn cy:run`
4. watch the videos or run `export CI=true && yarn cy:dev` and play around. Note, these tests are not idempotent yet and running them multiple times without re-seeding the environment will probably produce unreliable results

Cypress can also be utilized easily on the the dev stack. Just setup the stack as laid out in package.json and run `yarn cy:dev`


cypress/integration/login.ts
```diff
@@ -0,0 +1,20 @@
...
+    cy.on('window:before:load', win => {
+      const provider = new providers.Web3Provider(
+        new TestProvider('http://localhost:8545')
+      );
+      Object.defineProperty(win, 'ethereum', { value: provider });
+    });
+  });
+  it('can login', () => {
+    cy.visit('http://localhost:3000');
+    cy.get('button').click();
```
crabsinger, line 16: I wonder if we should setup some "best practices" and guides for how to interact with our interface and app. I've sometimes thought this as:

- minimize the possibility of text changes breaking UI tests. For example, this test is less focused on the exact text/design of the login button, but more on the functionality of it. In this way, I've thought in the past that it is preferable to use HTML #id or .class tags for selecting elements, rather than doing text searching and 'contains" checks. In this way, if we have a .login element, we can change the design whenever we want, but also have a nice interface for referring to this element in our tests at a "semantic" level. Does this make sense? A pattern we should use?

We might need to add more id and class elements to our UI.

I know some folks think this is a bad pattern, and would love to learn more about the downsides of this approach as well. @levity can you add your thoughts on the negatives of this id/class pattern?

@topocount where are the guides that Cypress recommends on using id/class?

topocount: @crabsinger here's the explainer. Turns out the don't recommend using id/class but rather custom attributes. https://docs.cypress.io/guides/references/best-practices#Selecting-Elements

levity: @crabsinger Here's React Testing Library's take:

> The React Testing Library is a very light-weight solution for testing React components. It provides light utility functions on top of react-dom and react-dom/test-utils, in a way that encourages better testing practices. Its primary guiding principle is:
>
> [The more your tests resemble the way your software is used, the more confidence they can give you.](https://testing-library.com/docs/guiding-principles)
>
> So rather than dealing with instances of rendered React components, your tests will work with actual DOM nodes. The utilities this library provides facilitate querying the DOM in the same way the user would. Finding form elements by their label text (just like a user would), finding links and buttons from their text (like a user would). It also exposes a recommended way to find elements by a data-testid as an "escape hatch" for elements where the text content and label do not make sense or is not practical.

levity: I like how Cypress and RTL suggest that if you do need to select a node by a non-human-readable characteristic, that it's a custom attribute (`data-cy` or `data-testid`) so you don't overload an attribute that already has a different purpose.

cypress/integration/login.ts
```diff
@@ -0,0 +1,20 @@
+import { Web3Provider } from '@ethersproject/providers';
+
+import { TestProvider } from '../util';
+
+context('Coordinape', () => {
+  before(() => {
+    cy.on('window:before:load', win => {
+      const provider = new Web3Provider(
+        new TestProvider('http://localhost:8545')
```
levity, line 9: How about using process.env.HARDHAT_GANACHE_PORT for this, in preparation for using the testchain?

cypress/integration/login.ts
```diff
@@ -0,0 +1,20 @@
...
+      );
+      Object.defineProperty(win, 'ethereum', { value: provider });
+    });
+  });
+  it('can login', () => {
+    cy.visit('http://localhost:3000');
```
levity, line 15: hmmmmm... oops, wasn't thinking about this until now. if cypress needs the app to be running, then for CI, that's another piece of the stack we'll have to setup & teardown along with the testchain & hasura.

levity: now i'm rethinking what i was saying earlier about prioritizing getting this into the CI step along with everything else... wondering if it's going to be too heavyweight.

CryptoGraffe: If only we could use the test hasura/vercel with simple seeding?

levity: We definitely can. But I'm talking about hosting the web app. Ordinary Jest tests don't need this because components are rendered in a virtual DOM

levity: that's what we'd want to do, yeah, it's just a bit of a devops boondoggle to get that all running both in dev & in a Github Action. my ideal goal here is that let you just run `yarn test:ci` in any environment, and everything would spin up with fresh data before running Jest tests and get cleaned up afterward.

i got most of the way through the Hasura side of it today, but ofc due to the auth webhook, vercel needs to run as well. i wonder if there's a way to spin up a lightweight process just for the auth hook...

CryptoGraffe:
Is this ready for another full review or are you burning down the tasks in the README and addressing this other feedback?

CryptoGraffe:
Nice job getting this all working. I can't wait to stop doing so much manual testing. Anyway its failing for me and I wonder if its a timing issue?


```
  Coordinape
    1) can login
    ✓ can opt In (263ms)
    2) can select teammates
    3) can allocate to teammates


  1 passing (13s)
  3 failing

  1) Coordinape
       can login:
     Error: The following error originated from your test code, not from Cypress.

  > PollingBlockTracker - encountered an error while attempting to update latest block:
Error: [object ProgressEvent]
    at XMLHttpRequest.errorFunc (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:101917:19)

When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.
      at PollingBlockTracker._performSync (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:63718:24)

  2) Coordinape
       can select teammates:
     Error: The following error originated from your test code, not from Cypress.

  > PollingBlockTracker - encountered an error while attempting to update latest block:
Error: [object ProgressEvent]
    at XMLHttpRequest.errorFunc (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:101917:19)

When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.
      at PollingBlockTracker._performSync (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:63718:24)

  3) Coordinape
       can allocate to teammates:
     Error: The following error originated from your test code, not from Cypress.

  > PollingBlockTracker - encountered an error while attempting to update latest block:
Error: [object ProgressEvent]
    at XMLHttpRequest.errorFunc (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:101917:19)

When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.
      at PollingBlockTracker._performSync (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:63718:24)

```

topocount:
> Nice job getting this all working. I can't wait to stop doing so much manual testing. Anyway its failing for me and I wonder if its a timing issue?
> 
> ```
>   Coordinape
>     1) can login
>     ✓ can opt In (263ms)
>     2) can select teammates
>     3) can allocate to teammates
> 
> 
>   1 passing (13s)
>   3 failing
> 
>   1) Coordinape
>        can login:
>      Error: The following error originated from your test code, not from Cypress.
> 
>   > PollingBlockTracker - encountered an error while attempting to update latest block:
> Error: [object ProgressEvent]
>     at XMLHttpRequest.errorFunc (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:101917:19)
> 
> When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.
>       at PollingBlockTracker._performSync (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:63718:24)
> 
>   2) Coordinape
>        can select teammates:
>      Error: The following error originated from your test code, not from Cypress.
> 
>   > PollingBlockTracker - encountered an error while attempting to update latest block:
> Error: [object ProgressEvent]
>     at XMLHttpRequest.errorFunc (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:101917:19)
> 
> When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.
>       at PollingBlockTracker._performSync (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:63718:24)
> 
>   3) Coordinape
>        can allocate to teammates:
>      Error: The following error originated from your test code, not from Cypress.
> 
>   > PollingBlockTracker - encountered an error while attempting to update latest block:
> Error: [object ProgressEvent]
>     at XMLHttpRequest.errorFunc (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:101917:19)
> 
> When Cypress detects uncaught errors originating from your test code it will automatically fail the current test.
>       at PollingBlockTracker._performSync (http://localhost:3000/__cypress/tests?p=cypress/integration/login.ts:63718:24)
> ```

@CryptoGraffe  This error means your environment is missing some setup, my apologies, I added a [test plan above](https://github.com/coordinape/coordinape/pull/769#issue-1200986843).

.github/workflows/CI.yml
```diff
@@ -48,10 +48,13 @@ jobs:
           key: hardhatNodeModules-${{ runner.os }}-${{ hashFiles('**/hardhat/yarn.lock') }}
 
       - run: yarn --frozen-lockfile
+      - run: yarn cypress install
```
META-DREAMER, line 51: I feel like we should create a separate GitHub action for running e2e tests vs adding more stuff to this one. That way cypress tests failing will still let people see typecheck/ linting pass separately. Also will help cut down the time CI action is running, esp as we start adding more cypress tests

ci.env
```diff
@@ -0,0 +1,25 @@
+CI=true
+DOCKER_GATEWAY_HOST=172.17.0.1
+ETHEREUM_RPC_URL=https://mainnet.infura.io/v3/275c8fc2ec294d6aac6bfc881d9a9dff
```
META-DREAMER, line 3: Who's Infura key is this? Should probably not be exposed to public in the repo since its rate limited. Putting it in GH secrets likely best option

cypress/README.md
```diff
@@ -0,0 +1,8 @@
+# Cypress Setup Path
+
+- [x] install cypress
+- [x] set up an auto-signing web3 provider that works well with the test chain
+- [x] create a static seed that covers all use cases and patterns
+- [x] set up a more advanced test script to showcase cypress' functionality
+- [x] configure a docker-compose file that doesn't clobber the dev container setup
```
META-DREAMER, line 7: This seems like a todo list of stuff you completed vs a readme on how cypress tests should be run

cypress/integration/login.ts
```diff
@@ -0,0 +1,61 @@
...
+    cy.get('[data-testid=eligibleTeammates]').within(() => {
+      cy.get('button').first().click();
+      cy.get('button').last().click();
+    });
+    cy.contains('Save Teammate List').click();
+    cy.contains('Thank your teammates by allocating them GIVE', {
+      timeout: 45000,
+    });
+  });
+  it('can allocate to teammates', () => {
```
META-DREAMER, line 32: The file name makes it seem like these tests are just for login

src/App.tsx
```diff
@@ -27,7 +27,17 @@ import { globalStyles } from './stitches.config';
...
-  const library = new Web3Provider(provider);
+  // This checks specifically whether the provider is
+  // an instance of a Web3Provider by checking the existence of this
+  // uniquely named method. Normally, we would want to use `instanceof`
+  // to check if a provider conforms to a specific class, but because
+  // some providers are injected into the window from other contexts,
+  // this check will fail, since those providers aren't derived from
+  // the same prototype tree.
+  const library =
+    typeof provider.jsonRpcFetchFunc !== 'undefined'
```
META-DREAMER, line 38: Why check this specific field?

META-DREAMER (changes requested):
Few minor nits, looks great overall!

ci.env
levity: This is ours. I check on usage once in a while and we've never come close to hitting the limits. But yeah, we can move it to a secret.

.github/workflows/CI.yml
levity: we could split out linting, but i think it's good to run all the tests together. we'll need it for a combined coverage metric

levity (changes requested):

.github/workflows/CI.yml
META-DREAMER: Yea thats what I meant, linting + typecheck in one, E2E and unit testing in another

ci.env
topocount: @levity I'll leave this to you

src/App.tsx
topocount: @META-DREAMER Our code always expects a `Web3Provider`-wrapped provider for some reason. This is the method that actually communicates onchain, so I figure since we can't go off prototype interitance, we can at least disabiguate a little bit based on the functional piece of the API that is at least a little distinctive, and failing that, named reasonably well such that we can expect an alternative implementation to be functionally equivalent.

topocount: If there was anything at least a bit more introspective available, I'd gladly use it.

CryptoGraffe (approved):
Approved! Would like a little more detail in the README.md to say you need to be running hardhat:ganache and vercel dev, and db-seed-fresh etc.

META-DREAMER (approved):

META-DREAMER (approved):
Tested, works great, good stuff!

PR #772, opened 2022-04-12
Create a Contributing Doc

topocount:
I added a lot of low-key opinionated ideas and concepts, but they're all
for the purpose of making the review process simpler, and believe these
are all practices we should be working towards.

It's worth noting that much of this is borrowed from SourceCred's
CONTRIBUTING doc, with the language softened a bit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -56,7 +56,7 @@ reviewer to be able to say, “yep, this commit is obviously correct” as
 many times in a row as are necessary for a full feature to be developed.
 
 <details>
-<summary>Why create small commits?</summary>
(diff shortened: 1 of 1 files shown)


levity (approved):
💯

crabsinger:
@topocount Awesome first draft of a contributing doc!

* We don't speak directly about PR etiquette and requirements for PRs, but rather about commits. When should a series of commits become a PR? or become multiple PRs? If we have many clean, clear, test-passing, atomic commits, can these all be in one PR even if the overall function of the changes is disparate?

crabsinger (approved):

topocount:
Great edits, @crabsinger

PR #783, opened 2022-04-15
Implement createUsers in bulk Hasura action

### LINKED ISSUE #775: Feature: Bulk Creation of users + Restore
This task should be done using hasura's action handler

The GQL mutation should accept an input of `circle_id` + an array of user inputs consisting of
`name`, `address`,  (`non_giver`, `fixed_non_receiver`, `non_receiver`, `starting_tokens`, `role` ) optional will revert to DB defaults. Only a circle admin (`role` == 1) of the specific circle can perform this action

Validate that there are no duplicates of address among the array of inputs and including the (non deleted) users within the circle.
If there is a duplicate with a soft deleted user (`deleted_at` with value) overwrite the inputs into the existing record and set the `deleted_at` value to null to restore the user (there maybe more than 1 so only overwrite 1)

if there are any duplicates throw an 422 error and indicate the specific address that has issue.

for the response return the array of user objects.

@amrro

amrro:
Resolves #775
related: #708 #508 
____ 
### Summary
It's similar to #508 in logic but in bulk action: new users will be inserted and soft-deleted users will be restored. 

### Test Plan
1. Prepare a list of new users you want to add. 
2. Soft delete single/some users and add them to the list. 
3. Try the following,
```gql
mutation {
  createUsers(payload: {circle_id: 6, users: [
    {address: "0xac0D5FD72BD52B55Fb559eee0Fb3A0638f73AdaD", name: "New User 01"},
    {address: "0x7FEA962EC07B7beA2c94672C8DA7ece097a2Af0a", name: "New User 02"}
  ]}) {
    name
    id
    fixed_non_receiver
    circle_id
    starting_tokens
  }
}
```
4. verify that new users is added and existing users is restored again, also nomination is ended & 'give_token_remaining` is updated if provided.

### DIFF
--- a/api/hasura/actions/createUsers.ts
+++ b/api/hasura/actions/createUsers.ts
@@ -1,0 +1,164 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { authCircleAdminMiddleware } from '../../../api-lib/circleAdmin';
+import { ValueTypes } from '../../../api-lib/gql/__generated__/zeus';
+import { adminClient } from '../../../api-lib/gql/adminClient';
+import { errorResponseWithStatusCode } from '../../../api-lib/HttpError';
+import {
+  composeHasuraActionRequestBody,
+  createUsersBulkSchemaInput,
+} from '../../../src/lib/zod';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    input: { payload: input },
+  } = composeHasuraActionRequestBody(createUsersBulkSchemaInput).parse(
+    req.body
+  );
+
+  const { circle_id, users } = input;
+
+  const uniqueAddresses = [...new Set(users.map(u => u.address.toLowerCase()))];
+
+  // Check if bulk contains duplicates.
+  if (uniqueAddresses.length < users.length) {
+    return errorResponseWithStatusCode(
+      res,
+      {
+        message: `Users list contains duplicate address.`,
+      },
+      422
+    );
+  }
+
+  // Load all members of the provided circle with duplicate addresses.
+  const { users: existingUsers } = await adminClient.query({
+    users: [
+      {
+        where: {
+          circle_id: { _eq: circle_id },
+          _or: uniqueAddresses.map(add => {
+            return { address: { _ilike: add } };
+          }),
+        },
+      },
+      {
+        id: true,
+        address: true,
+        deleted_at: true,
+        starting_tokens: true,
+      },
+    ],
+  });
+
+  const usersToUpdate = existingUsers.map(eu => {
+    const updatedUser = users.find(
+      u => u.address.toLowerCase() === eu.address.toLowerCase()
+    );
+    return {
+      ...eu,
+      ...updatedUser,
+      give_token_remaining: updatedUser?.starting_tokens,
+    };
+  });
+
+  // Check if a user exists without a soft deletion conflicts with the new bulk.
+  const conflict = usersToUpdate.filter(u => !u.deleted_at);
+  if (conflict.length) {
+    return errorResponseWithStatusCode(
+      res,
+      {
+        message: `Users already exist: [${conflict
+          .map(e => e.address)
+          .join(', ')}]`,
+      },
+      422
+    );
+  }
+
+  const newUsers = users
+    .filter(u => {
+      return !existingUsers.some(
+        eu => eu.address.toLowerCase() === u.address.toLowerCase()
+      );
+    })
+    .map(u => ({ ...u, circle_id }));
+
+  const updateUsersMutation = usersToUpdate.reduce((opts, user) => {
+    opts[user.address] = {
+      update_users_by_pk: [
+        {
+          pk_columns: { id: user.id },
+          _set: {
+            ...user,
+            deleted_at: null,
+          },
+        },
+        {
+          id: true,
+          circle_id: true,
+          name: true,
+          address: true,
+          non_giver: true,
+          fixed_non_receiver: true,
+          non_receiver: true,
+          starting_tokens: true,
+          role: true,
+        },
+      ],
+    };
+
+    // End any active nomination
+    opts[user.address + '_update_nominee'] = {
+      update_nominees: [
+        {
+          _set: { ended: true },
+          where: {
+            circle_id: { _eq: circle_id },
+            address: { _ilike: user.address },
+            name: { _eq: user.name },
+            ended: { _eq: false },
+          },
+        },
+        {
+          returning: {
+            id: true,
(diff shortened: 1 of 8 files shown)


amrro:
Hasura seems to insists on reformatting the existing files, I don't know why!

api/hasura/actions/createUsers.ts
```diff
@@ -0,0 +1,155 @@
...
+    const updatedUser = users.find(u => u.address === eu.address);
+    return {
+      ...eu,
+      ...updatedUser,
+      give_token_remaining: updatedUser?.starting_tokens,
+    };
+  });
+
+  // Check if a user exists without a soft deletion conflicts with the new bulk.
+  const conflict = usersToUpdate.find(u => !u.deleted_at);
```
zashton, line 61: we should compile the full list of addresses with conflicts before returning a response

hasura/metadata/actions.graphql
```diff
@@ -1,69 +1,107 @@
 type Mutation {
-  adminUpdateUser(payload: AdminUpdateUserInput!): UserResponse
+  adminUpdateUser(
+    payload: AdminUpdateUserInput!
+  ): UserResponse
 }
 
 type Mutation {
-  createCircle(payload: CreateCircleInput!): CreateCircleResponse
+  createCircle(
```
zashton, line 8: how come theres some many blank lines in this file ?

hasura/metadata/actions.graphql
```diff
@@ -1,69 +1,107 @@
...
+  createUser(
+    payload: CreateUserInput!
+  ): UserResponse
 }
 
 type Mutation {
-  deleteEpoch(payload: DeleteEpochInput!): DeleteEpochResponse
+  createUsers(
+    payload: CreateUsersInput!
+  ): [UserOutput]
```
zashton, line 34: you can just reuse : `[UserResponse]`

api/hasura/actions/createUsers.ts
```diff
@@ -0,0 +1,155 @@
...
+    ],
+    __alias: { ...updateUsersMutation },
+  });
+
+  // Get the returning values from each update-user aliases.
+  const results = usersToUpdate
+    .map(u => mutationResult[u.address].update_users_by_pk)
+    .concat(mutationResult.insert_users?.returning as []);
+
+  return res.status(200).json({ users: results });
```
zashton, line 152: the response is expecting an array but here its returning an object

zashton (changes requested):
Thanks for the quick work @amrro , highlighted some minor issues that need to be changed

hasura/metadata/actions.graphql
amrro: **Save action** in Hasura Console is reformatting this file.

api/hasura/actions/createUsers.ts
amrro: This is reasonable, but do you mean returning the whole list of `conflict` with the error message? if yes, Is there a way to include it which is easier to read in front-end.

api/hasura/actions/createUsers.ts
```diff
@@ -0,0 +1,155 @@
...
+      422
+    );
+  }
+
+  const { users: existingUsers } = await adminClient.query({
+    users: [
+      {
+        where: {
+          circle_id: { _eq: circle_id },
+          address: { _in: uniqueAddresses },
```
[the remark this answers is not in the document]
amrro: I queried the whole circle's member and later filtered them based on the lowercased address. 

I couldn't figure out how to achieve that using `_or` combined with `_ilike`. If there are any direct query that can achieve it without the needing to filter out, please point that out.

amrro:
@zashton Thanks for your thorough review. 

I resolved most of requests except couple of questions.

api/hasura/actions/createUsers.ts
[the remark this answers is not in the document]
zashton: try this
```js
const { users: circleMembers } = await adminClient.query({
    users: [
      {
        where: {
          circle_id: { _eq: circle_id },
          _or: uniqueAddresses.map(add => {
            return { address: {_ilike: add} }
          })
        },
      },
      {
        id: true,
        address: true,
        deleted_at: true,
        starting_tokens: true,
      },
    ],
  });
  ```

File: api/hasura/actions/createUsers.ts (hunk 1 above)
Comment (zashton) [reply api/hasura/actions/createUsers.ts:68]: its good enough if you just join the array with a comma seperator within the `message`

File: api/hasura/actions/createUsers.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (amrro) [reply api/hasura/actions/createUsers.ts]: Ahaaa. Got it.

File: api/hasura/actions/createUsers.ts [hunk 6]
```diff
@@ -0,0 +1,162 @@
...
+    return {
+      ...eu,
+      ...updatedUser,
+      give_token_remaining: updatedUser?.starting_tokens,
+    };
+  });
+
+  // Check if a user exists without a soft deletion conflicts with the new bulk.
+  const conflict = usersToUpdate.filter(u => !u.deleted_at);
+  if (conflict) {
```
zashton, line 67: shouldn't this be `conflict.length` ?

api/hasura/actions/createUsers.ts
```diff
@@ -0,0 +1,162 @@
...
+    return errorResponseWithStatusCode(
+      res,
+      {
+        message: `Users list contains duplicate address.`,
+      },
+      422
+    );
+  }
+
+  // Load all members of the provided circle.
```
zashton, line 34: // Load all members of the provided circle with duplicate addresses

zashton (changes requested):
just would need to update the if condition otherwise the mutation can't run

zashton (approved):

api/hasura/actions/createUsers.ts
```diff
@@ -0,0 +1,164 @@
...
+          },
+        },
+      ],
+    };
+    return opts;
+  }, {} as { [aliasKey: string]: ValueTypes['mutation_root'] });
+
+  // Update the state after all validations have passed
+  const mutationResult = await adminClient.mutate({
+    insert_users: [
```
zashton, line 135: Sorry, just one more new bug we discovered regarding adding of users, we need to insert the profile as well due to fk constraints in the same mutation
you can refer to it here
https://github.com/coordinape/coordinape/pull/805

zashton (changes requested):
all good now just need to rebase an i can merge it

amrro:
@zashton It has been rebased on `main`.

zashton (approved):

PR #791, opened 2022-04-18
Implement disable_cron script

topocount:
This script is intended for testing and development environments. It
disables cron tasks in a local hasura docker container without modifying
the repo metadata, which makes it sort of fix-and-forget. This has pros
and cons, but I don't think most devs want to be bothered with what's
happening with cron triggers when working locally anyhow.

This script has been inserted into package.json and the seed script as a
convenience.

test plan:

`yarn db-disable-cron`

@levity please ensure this works on macs


levity (approved):
LGTM. Noticing we have some kebab case & some snake case in script names though, would be good to be consistent

PR #797, opened 2022-04-19
Add  fk from users.address->profile.address

### LINKED ISSUE #796: make users.profile_id a foreign key to the profiles table
Without a Foreign Key, this relationship results in optional `profile` relationship in the generated zeus types from hasura. 

There are currently 32 rows in prod that violate this relationship. They are all in cases where the user has been deleted. I'm not sure how the profile itself is getting deleted - when it happens we should cascade delete the users row. 

Here is the investigative query I used to find out the problematic rows:

```
select u.deleted_at from users u left outer join profiles p on p.address=u.address where p.id is null and u.deleted_at is not null;
```

CryptoGraffe:
Fixes #796 

See the issue for context. This deletes some rows and adds an FK that makes `user.profile` non-optional.

### DIFF
--- a/api-lib/gql/__generated__/zeus/index.ts
+++ b/api-lib/gql/__generated__/zeus/index.ts
@@ -13377,7 +13377,7 @@ columns and relationships of "distributions" */
     /** An aggregate relationship */
     pending_sent_gifts_aggregate: ModelTypes['pending_token_gifts_aggregate'];
     /** An object relationship */
-    profile?: ModelTypes['profiles'];
+    profile: ModelTypes['profiles'];
     /** An array relationship */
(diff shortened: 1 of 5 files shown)


zashton (approved):

topocount (changes requested):
@CryptoGraffe can you share your verification that this constraint will apply successfully in prod

topocount (approved):
@CryptoGraffe I don't see the need for the deleted_at filter but the query returns 0 without it anyway

CryptoGraffe:
I was using that query because I was going to delete the matching rows -
but instead zashton connected them to new profiles rows

PR #798, opened 2022-04-19
Improve performance of auth handler + updateUser mutation

META-DREAMER:
Updated GraphQL Zeus to v5.0 which allows using GraphQL Directives (needed so we can use [Hasura's query caching feature](https://hasura.io/docs/latest/graphql/cloud/response-caching/)). Had to fork it to fix some bugs, made a PR upstream here: https://github.com/graphql-editor/graphql-zeus/pull/270, once that gets released we can go back to using the original NPM package.

The auth handler was the main bottleneck in mutation execution, this PR reduces the amount of requests made back and forth between Vercel and Hasura + enables caching of the Hasura query to fetch the auth token. Also simplified the updateUser mutation to query only the fields it needs.

Refactored some code for new version of GraphQL Zeus to handle aliases better and add typesafety to variables used in queries/mutations.


aexol:
Done in 5.0.1

CryptoGraffe:
Your two PRs are in 5.0.1 , can you move to that and also rebase this? https://github.com/graphql-editor/graphql-zeus/releases/tag/5.0.1

CryptoGraffe (approved):
I like all the aliasing for the derived fields, thats clever. I approve this but it would be nice for someone else to give it eyes, maybe @zashton or @topocount

api/hasura/actions/updateUser.ts
```diff
@@ -25,7 +24,27 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
+        },
+      },
+      {
+        id: true,
+        fixed_non_receiver: true,
+        give_token_received: true,
+      },
+    ],
+  });
+
```
topocount, line 47: why write out the query without deleting the query from the library?

topocount (approved):
will re-approve post-rebase

api/hasura/auth.ts
```diff
@@ -21,31 +21,33 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
...
+            tokenable_id: true,
+            profile: {
+              admin_view: true,
+              address: true,
+            },
+          },
+        ],
+        // cache query with ttl of 30s
+      },
+      { operationName: 'GetAuthToken @cached(ttl: 30)' }
```
zashton, line 44: Nice . Caching in the auth layer is going to reduce alot of unnecessary overhead for every request , i wonder if we should push the time to the max 300s and force a cache refresh in a `delete_personal_access_tokens` event trigger

zashton (approved):

CryptoGraffe:
@META-DREAMER do you have time to rebase this and get it in?

levity:
If he doesn’t, I can take a look today, since it’s my changes to the
history page that are the bulk of the conflicts

levity (dismissed):

CryptoGraffe (dismissed):
This looks ok to me but it could use some more testing, maybe @zashton you have a checklist  you could throw at it?

api/hasura/actions/updateUser.ts
META-DREAMER: `getUserAndCurrentEpoch` is used in other places. Just was overkill here.

zashton:
@CryptoGraffe @META-DREAMER 
### Test Plan

1. Make sure the createsUsers mutation is working.

```gql
mutation MyMutation {
 createUsers(payload: {circle_id: 7, users: [
    {address: "0x2d407ddb06311396fe14d4b49da5f0471447d45c", name: "New User 1"},
    {address: "0x2f50d538606fa9edd2b11e2446beb18c9d5846bb", name: "New User 2"}
  ]}) {
     UserResponse {
      address
      name
    }
  }}

  ```
  ✅ 
2. Create an epoch , allocate some gifts. Set the telegram_id / discord_webhook in your circle. Update the dailyUpdate cron job to send every minute. Make sure the daily updates are sent. ⛔ 

3.  Set the end date of the epoch to end immediately .Make sure the end epoch cron is running without errors and gifts are transferred to the `token_gift` table and user gives received are reset. ⛔ 

5. Test image uploading in your profile via the UI ✅ 


for the crons i'm getting this errors. Certain crons like start epoch are working
```
A new Collier and Sons/Games epoch is active!
7 users will be participating and the duration of the epoch will be:
May 7, 2022, 10:03 PM UTC to May 25, 2022, 10:03 PM UTC
```

```json
{
  errors: [
    {
      extensions: [Object],
      message: 'expected an enum value for type "pending_token_gifts_select_column", but found a string'
    }
  ]
}
{
  errors: [ { extensions: [Object], message: 'not a valid graphql query' } ]
}
```

META-DREAMER:
@zashton  Thanks for catching these! Fixed in latest commit, if you could test again to verify that would be awesome.

zashton:
> @zashton Thanks for catching these! Fixed in latest commit, if you could test again to verify that would be awesome.

I've tested and the errors are gone now, nice work with this.

zashton (approved):

PR #803, opened 2022-04-20
Rename repo root scripts to use snake case

topocount:
Snake_case is the common convention for shell scripts. ts scripts have
also been renamed to use snake_case to denote that they are scripts, and
not production code.

I'm unsure if we want to apply this convention in the hardhat subproject
at this point, and will leave that as-is for now.

Additionally, I also added the backend files to the typecheck yarn
script. This was an initial oversight, but has since been corrected. The
unused tsconfig.esrun.json has been deleted as well.

test plan:
`yarn db-seed-fresh`
`yarn generate`
`yarn clean`


topocount:
@levity it looks like we'll need to update the vercel build command before merging: https://vercel.com/coordinape/coordinape/settings

to 
```sh
./scripts/link_hardhat.sh && yarn build
```

levity (approved):

topocount:
@levity had to rebase. nothing has changed since you approved

levity (approved):

PR #805, opened 2022-04-20
Make sure profile exists before user is created to keep FK happy

### LINKED ISSUE #804: createUser is failing for brand new addresses
profile is created using an eventTrigger rather than during user creation, which has an FK violation.

CryptoGraffe:
Fixes #804 

To Test (locally, i dont think we can test this on test vercel cuz it has hasura dependencies):

1. Try to add users with existing and non-existing address to a circle you are an admin of

### DIFF
--- a/api/hasura/actions/createUser.ts
+++ b/api/hasura/actions/createUser.ts
@@ -42,6 +42,25 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     });
   }
 
+  // make sure a profile exists first, for FK purposes
+
+  const { profiles } = await adminClient.query({
+    profiles: [{ where: { address: { _ilike: input.address } } }, { id: true }],
+  });
+
+  const profile = profiles.pop();
+
+  if (!profile) {
+    await adminClient.mutate({
+      insert_profiles_one: [
+        { object: { address: input.address } },
+        { id: true },
+      ],
+    });
+  }
+
+  // create the user
+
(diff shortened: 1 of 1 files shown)


api/hasura/actions/createUser.ts
```diff
@@ -42,6 +42,25 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     });
   }
 
+  // make sure a profile exists first, for FK purposes
+
+  const { profiles } = await adminClient.query({
```
zashton, line 47: you should do an upsert here to save 1 less query

```gql
const { insert_profiles_one } = await adminClient.mutate({
        // Create a profile if none exists yet
        insert_profiles_one: [
          {
            object: { address: input.address },

            // This clause allows gql to catch the conflict and do nothing
            // hasura calls this an "upsert"
            on_conflict: {
              constraint: profiles_constraint.profiles_address_key,
              // Don't update the entry at all if a profile exists
              // Don't want to touch the timestamp if we aren't actually
              // modifying anything
              update_columns: [],
            },
          },
          {
            id: true,
          },
        ],
      });
```

zashton (changes requested):

api/hasura/actions/createUser.ts
CryptoGraffe: @zashton  This doesnt handle the ilike case

zashton (approved):

zashton (approved):

PR #814, opened 2022-04-22
Fix/741/epoch creating loading

### LINKED ISSUE #741: Create Epoch form doesn't immediately go into loading state upon clicking save
When I click save (which should probably be "Create") on the Create Epoch modal, the loading state does not immediately initiate. There is some lag, and then it initiates. This allows for double clicks of the save button. I thought it wasn't working and clicked again, which eventually ended up in an error state. 

The loading indicator should be immediately present and the save button should be disabled upon clicking "Save"

3. **Steps to recreate:**
Visit the admin page for a circle, click create epoch, click save. 

https://www.loom.com/share/0970c5d194004a689788b5aefd14e483

sshmm:
## Summary
fix #741 by enable loading Modal
## Describtion
-Enable loading Modal on Create or Update Epoch to prevent the user from clicking the save button multiple times
## Test Plan
1- Go to Admin Page
2- Click Create Epoch or edit one of the existing Epochs
3- Update the Modal with correct values and click Save
Expected:
Loading Spinner should appear and prevent further clicking on Save button

### DIFF
--- a/src/hooks/useApiAdminCircle.ts
+++ b/src/hooks/useApiAdminCircle.ts
@@ -39,7 +39,7 @@ export const useApiAdminCircle = (circleId: number) => {
       await fetchManifest();
     },
     [circleId],
-    { hideLoading: true }
+    { hideLoading: false }
   );
 
   const updateEpoch = useRecoilLoadCatch(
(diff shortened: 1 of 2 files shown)


sshmm:
@levity the first commit is using disable saving button approach upon submitting
The second one I reverted those changes and enabled loading spinner
Both commits will achieve the desired output which is disallowing multiple clicks but it seems from the bug description that the second behavior is hat is needed.

levity (approved):
Note that you could have omitted the options arg entirely, but this is fine too 👍

PR #825, opened 2022-04-24
fixes #824 by cleaning up UserResponse

### LINKED ISSUE #824: updateUser triggers error when saving - missing circle_id field
When saving your epoch update with updateUser, changes are applied to the database but an error is thrown back to user. This was caused by a schema change in #783 that had wider ranging impacts than the developer intended. 

https://user-images.githubusercontent.com/95149165/164993023-f25aef5d-3891-4fbf-92a3-1fa33abe3f38.mov

CryptoGraffe:


topocount (approved):
Tested the epoch status page again, an allocation, and an epoch creation for good measure. I really want to get these happy paths and vouching covered by e2e tests some point soonish so we can rest easier and not labor so hard on these reviews.

Issue #692, opened 2022-03-25
Improve seed script


levity:
### Motivation

The current behavior of the [db seed script](https://github.com/coordinape/coordinape/blob/main/scripts/db-seed.ts) is sub-optimal, because it creates some data that is semantically incorrect, & doesn't cover some useful test cases.

### Specification / Wishlist

Devs, please add what you want to see below.

- [x] Create some token gifts to & from the dev's address, both within and prior to the last 24 hours
- [x] Make circle names unique
- [x] Create some users with valid avatar images
- [x] Create a circle where you're not an admin
- [x] Create a circle in your org where you're not a member
- [x] Create a circle that just began an epoch
- [x] Create a circle that is about to end an epoch (within 24 hours of ending)
- [x] Create circles that exited the epoch period but haven't `ended` yet and has pending token gifts and bios for users. One will have `auto_opt_out` enabled, the other won't.
- [x] Create a finished epoch with gifts that were given and received by multiple users
- [x] Bug fix: don't create nominees for circles with vouching disabled
- [x] Bug fix: don't create pending_token_gifts for epochs that haven't started yet

PR #831, opened 2022-04-26
Write new db seed script

topocount:
This new seed script checks every box in https://github.com/coordinape/coordinape/issues/692 and is meant to be at
least a little bit more maintainable and extensible than its
predecessor.

Nominees were omitted but can be added if desired. I figured the
workflow for testing vouching and nominees is simple enough that it's
not worth setting something up mid-stream.

Avatars are now generated as a part of the seed after running `yarn docker:clean`

All accounts used in this seed can be accessed by utilizing the seed
`test test test test test test test test test test test junk` in one's
wallet.

test plan:

- yarn docker:clean
- yarn docker:start
- yarn db-seed-fresh

Observe that all details in https://github.com/coordinape/coordinape/issues/692 are taken care of


META-DREAMER:
FWIW I like using https://www.getsynth.com/ for stuff like this, check it out and see if it tickles your fancy

Also there's this, not sure if its helpful here but if we have zod schemas for stuff would be nice to reuse for generating mock data as well: https://github.com/anatine/zod-plugins/tree/main/libs/zod-mock

scripts/db_seed_test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+async function createCircleWithPendingGifts() {
+  const result = await insertMemberships(
+    getMembershipInput({ protocolInput: { name: 'Open Epoch With Gifts' } }, {})
+  );
+  const circleId = result[0].circle_id;
+  const epochId = await makeEpoch(
+    circleId,
+    DateTime.now().minus({ days: 6 }),
+    DateTime.now().plus({ hours: 6 }),
```
levity, line 132:
```suggestion
    DateTime.now().plus({ days: 14 }),
```

levity (changes requested):
I don't want to context-switch rn to test the output of this, so here's some feedback just on the code. Happy to just merge it once the issues below are addressed and then tweak it if necessary later on.

scripts/db_seed_test.ts
topocount: @levity this epoch is supposed to be close to ending.

scripts/util/seed.ts
```diff
@@ -0,0 +1,323 @@
...
+  input: MemberInput,
+  epochId: number,
+  pending = true,
+  userIdx = 0,
+  memberSlice = [1, 6]
+) {
+  const amount = 5;
+  const user = input[userIdx];
+  const members = input.slice(...memberSlice);
+  for (let i = 0; i < members.length; i++) {
```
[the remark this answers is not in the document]
topocount: @levity I wanted to use `for...of` syntax here but I need to use the index to generate timestamps.

scripts/db_seed_test.ts
levity: ah, i see. i was thinking of a different use case, for generating a current epoch that sticks around for a while for use in development. never mind!

levity (approved):

Issue #799, opened 2022-04-19
Clickable Circles Profile Page


ivyquinzel:
### Summary
Users are having trouble navigating to different circles. The profile page is an opportunity to create a workaround. 

### Specification
Make the circles displayed on the profile clickthrough the referenced circles. 

### Acceptance Criteria
Scenario: A user is trying to find the other circles they are in
When: They click on their profile image
Then: Navigate to My Profile
Given: They are in multiple circles
Then: The user clicks a "Core Coordinape" under "My Circle."
Given: The user is a member of "Core Coordinape"
Then: The user will land on the history page of the circle "Core Coordinape."

URL: https://app.coordinape.com/profile/me

<img width="726" alt="Screen Shot 2022-04-22 at 3 49 12 PM" src="https://user-images.githubusercontent.com/83605543/164814530-2564b741-4ec6-4ab3-ba4a-84824e4367cd.png">

### Stakeholders
@smombartz @AlexDistill

levity:
Implementation note: these show up for all profiles, not just your own, and include circles that the viewer is not a member of. So only circles that the viewer is a member of should be clickable.

@ivyquinzel btw, I added a "My Circles" link to the avatar menu that just does the same thing as clicking "Circles" in the navbar.

reeserj:
@spydermonkee could we assign this any dev or do we need something additional?

spydermonkee:
@reeserj - This is definitely a good Community Dev issue!

akkikumar72:
Please assign this to me. I can work on it

reeserj:
> Please assign this to me. I can work on it

You got it! @akkikumar72

akkikumar72:
@ivyquinzel - I  saw that we have a `My circles` page, which does the same thing. Either we can make this page a bit better or we can use the existing profile page to navigate b/w circles

![Mycircles](https://user-images.githubusercontent.com/4746256/164979295-47b2cc68-d381-4167-bcf7-bd2b80fa3e9b.png)

reeserj:
This is needed because if you're in an organization with circles you're not
in this will display them.

akkikumar72:
We can navigate b/w circles here as well. 

Trying to understand the difference b/w the `My Circles` menu page and the "My Circles" section on the profile page.

reeserj:
Menu page will have all circles in the org, the profile page with just have circles you're in
________________________________

PR #800, opened 2022-04-20
Allow viewing a submitted distribution

levity:
Closes #651 

- [x] performance improvement: don't lookup previous distribution data until form is submitted

PR #832, opened 2022-04-26
799 clickable circle profile page

akkikumar72:
This PR is fixing https://github.com/coordinape/coordinape/issues/799 issue.


src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -328,15 +352,19 @@ const ProfilePageContent = ({
               u =>
                 u.circle && (
                   <div key={u.id} className={classes.circle}>
-                    <Avatar
-                      alt={u.circle.name}
-                      src={getCircleAvatar({
+                    <NewApeAvatar
```
crabsinger, line 355: Hmm, I wonder if we could have a more descriptive name than `NewApeAvatar`

akkikumar72: Actually, I used the existing `NewApeAvatar` 

It's Intended to replace ApeAvatar. Maybe once we get rid of ApeAvatar, then it will be named accordingly

levity: Yeah... Note that @akkikumar72 is just reusing an existing component.

The history here is that "Avatar" itself is a Material-UI component, so when exrhizo added functionality he called the new component ApeAvatar, but it had some weird workarounds for data inconsistency. 

So @teeolendo started a new version of that, and called it NewApeAvatar, but it never got finished. 

Now there's also a second reworking called just "Avatar" again by @amrro in #739 but there's still outstanding feedback on that.

One way forward would be for @akkikumar72 to pull in the new Avatar component from #739, fix the one weird thing about it, and use that in ProfilePage. We could later go through the code and replace instances of ApeAvatar and NewApeAvatar with just the new Avatar, and remove the two old components entirely.

crabsinger: Oh, my bad @akkikumar72 didn't realize this historical context.

akkikumar72: @crabsinger np, I was also not fully sure of its historical context. Thanks, @levity for letting us know.

I have added a small comment for https://github.com/coordinape/coordinape/pull/739, as it's a huge PR and I don't want to pull all of it without it getting approved. 

So, either I will wait, or if it takes time then I will cherry-pick avatar change only.

levity:
Yes, feel free to cherry-pick

akkikumar72:
Added the new avatar. As a fallback, showing the text of the avatar if the name is present

levity:
> Added the new avatar. As a fallback, showing the text of the avatar if the name is present

If you look at the implementation of `getAvatarPathWithFallback`, you'll see it uses an API to generate an image with the initials of the name provided, so this is already a sufficient fallback. I don't think it makes sense to provide text as fallback for an image, because the text will be a variable size and throw off the visual layout.

levity:
great!

src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -1,26 +1,29 @@
 import React from 'react';
 
 import { transparentize } from 'polished';
+import { useNavigate } from 'react-router';
 import { useParams } from 'react-router-dom';
 
-import { makeStyles, Avatar, Button } from '@material-ui/core';
+import { makeStyles, Button } from '@material-ui/core';
 
+import { Avatar } from '../../ui/Avatar/Avatar';
```
levity, line 9: can you add `export * from './Avatar/Avatar'` to `src/ui/index`? then this import can become
```
import { Avatar } from 'ui';
```

src/utils/domain.ts
```diff
@@ -56,6 +56,9 @@ export const getAvatarPathWithFallback = (avatar?: string, name?: string) => {
   return avatar ? getAvatarPath(avatar) : placeholder;
 };
 
+export const getInitialFromName = (name: string) =>
```
levity, line 59: when the name is two words, the initials should be the first letters of each word.

src/ui/Avatar/Avatar.tsx
```diff
@@ -0,0 +1,69 @@
...
+  height: 60,
+  margin: '$sm',
+  display: 'inline-flex',
+  alignItems: 'center',
+  justifyContent: 'center',
+  verticalAlign: 'middle',
+  overflow: 'hidden',
+  userSelect: 'none',
+  borderRadius: '100%',
+  backgroundColor: 'Black',
```
levity, line 19:
```suggestion
  backgroundColor: '$black',
```

src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -70,6 +73,17 @@ const useStyles = makeStyles(theme => ({
     width: 143,
     height: 143,
   },
+  myCircleAvatar: {
```
levity, line 76: remove this now that it's not used anymore

levity (changes requested):
This is looking good. Once you address the issues below, and remove the changes to the NewApeAvatar files, it should be ready to merge.

If you want, you can also change the other places in the code that use NewApeAvatar to use this instead, and then remove the NewApeAvatar files completely.

akkikumar72:
@levity I have done changes as per your PR comments. I was thinking to remove existing material `avatar` and `NewApeAvatar ` files and replace them with this `avatar` in a separate PR.
 
If that's fine, I will create a separate PR quickly after we merge this PR.

levity:
> @levity I have done changes as per your PR comments. I was thinking to remove existing material `avatar` and `NewApeAvatar ` files and replace them with this `avatar` in a separate PR.
> 
>  
> 
> If that's fine, I will create a separate PR quickly after we merge this PR.

Yep, that's fine

src/pages/ProfilePage/ProfilePage.tsx
```diff
@@ -241,6 +245,15 @@ const ProfilePageContent = ({
 
   const setEditProfileOpen = useSetEditProfileOpen();
   const { updateBackground } = useApiWithProfile();
+  const navigate = useNavigate();
+  const { selectAndFetchCircle } = useApiBase();
+
+  const goToCircleHistory = (id: number, path: string) => {
+    selectAndFetchCircle(id).then(() => {
```
levity, line 252: You actually want to use selectCircle here instead (I made the same mistake myself and just fixed it in #800)

levity (changes requested):

src/pages/ProfilePage/ProfilePage.tsx
akkikumar72: good to know, fixed now

akkikumar72:
@levity Seems like I have fixed all the comments. Can we merge this PR then?

levity:
> @levity Seems like I have fixed all the comments. Can we merge this PR then?

What I mean about the initials is that if the name is "Peter", the initials should be "PE", but if the name is "Peter White" they should be "PW".

akkikumar72:
@levity - thanks for the correction. I have fixed it and added a test file to make it more robust.

levity (approved):

levity:
Thanks for the fix and especially for adding a test @akkikumar72!

PR #834, opened 2022-04-26
Fix for duplicate address issue

### LINKED ISSUE #833: Users with same address in the same circle
**Summary:**

 Users with the same address that are not soft deleted are appearing in the same circle

There shouldn't be duplicated address appearing in the same circle

3. **Steps to recreate:**
a. Create a nominee
b. Create a new user with the same address but different name
c. Give sufficient vouches to that nominee
d. Nominee gets created into a user and two users with the same address are in the same circle

zashton:
Fixes #833

### Test Plan
1. Create a nominee
2. Create a user with the same address but different name (or update a user to the same address)
3. Make sure nomination has ended and you cannot vouch for the user anymore
4. Ensure `checkNominee` cron job is running
5. Check for DB conflicts run:

```sql 
SELECT address, circle_id, deleted_at
FROM public.users
group by address, circle_id, deleted_at having  count(*) > 1;
```

### DIFF
--- a/api/hasura/actions/adminUpdateUser.ts
+++ b/api/hasura/actions/adminUpdateUser.ts
@@ -91,6 +91,21 @@ async function handler(req: VercelRequest, res: VercelResponse) {
         },
       },
     ],
+    update_nominees: [
+      {
+        _set: { ended: true },
+        where: {
+          circle_id: { _eq: circle_id },
+          address: { _ilike: input.new_address },
+          ended: { _eq: false },
+        },
+      },
+      {
+        returning: {
+          id: true,
+        },
+      },
+    ],
   });
 
   const returnResult = mutationResult.update_users?.returning.pop();
--- a/hasura/migrations/default/1651000685281_alter_table_public_users_add_unique_address_circle_id_deleted_at/up.sql
+++ b/hasura/migrations/default/1651000685281_alter_table_public_users_add_unique_address_circle_id_deleted_at/up.sql
@@ -1,0 +1,1 @@
+alter table "public"."users" add constraint "users_address_circle_id_deleted_at_key" unique ("address", "circle_id", "deleted_at");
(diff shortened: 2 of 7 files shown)


zashton:
> can you also create a database migration to add a uniqueness constraint for `(address, circle_id, deleted_at)`?

yeap, def could add that

zashton:
just added them, to check for db conflicts

```sql SELECT address, circle_id, deleted_at
FROM public.users
group by address, circle_id, deleted_at having  count(*) > 1;
```

hasura/migrations/default/1651000685281_alter_table_public_users_add_unique_address_circle_id_deleted_at/up.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."users" add constraint "users_address_circle_id_deleted_at_key" unique ("address", "circle_id", "deleted_at");
```
topocount, line 1: I tried to implement logic in createUser that "undeletes" soft-deleted users, and therefore this constraint should just require the `address` and `circle_id` columns. Is the logic not airtight enough to do that, or is existing production data preventing this? I think having a bunch of deleted users in a circle with the same address will make for some messy situations.

zashton: yeah prior to your undelete feature there were cases of creating a new user after soft deleting one with the same address.The soft deleted users can't be hard deleted because of token_gifts fk constraints

api/hasura/actions/adminUpdateUser.ts
```diff
@@ -91,6 +91,21 @@ async function handler(req: VercelRequest, res: VercelResponse) {
         },
       },
     ],
+    update_nominees: [
```
topocount, line 94: nice

topocount (approved):
send it!

Issue #835, opened 2022-04-26
Bug Template


amrro:
**Summary:**
There is a bug in production Date and time is not showing properly inside the GIVE receive info summary.

1. **What is actually happening:**
It's showing that following message instead of date: **Invalid DateTime**

3. **What should be happening:**
Shows the correct time of of the GIVE receiving.  

5. **Steps to recreate:**
- In home page of app.coordinape.com.
- click on the GIVE count. 
- view the history of receiving GIVEs. 

7. **Screenshots/GIF/Video recording**
<img width="356" alt="Screen Shot 2022-04-26 at 2 52 44 PM" src="https://user-images.githubusercontent.com/12991700/165304067-28fab28a-ab66-4f6c-9fab-2e32a96565a2.png">

PR #836, opened 2022-04-26
Fix `DateTime` format in receive Info summary

amrro:
closes: #835
____
The received dates from the database values are ISO 8601 format
example: `2022-04-26T12:13:25.976896`.

when using `DateTime.fromSQL()` the luxon library gives this explanation: 
```
  reason: 'unparsable',
  explanation: `the input "2022-04-26T12:13:25.976896" can't be parsed as SQL
```


levity:
Thanks @amrro!

PR #839, opened 2022-04-27
schema changes for #800

levity:
pulling this out to make the main PR easier to review

checked the migration on prod:

```
> select count(*) from epoches where circle_id not in (select id from circles);
0
```


hasura/metadata/databases/default/tables/public_vaults.yaml
```diff
@@ -23,9 +23,12 @@ insert_permissions:
...
           users:
-            profile:
-              id:
-                _eq: X-Hasura-User-Id
+            _and:
+            - deleted_at:
+                _is_null: true
+            - profile:
+                id:
+                  _eq: X-Hasura-User-Id
```
zashton, line 31: not too sure but would it expose any security issue if any random member of the circle use the api directly to create a vault record ?

levity: good point! we should really disable this and write a custom mutation.

levity: @teeolendo this is an interesting one... i think we can still pass the vault address from the frontend when creating a vault, but the server will need to validate it. otherwise a circle member could create a "vault" with the address of a malicious contract

zashton: if there is a vault registry i guess you can pass some sort of id or index instead of address to prevent address spoofing

levity: This issue predates the changes in this PR, so I'm not going to consider it a merge blocker, but I'll make a task for it

topocount (approved):

PR #840, opened 2022-04-27
Make edge thickness a factor of node diameter

### LINKED ISSUE #733: Map line weights need a max thickness
**Summary:**

2. When a user allocated significantly more than other users the line weights look bad in the map tab

4. There should be a max thickness for the lines

5. **Steps to recreate:**
6. Set yourself 1000 give and allocate to a user, 

7. **Screenshots/GIF/Video recording**
8. 
Screenshot_20220402-001722

sshmm:
## Summary
fixes #733 
Make edge thickness a factor of node diameter
add `minZoom` to `ForceGraph2D`
remove unnecessary `nodeRelSize` 

## Description
Accoriding to `force-graph` `linkWidth` [docs](https://github.com/vasturiano/force-graph) : _link widths remain visually constant through various zoom levels, where as node sizes scale relatively._
so it is not possible to make the link width resisable as the nodes diameter

My suggested solution made link width a factor of 80% of the node diameter multiplied by 3 and minZoom is set to 3 so that users can't zoom out until link width is greater than circle width.

### Preconditions:
1- Active Epoch with multiple opt-in users
2- Set yourself 1000  gives
### Test scenario
1- allocate to a user 500 and distribute the rest to the other users
### Expected:
Node edges width should be visible correctly and max width shouldn't exceed the circles diameters
image

### DIFF
--- a/src/pages/AssetMapPage/AMForceGraph.tsx
+++ b/src/pages/AssetMapPage/AMForceGraph.tsx
@@ -46,7 +46,7 @@ const COLOR_LINK_DIM = '#00000008';
 
 const MIN_ZOOM = 3;
 
-const edgeWidthScaler = (f: number) => f * MIN_ZOOM + 1;
+const edgeWidthScaler = (f: number) => f * MIN_ZOOM + MIN_ZOOM;
 // const nodeSizeScaler = (f: number) => NODE_R + f * 8;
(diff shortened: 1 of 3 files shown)


sshmm:
re-basing to latest commit in coordinape/main

levity (approved):

PR #841, opened 2022-04-27
Fixes the refunding issue #829

### LINKED ISSUE #817: AssertionError: panic: user must exist
Sentry Issue: [APP-RE](https://sentry.io/organizations/coordinape/issues/3187150198/?referrer=github_integration)

```
AssertionError: panic: user must exist
  File "/vercel/path3/api-lib/event_triggers/refundPendingGift.ts", line 30, in Object.handler [as refundPendingGift]
```

### LINKED ISSUE #829: Deleting users doesn't remove allocations to user from pending_sent_gifts
**Summary:**

Given:
 you have a user X in a circle, and an epoch starts, and that user X receives GIVE from user A, and then the admin deletes the user X, the `pending_token_gifts` from A -> X record still exists, and creates an error when rendering the page.

The UI errors with a "e.user is undefined" error.

All related records should be deleted or deactived.

3. **Steps to recreate:**

4. **Steps to fix:**

Right now, the delete user mutation: https://github.com/coordinape/coordinape/blob/main/api/hasura/actions/deleteUser.ts#L49

only deletes teammates, but does not delete records from the `pending_token_gifts` table where `recipient_id` is the deleted user.

zashton:
Fixes #829 
Fixes #817 

issue was `getUserByIdAndCurrentEpoch` was not fetching deleted users
changed to `getUserByIdAndCurrentEpoch` so it would target the right user regardless if its deleted or not.

this issue is leaving quite a bit of bad data that i had to manually fix. you won't be able to access your page if the deleted user sent to you or you received gives from him.

### DIFF
--- a/api-lib/gql/queries.ts
+++ b/api-lib/gql/queries.ts
@@ -148,6 +148,80 @@ export async function getUserAndCurrentEpoch(
   return user;
 }
 
+export async function getUserByIdAndCurrentEpoch(
+  id: number,
+  circleId: number
+): Promise<typeof user | undefined> {
+  const {
+    users: [user],
+  } = await adminClient.query({
+    users: [
+      {
+        limit: 1,
+        where: {
+          id: { _eq: id },
+          circle_id: { _eq: circleId },
+        },
+      },
+      {
+        id: true,
+        fixed_non_receiver: true,
+        non_giver: true,
+        non_receiver: true,
+        starting_tokens: true,
+        give_token_received: true,
+        give_token_remaining: true,
+        pending_sent_gifts: [
+          // the join filters down to only gifts to the user
+          {},
+          {
+            id: true,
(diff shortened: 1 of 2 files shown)


zashton:
this issue is creating quite a bit of bad data that i had to manually fix. you won't be able to access your page if the deleted user sent or you received gives from him.

api-lib/gql/queries.ts
```diff
@@ -148,6 +148,80 @@ export async function getUserAndCurrentEpoch(
...
+              id: true,
+            },
+          ],
+        },
+      },
+    ],
+  });
+  return user;
+}
+
```
topocount, line 224: I wish we could make this more DRY, but I understand this is a significant break, and the refactor can wait for another PR

topocount (approved):

PR #847, opened 2022-04-29
Fix/733/map line max thickness

levity:
Adding to #840

### DIFF
--- a/src/pages/AssetMapPage/AMForceGraph.tsx
+++ b/src/pages/AssetMapPage/AMForceGraph.tsx
@@ -46,7 +46,6 @@ const COLOR_LINK_DIM = '#00000008';
 
 const MIN_ZOOM = 3;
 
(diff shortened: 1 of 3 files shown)


levity:
@sshmm please check this out. I was thinking about ways to avoid hardcoding `100` in your PR and ended up deciding to try some things out myself; this is what I came up with. If you like it, you can pull the changes into your PR and we can merge that.

sshmm:
> @sshmm please check this out. I was thinking about ways to avoid hardcoding `100` in your PR and ended up deciding to try some things out myself; this is what I came up with. If you like it, you can pull the changes into your PR and we can merge that. (Or merge this one)
@levity 

Nice fix. you  made it a factor of max gives between.
My approach assumed that the normal gives will be a factor 100 so until 100 gives the thickness will be 0.8 * diameter* gives/100
anything greater than 100 will be 0.9* diameter.

I only made a small change in my PR instead of multipying by 10 it will be radius *2 so that if someone changed the radius it will change accordingly

PR #848, opened 2022-04-29
Hasura ci

### LINKED ISSUE #705: Set up Hasura for CI
That is, whenever `yarn test:ci` is run, whether on a local dev machine or during a Github Action, a separate Hasura instance starts up so it can be accessed from Jest tests, and is cleaned up afterward. Just like Ganache does.

Should be completely standalone, i.e. we can add and remove data without affecting the Hasura instance that is being used for development.

Depends on starting a Vercel instance as well, for auth & other webhooks.

topocount:
Reopening #774 as the author so @levity can give an approving review.
closes #705

This also includes a working development build of the frontend.


scripts/ci.sh
```diff
@@ -12,5 +18,42 @@ if [ ! "$HARDHAT_FORK_BLOCK" ]; then
   echo
 fi
 
+# TODO: skip this if it's already running
```
topocount, line 21: docker compose up is idempotent so we can probably delete this todo.

scripts/ci.sh
```diff
@@ -12,5 +18,42 @@ if [ ! "$HARDHAT_FORK_BLOCK" ]; then
...
+
+if [ "$VERBOSE" ]; then
+  "${CMD[@]}" 2>&1 & PID=$!
+else
+  LOGFILE=${TMPDIR:-.}/hasura-ci-$(date +%s).log
+  echo "Writing output to" "$LOGFILE"
+  "${CMD[@]}" > "$LOGFILE" 2>&1 & PID=$!
+fi
+
+VERCEL_CMD=(vercel dev -t "$VERCEL_TOKEN" --confirm)
```
levity, line 34: can we call this `CI_VERCEL_TOKEN` to avoid confusion when it's set in the local .env file?

scripts/ci.sh
```diff
@@ -12,5 +18,45 @@ if [ ! "$HARDHAT_FORK_BLOCK" ]; then
   echo
 fi
 
+echo "Starting Hasura on port ${CI_HASURA_PORT}..."
+
+# comment out DOCKER_GATEWAY_HOST if running locally on macOS
+export DOCKER_GATEWAY_HOST=172.17.0.1
```
levity, line 24: can we put this in the .env instead?

topocount: yes, it should also be put in the github environment

levity (approved):

PR #859, opened 2022-05-03
Enables the remaining event triggers and fix the double trigger issue

zashton:
### Test Plan

1. Allocate to a teammate some GIVES
2. Delete that teammate from the circle
3. Make sure the GIVES are refunded correctly and not double counted.


META-DREAMER (approved):
Tested as described, works without issue

PR #860, opened 2022-05-04
fix #536 - cleanup created at updated at on all tables

### LINKED ISSUE #536: all tables need a trigger to update the `updated_at` column automatically. This behavior was handled in Laravel ORM previously.
e.g. 
```
CREATE FUNCTION trigger_set_timestamp()
RETURNS TRIGGER AS $$
BEGIN
  NEW.updated_at = NOW();
RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER set_timestamp
BEFORE
UPDATE ON profiles
FOR EACH ROW
EXECUTE PROCEDURE trigger_set_timestamp();
```

CryptoGraffe:
This PR fixes #536 by performing a number of db schema updates. 

* ensures that all `created_at` columns are non_nullable and have defaults
* ensures that all `updated_at` columns are 
* add updated_at triggers to all tables w/ updated_at columns

This has the nice side effect of making the date fields non-optional in the zeus generated types.


CryptoGraffe:
After @zashton deployed latest data to staging i ran this migration, found a bug, fixed it. I think this is well qualified now and just needs review.

topocount (approved):

PR #879, opened 2022-05-06
improving our error reporting to sentry

CryptoGraffe:


src/lib/zod/index.ts
```diff
@@ -261,11 +262,14 @@ export const HasuraUserSessionVariables = z
     'x-hasura-role': z.union([z.literal('user'), z.literal('superadmin')]),
     'x-hasura-address': zEthAddressOnly,
   })
-  .transform(vars => ({
-    hasuraProfileId: vars['x-hasura-user-id'],
-    hasuraRole: vars['x-hasura-role'],
-    hasuraAddress: vars['x-hasura-address'],
-  }));
+  .transform(vars => {
+    Sentry.setTag('action_user_id', vars['x-hasura-user-id']);
```
gregpassmore, line 266: I know this is sketchy but there is no other way to do this easily for every action with the way it is currently implemented - I might change it wrap the parse function and do it that way but this is ok for now

CryptoGraffe, line 266: I know this is sketchy but there is no other way to do this easily for every action with the way it is currently implemented - I might change it wrap the parse function and do it that way but this is ok for now

src/utils/reporting.ts
```diff
@@ -30,3 +41,21 @@ export const initSentry = () => {
...
+  if (error instanceof GraphQLError) {
+    // graphql error?
+    if (error.message) {
+      // this is fine
+      return error;
+    } else if (error.response.errors) {
+      // elevate the error to the top level so sentry (and logging) can handle it better
+      if (error.response.errors.length > 0) {
+        // return
+        error.message = error.response.errors[0].message;
```
levity, line 55:
```suggestion
        error.message = error.response.errors.map(e => e.message).join('; ');
```

levity (approved):

levity (approved):

levity (approved):

levity (approved):

Issue #808, opened 2022-04-21
allow selecting fixed payment token type in circle settings


levity:
- also have a "Disabled" option--this is the default
- autocomplete field that includes the token types from any existing vaults

Depends on #888

sshmm:
@levity Can I work on this?

levity:
sure!

sshmm:
@topocount  I was working on it. did you start working on it?

levity:
@sshmm how's it going?

levity:
@sshmm there have been a lot of changes in the frontend recently (e.g. #800 and #856). if you have some WIP, please push a draft request so we can see if there are any conflicts

sshmm:
@levity  I had several health issues in the past two weeks so I didn't do much. I replaced the AdminCircleModal and its subcomponents with stitches components and still working on the styling and the behavior with use-form then add the fixed payment option. 
if @topocount already started it and made better achievement, I can pick another issue.

sshmm:
> No worries. How about you push a draft PR when you have a chance, and then @topocount can consider building on top of it so your work doesn't go to waste?

ok, I will update it to an acceptable state and push it

PR #882, opened 2022-05-08
initial AdminCircleModal styling with stitches

sshmm:
## Summary
intial version for #808

## What is done
1- ApeToggleNew: a new react component using stitches for yes/No toggle switches.
2- AdminCircleModal is updated to use stitches for the Modal, Form, button, input fields and areas, toggles, and button
3- styles are updated to match the modal in production

## what needs to be done
1- ApeToggleNew needs to use the ui Tooltip 
2- the form needs to use use-form to function correctly
3- Hard coded values of the styles needs to be replaced with values from stitches.config
4- ApeAvatar needs to be replaces
5- add new new functionality


src/components/ApeToggleNew/ApeToggleNew.tsx
```diff
@@ -0,0 +1,111 @@
...
+  textTransform: 'none',
+  textAlign: 'center',
+  fontSize: 19,
+  lineHeight: 1.26,
+  padding: '12px 24px',
+  variants: {
+    active: {
+      true: {
+        color: '$white',
+        background: '$lightBlue',
```
levity, line 39: as you'll see if you look at the main branch, we've changed the color names. Please use `$primary` as the background for active toggles and `$surface` for inactive ones. Just set `opacity: 0.8` on hover instead of changing colors.

PR #883, opened 2022-05-09
Claim Distribution

teeolendo:
This PR covers the single claim workflow. It comprises of changes to the Main Nav by way of a claim button that shows up for unclaimed distributions scoped to the user. Also includes a new Claim Page folder to Pages which contains a hook for claiming, UI for the ClaimsPage and Hasura mutations and queries.

To test this feature

- Create a vault
- Select a completed epoch and submit a distribution for it
- If the epoch contained your user, they should see the claim button visible on the top nav
- When the navigate to the Claims page, they should be able to see unclaimed claims and previous claims, if any.
- Select `Claim (tokenName)` to process claim and that should result in the token being transferred to the user, a trx hash being added to the DB, the claim should move from the unclaimed to the claim table on the UI.

WIP:
- Still need to figure out the logic for multiple epochs grouping. Will talk with Lawrence about this.
- Claim Many


src/ui/TableBorder/TableBorder.tsx
```diff
@@ -0,0 +1,9 @@
+import { styled } from '@stitches/react';
```
levity, line 1: Can you just put this with Table? It's not really meaningful by itself

src/pages/ClaimsPage/Table.tsx
```diff
@@ -0,0 +1,72 @@
+import { ReactNode, useMemo, useState } from 'react';
```
levity, line 1: Can you move this to `src/components` and change `AllocationsTable` to use it as well, instead of duplicating code?

src/pages/VaultsPage/CreateForm.tsx
```diff
@@ -112,7 +112,7 @@ export const CreateForm = ({
   };
 
   const onSubmit = ({ symbol, customAddress }: any) => {
-    setSaving(true);
```
levity, line 115: is this change unintentional?

levity (approved):

PR #885, opened 2022-05-09
Implement createVault Hasura Action

topocount:
test plan:
1. `docker compose --profile dev down -v` && `yarn docker:start`
2. vercel dev
3. browse to http://localhost:3000/admin/vaults
4. create a new vault with any token and observe it populate

Here's a template mutation for ensuring the endpoint is secure:
```gql
mutation testCreateVault {
  createVault(payload: {chain_id: 1338, org_id: 2, vault_address: "0x<address>"}) {
    id
    vault {
      id
      created_at
      created_by
      decimals
      org_id
      simple_token_address
      symbol
      token_address
      updated_at
      vault_address
    }
  }
}

```

try different org IDs that the user might not have access to, and test out an invalid vault address.

### DIFF
--- a/api/hasura/actions/createVault.ts
+++ b/api/hasura/actions/createVault.ts
@@ -1,0 +1,93 @@
+import { AddressZero } from '@ethersproject/constants';
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { Contracts } from '../../../api-lib/contracts';
+import { adminClient } from '../../../api-lib/gql/adminClient';
+import * as queries from '../../../api-lib/gql/queries';
+import {
+  UnauthorizedError,
+  UnprocessableError,
+  InternalServerError,
+} from '../../../api-lib/HttpError';
+import { getProvider } from '../../../api-lib/provider';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+import {
+  createVaultInput,
+  composeHasuraActionRequestBodyWithSession,
+  HasuraUserSessionVariables,
+} from '../../../src/lib/zod';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    session_variables,
+    input: { payload },
+  } = composeHasuraActionRequestBodyWithSession(
+    createVaultInput,
+    HasuraUserSessionVariables
+  ).parse(req.body);
+
+  const { org_id, chain_id, vault_address } = payload;
+  const { hasuraAddress, hasuraProfileId } = session_variables;
+
+  const isOrgAdmin = await queries.checkAddressAdminInOrg(
+    hasuraAddress,
+    org_id
+  );
+
+  if (!isOrgAdmin)
+    throw new UnauthorizedError(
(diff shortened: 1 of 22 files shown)


hasura/metadata/actions.yaml
```diff
@@ -216,9 +226,10 @@ custom_types:
   - name: UpdateEpochInput
   - name: Allocation
   - name: Allocations
-  - name: AllocationCsvInput
   - name: CreateUsersInput
   - name: UserObj
+  - name: AllocationCsvInput
```
zashton, line 231: its weird, I have the same issue of this line moving whenever i update the metadata even though i've downgraded the hasura cli,

topocount: @zashton I have a feeling we merged changes with this line from a user with the CLI upgraded? Not sure how else this is possible

api/hasura/actions/createVault.ts
```diff
@@ -0,0 +1,93 @@
...
+          org_id,
+          vault_address,
+          created_by: hasuraProfileId,
+          token_address: yTokenAddress,
+          simple_token_address: simpleTokenAddress,
+        },
+      },
+      { id: true },
+    ],
+  });
```
zashton, line 86: we could add an operation name here

api/hasura/actions/createVault.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (!tokenAddress)
+    throw new UnprocessableError('No token specified for vault');
+
+  const token = contracts.getERC20(tokenAddress);
+  const [symbol, decimals] = await Promise.all([
+    token.symbol(),
+    token.decimals(),
+  ]);
+
+  const { insert_vaults_one: result } = await adminClient.mutate({
```
zashton, line 70: Not sure whether its an issue if the admin can create multiple same vaults under the same org. If its allowed I wonder if we need to add a editable custom name for each of them so they could distinguish between each of them.

zashton (changes requested):
tested the mutation through hasura console and via FE and the vault record is saved correctly, just have a minor qns regarding admin creating duplicate vaults.

api/hasura/actions/createVault.ts
topocount: Yeah I'm wondering about adding a uniqueness constraint for vaults

levity:
We can use the abbreviated address to distinguish between them. There
should be a uniqueness constraint on vault_address though

hasura/migrations/default/1652824553996_alter_table_public_vaults_alter_column_vault_address/up.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."vaults" add constraint "vaults_vault_address_key" unique ("vault_address");
```
zashton, line 1: should the unique key be `address` + `org_id` ?

topocount: @zashton I considered this, but then we could have a situation where a separate org is somehow using the same vault. I don't think that's desirable right now.

PR #890, opened 2022-05-10
Fix zeus5 allocations

levity:
Zeus 5 apparently rejects mutation aliases that start with numbers

@zashton, are we doing this anywhere else? if so, please add to this branch


zashton (dismissed):

topocount (commented):
~~Please hold off on merging this. I've found at least one more alias that should be updated and some resultant changes that need to be made downstream~~

can't delete or dismiss my own review comment

topocount (changes requested):
Please hold off on merging this. I think I've found at least one more alias that should be updated and some resultant changes that need to be made downstream.

I'll push the changes and update my review.

topocount (approved):
looking good now

PR #892, opened 2022-05-12
Fixed payment schema

### LINKED ISSUE #888: Create backend schema for fixed payments
Migrations
- new varchar column in circle: `fixed_payment_token_type` (token symbol)
- new boolean column in circle: `fixed_payment_enabled`
- new numeric column in user: `fixed_payment_amount`

Hasura
- allow circle admins to view & edit the new columns above
- allow circle users to view their own `fixed_payment_amount` & the circle's fixed payment settings

zashton:
fixes #888 

@levity created a private view for users so to restrict the permissions of their own `fixed_payment_amount`
`fixed_payment_enabled` removed. if `fixed_payment_token_type` has value == enabled

### DIFF
--- a/hasura/metadata/databases/default/tables/public_circles.yaml
+++ b/hasura/metadata/databases/default/tables/public_circles.yaml
@@ -81,6 +81,8 @@ select_permissions:
     - created_at
     - default_opt_in
     - discord_webhook
+    - fixed_payment_enabled
+    - fixed_payment_token_type
     - id
     - is_verified
     - logo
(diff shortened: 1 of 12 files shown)


topocount (commented):
this seems reasonable to me. My only question: How would circle admins modify a user's fixed payment amount?

zashton:
> this seems reasonable to me. My only question: How would circle admins modify a user's fixed payment amount?

i think the updates probably need to be made in `adminUpdateUser` + `updateCircle` later-on when we implement it.

zashton:
`fixed_payment_enabled` removed. if `fixed_payment_token_type` has value == enabled

topocount (approved):

Issue #853, opened 2022-04-29
New Design For Map


reeserj:
- move drawer to the left side, tweak it to match [Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=4002%3A14867)
- add page title & description
- no changes to map design, lines, etc

karelianpie:
Working on this one

reeserj:
> Working on this one

Assigned it to you

PR #914, opened 2022-05-15
feat: New Design For Map

karelianpie:
## Description

Implement the New Design For Map, [Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=4002%3A14867)

## Related Issue

Closes https://github.com/coordinape/coordinape/issues/853

## Motivation and Context

New designs for the map page

## How Has This Been Tested?

Ran locally, screenshots attached

Screen Shot 2022-05-21 at 2 36 48 pm

Screen Shot 2022-05-21 at 2 37 07 pm

Filtering by tag

Screen Shot 2022-05-22 at 9 18 40 am


crabsinger (commented):
@karelianpie do we need to use Emotion for this, vs Stitches.dev? I'm generally cautious about adding new packages unless really needed, and seems Emotion is similar to Stitches which we're moving towards. Pulling in the select component makes sense.

karelianpie:
> @karelianpie do we need to use Emotion for this, vs Stitches.dev? I'm generally cautious about adding new packages unless really needed, and seems Emotion is similar to Stitches which we're moving towards. Pulling in the select component makes sense.

Thanks for the review @crabsinger, I think `Emotion` is a dependency of `react-select`, but after discussing with @CryptoGraffe just now I will be refactoring this to use https://www.radix-ui.com/docs/primitives/components/select instead. So will remove the react-select dependency, which in turn should also remove Emotion.

crabsinger:
@karelianpie the tag filter doesn't appear to work...

src/pages/AssetMapPage/AMDrawer.tsx
```diff
@@ -113,45 +190,83 @@ export const AMDrawer = () => {
...
+            <RedSelect
+              value={metric2}
+              options={metricOptions}
+              onChange={v => setMetric2(v as MetricEnum)}
+            />
           )}
           <ApeAutocomplete
             onChange={setSearch}
             freeSolo
             options={SKILLS}
```
crabsinger, line 246: @karelianpie we need to filter by  the selected skill here

karelianpie: Nice catch @crabsinger, but I did not change that functionality.

I just checked and it seems that it is currently broken, should we open a bug card and fix that bug in there or should we increase the scope of this PR and fix it here? What do you reckon?

karelianpie: @crabsinger besides this issue that @sshmm is working on, is there anything else that you feel needs addressing in this PR?

karelianpie: @crabsinger filtering by selected skill was fixed by @sshmm, just rebased with main and it's working here too

Screen Shot 2022-05-22 at 9 19 16 am

src/pages/AssetMapPage/AMDrawer.tsx
```diff
@@ -89,6 +119,35 @@ export const AMDrawer = () => {
...
+  const [amEpochId, setAmEpochId] = useStateAmEpochId();
+
+  // This is the AssetMapPage Controller
+  useEffect(() => {
+    if (amEpochs.length === 0) {
+      return;
+    }
+    setAmEpochId(amEpochs[amEpochs.length - 1]?.id);
+    // TODO: Load gifts for selected epoch when needed:
+    // https://linear.app/yearn/issue/APE-192/api-dont-always-load-everything-and-improve-fetching
```
crabsinger, line 133: we can remove these comments now

src/ui/Select/Select.tsx
```diff
@@ -0,0 +1,164 @@
...
+          <ChevronDownIcon />
+        </SelectScrollDownButton>
+      </SelectContent>
+    </RadixSelect>
+  );
+};
+
+/* Storybook utility for stitches variant props
+
+NOTE: this can't live in the stories file because the storybook navigator will take a story and will crash
```
crabsinger, line 153: hmm, can you say more about the error that was happening here? did we document this issue or talk about it somewhere else?

crabsinger:
@karelianpie - nice work this is coming together really well! A few remaining things I see before its ready for primetime:


on prod, I see the tags show on the users in the Drawer, but on your branch I am not seeing tags displayed in the drawer.

Second, the show/hide behavior on the Filters title of the drawer doesn't match the Figma.

What I see on your branch:
Screen Shot 2022-05-23 at 11 12 32 AM

The figma design:
Screen Shot 2022-05-23 at 11 12 37 AM

karelianpie:
> @karelianpie - nice work this is coming together really well! A few remaining things I see before its ready for primetime:
> 
> on prod, I see the tags show on the users in the Drawer, but on your branch I am not seeing tags displayed in the drawer.
> 
> Second, the show/hide behavior on the Filters title of the drawer doesn't match the Figma.
> 
> What I see on your branch: Screen Shot 2022-05-23 at 11 12 32 AM
> 
> The figma design: Screen Shot 2022-05-23 at 11 12 37 AM

@crabsinger I think you're previewing an older version of my branch. Can you please check the most current one?

src/ui/Select/Select.tsx
karelianpie: @crabsinger this is an older version of the code, unsure how you're seeing this still

src/pages/AssetMapPage/AMDrawer.tsx
karelianpie: Removed with https://github.com/coordinape/coordinape/pull/914/commits/0fd0cee5b2f40d800b1a6cea019b23f099352955

src/ui/Select/Select.tsx
karelianpie: It wasn't added by me and it's in most stories, just followed the pattern. The commit that introduced that was https://github.com/karelianpie/coordinape/commit/258a2bcd1cd042aeb57027af6680361c25f741e5 but it's in a few components `src/ui/Button/Button.tsx`, `src/ui/IconButton/IconButton.tsx`, etc.

crabsinger: 👍

crabsinger (approved):
lgtm - lets ship it!

crabsinger:
Fixing it as a separate PR and then rebasing the branch on top to pull in
the fix is preferred approach. Didn’t realIze this was a live bug.

karelianpie:
> Fixing it as a separate PR and then rebasing the branch on top to pull in the fix is preferred approach. Didn’t realIze this was a live bug.

Sorry, I think I've missed something here. What do you mean this was a live bug?

Issue #375, opened 2021-12-20
Support Sign-In with Ethereum [EIP-4361] On Coordinape 


obstropolos:
What:

Update Coordinape's signature method to implement EIP-4361. Currently, we sign messages with this format: https://github.com/coordinape/coordinape/blob/f4743c2bd6866275b177c41ae25741378a37ea51/src/services/api.ts#L40

Changing the data payload to match EIP-4361 should be an easy change that helps us meet the standard.

**Problem**
Coordinape currently uses a custom way to sign users connecting their wallet into the platform, containing a login message and what appears to be a nonce. 

**Solution**
We would love if Coordinape would adopt [EIP-4361](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4361.md), or the new Sign-In with Ethereum standard to establish a session with users. The standardization of SIWE was supported by the Ethereum Foundation and ENS, and put together over a number of weeks in the open with community calls, feedback and participation being critical to the process. All of this is included on [login.xyz](login.xyz) which was used to host core information. 

Sign-In with Ethereum's goal in Web3 is to improve interoperability across existing services, while also providing wallet vendors a reliable method to identify signing requests as Sign-In with Ethereum requests for improved UX.

**Specification**
There are libraries being built to ease any adoption or integration concerns, namely for dapps: 
- Our initial TypeScript library which can be found [here](https://github.com/spruceid/siwe)
- In-progress [Ruby](https://github.com/spruceid/siwe-ruby) and [Rust](https://github.com/spruceid/siwe-rs) implementations

**Additional Notes** 
If Coordinape is open to adopting, we'd also love to add the logo to the list of supporters on login.xyz.

EsteFilipe:
Interested in this task - authentication using Ethereum is currently my main area of interest.

Can I get it assigned to me?

zashton:
> Interested in this task - authentication using Ethereum is currently my main area of interest.
> 
> Can I get it assigned to me?

Sure go ahead !

awoie:
We have worked on a PR but it is WIP. The WIP PR should be there today @EsteFilipe @zashton . It would be awesome if you could get it across the finish line.

awoie:
Current state of ^^ is sign in works, but we didn't have time to verify that post sign functionality is intact.

EsteFilipe:
Awesome, will start working on this tomorrow

awoie:
> EsteFilipe

[...]

PR #930, opened 2022-05-18
Support Sign-In with Ethereum [EIP-4361] On Coordinape

theosirian:
As mentioned in #375, this is the WIP version of SIWE support for Coordinape.

@obstropolos @awoie


awoie:
WIP PR, potentially fixes #375.  It would be amazing if @EsteFilipe could push this to the finish line.

src/services/api.ts
```diff
@@ -28,16 +29,28 @@ export class APIService {
...
+      nonce = await nonceReq.text();
     } catch (e) {
-      now = Date.now();
+      nonce = generateNonce();
     }
 
-    const data = `Login to Coordinape ${Math.floor(now / 1000)}`;
+    const message = new SiweMessage({
+      // TODO: replace by configured domain
+      domain: 'domain.tld',
```
crabsinger, line 42: I've been thinking about CSRF attacks - would be great if we can have this domain be checked against the request origin, and be rejected if it doesn't match - to prevent someone from creating a phishing site that prompts for  this sign in message and then steals your api key.

obstropolos: Part of the reason for SIWE is to work with wallet implementers on checking for domain binding with respect to parsing SIWE messages. We're currently working with a few on this at the moment.

crabsinger:
Closing this - will move ahead in PR https://github.com/coordinape/coordinape/pull/991

PR #932, opened 2022-05-19
Vercel fixes for serverless

topocount:
Vercel doesn't utilize the project's configured "build command" for
serverless functions, and the backend now needs access to the
deployment.json file. The postinstall script is a lightweight way to
address this. It also potentially allows us to simplify the build command
in vercel to dedup the effort of linking to the protocol repo.

test plan:

```sh
rm -rf node_modules
git submodule deinit --all
yarn --frozen-lockfile

yarn docker:start
vercel dev
```

browse to http://localhost:8080
Run the following mutation:
```gql
mutation MyMutation {
  createVault(payload: {chain_id: 10, org_id: 10, vault_address: "0x0"}){
    id
  }
}
```

If you get a zod error, that means the deployment.json file resolved
correctly

### DIFF
--- a/api-lib/config.ts
+++ b/api-lib/config.ts
@@ -53,7 +53,9 @@ export const INFURA_PROJECT_ID = getEnvValue(
   'missing-infura-id'
 );
 
-export const HARDHAT_GANACHE_PORT = getEnvValue('HARDHAT_GANACHE_PORT');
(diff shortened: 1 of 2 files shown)


zashton (approved):
tried and works as per test case

Issue #854, opened 2022-04-29
Visual Style Updates


reeserj:
* Edit 1.5 design tokens
- main areas
    - heading and body text sizes and bold levels
    - typefaces
    - colors - check in with @smombartz  to ensure they're finalized before making changes.
    - Button variants
        - primary, secondary, form, destructive, ... (organize variants by functionality)
        - filled or outlined
        - small, medium, large
    - add these to stitches first, then make Material-UI theme file depend on stitches
- Update components to refer to variants/tokens in theme instead of specific values (un-adHoc-ing)
- Create new components from [Design Library](https://www.figma.com/file/ZHiNW0Wf2bj2VHVIqpjMuh/Design-Library?node-id=105%3A2519)  (if they dont already exist)
- change all the button colors — should be covered by the changes above
    - all buttons on vaults page → blue
    - all submit buttons on all form modals → blue
    - next step button on allocation flow → blue
    - all 4 buttons on circle admin page → blue
    - vouching page → blue
    - grey buttons on circle admin settings modal → grey outline, no icons

levity:
Part 1 (#856) is laying the groundwork in the code, and can be merged into production as soon as it's approved; Part 2 will be switching over the colors and buttons to the 1.5 style, and should be batch-released with the other tasks in the [milestone](https://github.com/coordinape/coordinape/milestone/8).

wingfeatherwave:
now part of https://github.com/coordinape/coordinape/pull/934

PR #934, opened 2022-05-19
852 new navigation bar

### LINKED ISSUE #852: New Navigation Bar
- all left-justified ([Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=2711%3A5812))
- when on Home/Dashboard or Vaults page, don’t show any links to the right (which are all specific to a circle)
    - don’t rename Allocation to Rewards just yet — see below
- the caret to the right of a circle name opens a dropdown that lets you select a different circle from across all your orgs ([Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=4734%3A15397))
- mobile: put in hamburger menu, show vertically
- ~~add ability to upload logo for an org~~ (moved to #861)

Depends on #861

wingfeatherwave:
Closes #852

- Style updates to font and colors in main header
- New "overview" nav component in upper left with links to circles, overview, and someday Vaults
- Style of "profile" nav in upper right to match v1.5 styles
- Some new svg icons
- stitches variant updates
- move getOrgData to global position within OverviewMenu 
- Use stitches / radix for popovers

Screen Shot 2022-05-19 at 3 02 40 PM
Screen Shot 2022-05-19 at 3 02 49 PM


src/components/MainLayout/MainHeader.tsx
```diff
@@ -48,53 +51,60 @@ export const MainHeader = () => {
...
-        src="/svgs/logo/logo.svg"
-      />
-      {hasCircles && <TopLevelLinks links={mainLinks} />}
-
+      <Box
+        css={{
+          mt: 'calc($sm - 1px)',
+          mb: '-2px',
+          pb: '$xs',
+          px: '20px',
```
META-DREAMER, line 65: Should use a theme token instead of a hard coded value here

src/components/MainLayout/MainHeader.tsx
```diff
@@ -284,46 +294,38 @@ const MobileAvatar = () => {
   );
 };
 
-const linkStyle = {
+export const menuGroup = {
```
META-DREAMER, line 297:
```suggestion
export const menuGroupStyle = {
```

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -24,25 +25,20 @@ const useStyles = makeStyles(theme => ({
...
   },
   popover: {
-    width: 237,
     marginTop: theme.spacing(0.5),
     padding: 0,
     borderRadius: 8,
-    background:
-      'linear-gradient(180deg, rgba(255, 255, 255, 0) 0%, rgba(223, 237, 234, 0.4) 40.1%), linear-gradient(180deg, rgba(237, 253, 254, 0.4) 0%, rgba(207, 231, 233, 0) 100%), #FFFFFF',
-    boxShadow: '0px 4px 6px rgba(181, 193, 199, 0.16)',
+    background: '$surface',
```
META-DREAMER, line 33: Yay for removing hard coded styles!

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -24,25 +25,20 @@ const useStyles = makeStyles(theme => ({
...
+    boxShadow:
```
META-DREAMER, line 34: Can we define this as a reusable shadow style in the theme?

src/components/OverviewMenu/OverviewMenu.tsx
```diff
@@ -0,0 +1,217 @@
...
+    background: '#FFF',
+    boxShadow:
+      '0px 0px 3px 0px #0000001C, 0px 0px 16px 0px #0000001F, 0px 0px 87px 0px #0000003D',
+    display: 'flex',
+    flexDirection: 'column',
+    top: '13.5px !important',
+    left: '45px !important',
+    transition: 'none !important',
+  },
+}));
```
META-DREAMER, line 43: Can we do this using stitches instead of Material UI?

src/components/OverviewMenu/OverviewMenu.tsx
```diff
@@ -0,0 +1,217 @@
...
+          key={path}
+          to={path}
+          className={matchPaths?.includes(location.pathname) ? 'active' : ''}
+        >
+          {label}
+        </Link>
+      ))}
+    </>
+  );
+};
```
META-DREAMER, line 204: Would be good to pull these out into separate component files to keep things more modular

META-DREAMER (changes requested):
great work! Few minor suggestions

src/components/MainLayout/MainHeader.tsx
wingfeatherwave: Yea, this one is a pretty tricky one to position exactly.  There weren't any tokens that got the alignment right (this is for the COORDINAPE brand text above the OverviewMenu button.  I'll see what I can do to token-ize it

src/components/OverviewMenu/OverviewMenu.tsx
wingfeatherwave: I'll look into transitioning to a radix / stitches popover component

src/components/OverviewMenu/OverviewMenu.tsx
wingfeatherwave: I'd like to punt on moving this out to another component for the time being.  The MainHeader component has something similar, but it seems like a good deal of complexity to refactor it neatly.

src/components/MainLayout/MainHeader.tsx
```diff
@@ -48,53 +51,60 @@ export const MainHeader = () => {
     <Box
       css={{
         px: '$1xl',
-        pt: '9px', // manual offset to align breadcrumb
-        height: '80px',
+        height: '91px',
```
META-DREAMER, line 54: magic number not ideal, should add it to theme.sizes under `headerHeight` or something to that effect

META-DREAMER:
The overview menu is showing in the logged out state with nothing in it, should probably disable it until the user is signed in

CleanShot 2022-05-24 at 23 06 45

src/components/MainLayout/MainHeader.tsx
META-DREAMER: I would put the onus on the designers to update the figma so it conforms to the standard sizing / spacing tokens.

In the designs there is much less padding on the top level box, so you could remove the `px: '$1xl'` there on line 53 and then just change this to `px: '$lg'` which matches the 24px in the designs.

The designs incorrectly have a 12px space between the overview menu button and the left edge, the space scale dictates it should be either 8px or 16px.

Don't have to deal with it in this PR if you dont want to, but would atleast put a TODO comment here and let design team know

image

src/components/MainLayout/MainHeader.tsx
```diff
@@ -385,7 +387,7 @@ const CircleNav = () => {
 
   const links: [string, string, string[]?][] = useMemo(() => {
     const l: [string, string, string[]?][] = [
-      [paths.history, 'History'],
+      // [paths.contributions, 'Contributions'],
```
META-DREAMER, line 390: should remove this line if this isn't needed anymore, or uncomment if it is

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -51,67 +40,78 @@ export const MyAvatarMenu = () => {
...
+        <Popover>
+          <PopoverTrigger asChild>
+            <Link href="#">
+              <ApeAvatar profile={myProfile} className={classes.avatarButton} />
             </Link>
-            <Link
+          </PopoverTrigger>
+          <PopoverContent
+            sideOffset={-66}
+            alignOffset={-16}
```
META-DREAMER, line 54: Magic numbers not ideal. Should at least extract into constants and/or add comments explaining how you got to these values, if there's no way to avoid them

src/ui/Popover.tsx
```diff
@@ -0,0 +1,17 @@
+import * as PopoverPrimitive from '@radix-ui/react-popover';
+import { styled } from '@stitches/react';
+
+const StyledContent = styled(PopoverPrimitive.Content, {
+  padding: 0,
+  borderRadius: 8,
```
META-DREAMER, line 6: should use value from theme tokens. `$3` should be equivalent to 8px
image

META-DREAMER (changes requested):
Great job refactoring to use Radix! one step closer to nuking Material UI

Few more small nits, after this should be good to merge!

src/components/MainLayout/MainHeader.tsx
wingfeatherwave: removed

src/components/MyAvatarMenu/MyAvatarMenu.tsx
wingfeatherwave: This is a tough one, as there are no easy calculations here, and we can only use integers (no css values).  IMO, this needs to be dialed in browser.  The problem is that the popover tool on open appends a new dom node to body with the popover positioned with Radix translate3d 'magic'.  Ideally the popover would be positioned absolutely relative to a parent of the trigger since the design shows the popover perfectly overlaying the trigger, but that's just not how this tool works :(  I guess we could roll our own, but that's not insubstantial as a ui task.  I added a short comments in the code.

wingfeatherwave:
@META-DREAMER I really appreciate all your feedback and notes.  Thank you!

META-DREAMER (approved):

src/components/MainLayout/MainHeader.tsx
levity: I agree with @META-DREAMER that it's good to explain magic numbers, but I don't think this should be put into the theme, because it's unlikely to be reused outside of this component. It could be a const declared at the top of this file with an explanatory comment

CryptoGraffe:
Coming in super late and way after the design phase, but I'm wondering why Circles is in the profile menu when it's already handled by the main nav? This is true for the currently deployed implementation as well.

levity:
Just to ease the transition for old users who were used to seeing it there

wingfeatherwave:
https://github.com/coordinape/coordinape/issues/854 merged into this branch

src/pages/CirclesPage/CirclesPage.tsx
```diff
@@ -17,8 +18,6 @@ import { Box, Button, Link, Panel, Text } from 'ui';
 import { Torso } from 'ui/icons';
 import { SingleColumnLayout } from 'ui/layouts';
 
-import { getOrgData } from './getOrgData';
```
levity, line 20: i'm still looking through this PR, but i wanted to give this comment right away: i think it would be better to write a smaller query for OverviewMenu's needs rather than use this one. There's stuff in this query like epochs and nominees that doesn't need to be fetched for that.

src/routes/paths.ts
```diff
@@ -40,6 +40,7 @@ export const paths = {
   allocation: '/allocation',
   circles: '/circles',
   connectIntegration: '/connect-integration',
+  contributions: '/contributions',
```
levity, line 43:
```suggestion
```

src/routes/paths.ts
```diff
@@ -59,6 +60,7 @@ export const paths = {
 const circleSpecificPathKeys: (keyof typeof paths)[] = [
   'adminCircles',
   'allocation',
+  'contributions',
```
levity, line 63:
```suggestion
```

src/ui/Spacer/Spacer.tsx
```diff
@@ -0,0 +1,21 @@
+import { styled } from '../../stitches.config';
```
levity, line 1: looks like this isn't used anymore and can be removed

levity (changes requested):
Just a few other small things; otherwise, looks great! Nice work

src/pages/CirclesPage/getOrgData.ts
```diff
@@ -40,6 +40,6 @@ export const getOrgData = (address: string) =>
       ],
     },
     {
-      operationName: 'getOrgData',
+      operationName: 'getMenuData',
```
levity, line 43:
```suggestion
      operationName: 'getOrgData',
```

levity (approved):

PR #941, opened 2022-05-20
fix input to update_users mutation by removing the new_address field

### LINKED ISSUE #940: field "new_address" not found in type: 'users_set_input'
`field "new_address" not found in type: 'users_set_input'` fired as a GraphQL error when updating a user's address.

CryptoGraffe:
Fixes #940 

Test plan:

1. As an admin, try to update the address for a circle member.
2. it should work


topocount (approved):
good catch!

PR #943, opened 2022-05-20
test for user address editing, related to #940

CryptoGraffe:


topocount (approved):
Nice work!

Updated some details because the seed has changed.

n.b. @CryptoGraffe you might want to freshly seed your db from origin/main

topocount (approved):

PR #950, opened 2022-05-22
Fix potential security issue in circleAdmin middleware

META-DREAMER:
It only checked if the role was not admin, but didnt explicitly test if it was admin for the default allow case. Updated it to reject the request by default and only allow it if it passes explicit checks for either hasura admin or circleAdmin roles

### DIFF
--- a/api-lib/circleAdmin.ts
+++ b/api-lib/circleAdmin.ts
@@ -16,19 +16,28 @@ import { verifyHasuraRequestMiddleware } from './validate';
 
 const middleware =
   (handler: VercelApiHandler) =>
-  async (req: VercelRequest, res: VercelResponse) => {
+  async (req: VercelRequest, res: VercelResponse): Promise<void> => {
     try {
       const {
         input: { payload: input },
         session_variables: sessionVariables,
       } = composeHasuraActionRequestBody(circleIdInput).parse(req.body);
 
-      if (sessionVariables.hasuraRole !== 'admin') {
+      // the admin role is validated early by zod
+      if (sessionVariables.hasuraRole === 'admin') {
+        await handler(req, res);
+        return;
+      }
+
+      if (sessionVariables.hasuraRole === 'user') {
         const { circle_id } = input;
         const profileId = sessionVariables.hasuraProfileId;
 
         const { role } = await getUserFromProfileId(profileId, circle_id);
         assert(isCircleAdmin(role));
+
+        await handler(req, res);
+        return;
       }
     } catch (err) {
       if (err instanceof z.ZodError || err instanceof GraphQLError) {
@@ -37,9 +46,9 @@ const middleware =
       }
       throw new UnauthorizedError('User not circle admin', err);
     }
-    // the admin role is validated early by zod
 
-    await handler(req, res);
+    // Reject request by if not validated above
+    throw new UnauthorizedError('User not circle admin');
   };
 
 const isCircleAdmin = (role: number): boolean => role === 1;


topocount (dismissed):

topocount:
@META-DREAMER if you rebase the tests should pass. sorry about the confusion

zashton (approved):

topocount:
> > @META-DREAMER if you rebase the tests should pass. sorry about the confusion
> 
> Rebased, still failing

It looks like these changes are falsely asserting that an admin user is not and admin when trying to update an epoch and updating a user's address.

api-lib/circleAdmin.ts
```diff
@@ -23,7 +23,12 @@ const middleware =
         session_variables: sessionVariables,
       } = composeHasuraActionRequestBody(circleIdInput).parse(req.body);
 
-      if (sessionVariables.hasuraRole !== 'admin') {
+      if (sessionVariables.hasuraRole === 'admin') {
+        await handler(req, res);
```
topocount, line 27: if we're invoking the handler early, we need to return from it to avoid executing any of the further code

api-lib/circleAdmin.ts
```diff
@@ -37,9 +42,9 @@ const middleware =
       }
       throw new UnauthorizedError('User not circle admin', err);
     }
-    // the admin role is validated early by zod
 
-    await handler(req, res);
+    // Reject request by if not validated above
+    throw new UnauthorizedError('User not circle admin');
```
topocount, line 47: In it's current state, all traffic from website will hit this error, since no hasura-level admin authentication is possible. This is why E2E tests are failing right now.

api-lib/circleAdmin.ts
```diff
@@ -37,9 +42,9 @@ const middleware =
       }
       throw new UnauthorizedError('User not circle admin', err);
     }
-    // the admin role is validated early by zod
```
topocount, line 40: This is/was a key comment

api-lib/circleAdmin.ts
```diff
@@ -23,7 +23,12 @@ const middleware =
         session_variables: sessionVariables,
       } = composeHasuraActionRequestBody(circleIdInput).parse(req.body);
 
-      if (sessionVariables.hasuraRole !== 'admin') {
+      if (sessionVariables.hasuraRole === 'admin') {
+        await handler(req, res);
+        return;
+      }
+
+      if (sessionVariables.hasuraRole === 'user') {
```
topocount, line 31: Given the new pattern you're implementing, this block requires a handler invocation at the end of its scope. This is also the default case.

topocount (changes requested):
@META-DREAMER This isn't a potential security issue with the current state of our permissioning schema, but it is worth revisiting now that we're adding api keys.

Currently, This middleware was designed to work as follows:

It receives an authenticated request from hasura from either the builtin `hasura-admin` token holder or a user who has authenticated via their wallet. If a hasura admin is making the request they are already authenticated and therefore are able to skip straight to the handler.

A user needs to be verified as a circle admin. If they have a role of 1 for the circle_id provided as input, they are allowed to continue and invoke the handler. Please refactor so user-level auth has a path to a handler invocation.

I'm also okay with holding off on these changes until api tokens need to be handled.

api-lib/circleAdmin.ts
META-DREAMER: restored in the handler for the admin case

api-lib/circleAdmin.ts
META-DREAMER: whoops, good catch

topocount (approved):

PR #952, opened 2022-05-23
Implement action to generate API keys

META-DREAMER:
Setup action to generate API keys and gave circle admins permissions to generate, select and delete them as well.


topocount (commented):
The code looks good to me! Will test in a bit

api/hasura/actions/generateApiKey.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  const apiKey = generateTokenString();
+  const hash = hashTokenString(apiKey);
+
+  await adminClient.mutate({
+    insert_circle_api_keys_one: [
+      { object: { ...input, hash, created_by: userId } },
+      { hash: true },
+    ],
+  });
```
zashton, line 41: missing the operation name

api/hasura/auth.ts
```diff
@@ -17,18 +17,16 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
...
+
     const tokenRow = await adminClient.query(
       {
         personal_access_tokens: [
           {
             where: {
               tokenable_type: { _eq: 'App\\Models\\Profile' },
-              id: { _eq: parseInt(expectedId) },
-              token: { _eq: hashedToken },
+              id: { _eq: parseInt(prefix, 10) },
```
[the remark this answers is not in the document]
META-DREAMER: updated. curious to hear your reasoning.

api/hasura/auth.ts
```diff
@@ -17,18 +17,16 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
     }
 
     assert(req.headers?.authorization, 'No token was provided');
-    const [expectedId, token] = req.headers.authorization
-      .replace('Bearer ', '')
-      .split('|');
-    const hashedToken = crypto.createHash('sha256').update(token).digest('hex');
+    const { prefix, tokenHash } = parseAuthHeader(req.headers.authorization);
```
crabsinger, line 20: nice refactor of this method into a helpers file!

hasura/metadata/databases/default/tables/public_circle_api_keys.yaml
```diff
@@ -2,6 +2,45 @@ table:
...
+    - read_circle
+    - read_epochs
+    - read_member_profiles
+    - read_nominees
+    - read_pending_token_gifts
+    - update_circle
+    - update_pending_token_gifts
+    - hash
+    - name
+    - created_at
```
crabsinger, line 26: if we're exposing `created_at` maybe also expose `updated_at` ?

hasura/migrations/default/1653331923087_add_comments_on_tables/down.sql
```diff
@@ -0,0 +1,8 @@
+
+comment on TABLE "public"."profiles" is E'NULL';
```
crabsinger, line 2: are these comments visible in our graphql explorer? if so, maybe we should add some content here while going through effort of making this migration?

hasura/migrations/default/1653331923087_add_comments_on_tables/up.sql
```diff
@@ -0,0 +1,8 @@
+
+comment on TABLE "public"."pending_token_gifts" is E'GIVE allocations made by circle members for the currently running epoch';
+
+comment on TABLE "public"."token_gifts" is E'GIVE allocations made by circle members for past epochs';
```
crabsinger, line 4:
```suggestion
comment on TABLE "public"."token_gifts" is E'GIVE allocations made by circle members for completed epochs';
```

crabsinger (dismissed):
this looks good to me at a high level!

hasura/metadata/databases/default/tables/public_circle_api_keys.yaml
META-DREAMER: They should never be updated, just created or deleted. If a user needs to change scope for an API key, they should regenerate it.

Maybe in the future we could allow changing scopes for an "installed integration" when we setup the flow for verified 3rd party apps, but with this API key setup delete + create only is best practice. (e.g. see Github and generating personal access tokens)

hasura/migrations/default/1653331923087_add_comments_on_tables/down.sql
META-DREAMER: This is a down migration for rolling back the migration that adds comments to these tables (`down.sql` vs `up.sql`). When set to null it wont show anything in the explorer, just like most of the tables right now.

You are right though that we should have comments on all our tables to document our API better. Happy to go do a pass of that after this work

zashton (approved):
tested generate api key action handler via hasura console 
and normal login via UI. No issues encountered

hasura/migrations/default/1653331923087_add_comments_on_tables/down.sql
crabsinger: oops - didn't realize this was a down.

PR #961, opened 2022-05-26
Cypress DX improvements

levity:
## Motivation and Context

I needed to fix a Cypress test and wanted to speed up the code-test cycle. Also, I was encountering an error when bringing the CI docker containers down, because it was trying to stop a docker network that my dev containers were still using.

## Description

- use a different project name in Docker for CI so we don't conflict with containers that are running for dev
- remove localstack as a dev dependency (does this break anyone's workflow? if so, we will need to have a different localstack container for CI with different ports)
- allow running only Cypress tests, and just one test file, e.g.: 
```
yarn test:ci --cypress-only -s "**/updateUser.ts"
```

## Test and Deployment Plan

- Tested locally
  - sort of... The Cypress command itself is hanging on my machine, but I believe this to be caused by something unrelated, so I subbed in a no-op command to test that the containers & script behave correctly
- Let's see if CI passes

## Reviewers

@topocount


scripts/ci.sh
```diff
@@ -21,7 +22,7 @@ fi
 echo "Starting Hasura on port ${CI_HASURA_PORT}..."
 
 # comment out DOCKER_GATEWAY_HOST if running locally on macOS
```
topocount, line 24:
```suggestion

```

topocount (changes requested):
@levity I'm naively running this according to the prescribed test plan and running into environmental port config issues, so the devX here still isn't really accessible to most. I wrote `scripts/ci_local.sh` to get around this issue, though setup is still fairly manual, even if detailed in the shell script. I also recommend running `yarn cypress open` when working locally.

My next steps to streamline cypress in our codebase include parallelizing its constituent integration tests, along with cypress itself alongside the unit test setup. It's the only way we can scale the integration tests sustainably to a significant extent, and also creating a single command setup for local cypress testing (that can work in CI more or dev mode).

This unfortunately means that streamlining this specific script for cypress further is likely not a great use of time, since it'll be changing soon. However, any progress made here will surely help with above future work, so it's up to you. If we can find a way to handle environment setup better here than I did in `ci_local.sh`, that'll be welcome progress.

levity:
> @levity I'm naively running this according to the prescribed test plan and running into environmental port config issues

nuts. what's the error you were seeing?

topocount:
> nuts. what's the error you were seeing?
It looks like we need to source .ci.env in seed_hasura.sh
```sh
$ scripts/seed_hasura.sh --clean
$ hasura --skip-update-check --envfile ../.env --project ./hasura metadata apply
ERRO connecting to graphql-engine server failed
INFO possible reasons:
INFO 1) Provided root endpoint of graphql-engine server is wrong. Verify endpoint key in config.yaml or/and value of --endpoint flag
INFO 2) Endpoint should NOT be your GraphQL API, ie endpoint is NOT https://hasura-cloud-app.io/v1/graphql it should be: https://hasura-cloud-app.io
INFO 3) Server might be unhealthy and is not running/accepting API requests
INFO 4) Admin secret is not correct/set
INFO
FATA[0000] making http request failed: Get "http://localhost:8080/v1/version": dial tcp [::1]:8080: connect: connection refused
```

topocount (commented):
@levity I'm still not a huge fan of making localstack solely part of the CI docker profile because developers should probably be more inclined to use yarn `cy:dev` and therefore wont really care if there is an error while the ci profile is coming down. What do you think?

topocount (approved):

PR #975, opened 2022-05-29
fix repeated epoch number for same circle

zashton:
Solves the issue of repeated epoch number for the circle

### DIFF
--- a/api/hasura/cron/epochs.ts
+++ b/api/hasura/cron/epochs.ts
@@ -646,7 +646,7 @@ async function setNextEpochNumber({
           {
             where: {
               circle_id: { _eq: circle_id },
-              ended: { _eq: true },
+              number: { _is_null: false },
             },
           },
           { aggregate: { max: { number: true } } },
--- a/api/hasura/cron/epochs.ts
+++ b/api/hasura/cron/epochs.ts
@@ -646,7 +646,7 @@ async function setNextEpochNumber({
           {
             where: {
               circle_id: { _eq: circle_id },
-              number: { _is_null: false },
+              end_date: { _lte: 'now()' },
             },
           },
           { aggregate: { max: { number: true } } },


topocount (approved):
Nice find

topocount (commented):
I'm not sure this makes sense, since we try to assign a number as early as possible

zashton:
@topocount 
I think the sequence of events should be like this
 
 1. Two epochs are created, having 1 end date the same as the upcoming start date
 2. 1st epoch starts and has the number issued to it during the 1st cron run
 3. Next 1st epoch ends, since 2nd epoch starts as well at the same datetime it is also processed during the same cron run
 ```ts
 const results = await Promise.all([
    notifyEpochStart(epochResult),
    notifyEpochEnd(epochResult),
    endEpoch(epochResult),
  ]);
 ``` 
 4. 2nd epoch gets issued the next epoch number in `notifyEpochStart` since 1st epoch hasn't been marked `ended` yet it gets issued to 1st epoch's number.
 
 ```ts
          where: {
            circle_id: { _eq: circle_id },
            ended: { _eq: true },
          },
        },
        { aggregate: { max: { number: true } } },
 ```
 5. 1st epoch gets marks as ended after 2nd epoch gets the same number
 
 If we want to ensure the epoch number moves in sequence of start date instead of creation order we probably could force `endEpoch` to run first before
```
    notifyEpochStart(epochResult),
    notifyEpochEnd(epochResult),
```

api/hasura/cron/epochs.ts
```diff
@@ -646,7 +646,7 @@ async function setNextEpochNumber({
           {
             where: {
               circle_id: { _eq: circle_id },
-              ended: { _eq: true },
+              number: { _is_null: false },
```
topocount, line 649:
```suggestion
              end_date: { _lte: 'now()' },
```

topocount (commented):
@zashton I think this solves it. Took me too long to think of this lol.

I believe this works because it doesn't matter whether or not the epoch ending sequence has been processes. If the cron just is asynchronusly handling the creation of a new epoch and closing an old one in the same run, this query will work regardless of the status of past queries.

Issue #981, opened 2022-05-31
Regression: users can hover over circles they are not a member of and the various buttons show up 


CryptoGraffe:
This allows them to navigate to a circle they don't have permissions for, even messing up their local storage and their ability to load the site in general.

gregpassmore:
Fixed by #984

CryptoGraffe:
Fixed by #984

Issue #951, opened 2022-05-23
New Circle page just spins if there's no wallet connected


reeserj:
**Summary:**

1. **What is actually happening:**
New Circle page will load forever if there's no wallet connected
2. **What should be happening:**
We should have a prompt to connect your wallet while that page is loading.
3. **Steps to recreate:**
Go to https://app.coordinape.com/new-circle with your wallet disconnected.
4. **Screenshots/GIF/Video recording**
<img width="1244" alt="image" src="https://user-images.githubusercontent.com/82838264/169853766-51651d97-36a7-4906-91c4-1acf99bf0170.png">

@AlexDistill or @smombartz do you have opinions on how this text should be displayed?

smombartz:
@reeserj depending on what makes most sense, forwarding users to a login page or showing an overlay both options here (there should probably be a standard way of doing this so its the same for all instances)

- [Connect Wallet Page](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=9329%3A28325)
- [Connect Wallet Overlay](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=9329%3A28576)

akkikumar72:
I would like to work on it. Please assign it to me.

reeserj:
Hey @akkikumar72 just following up on this one! 

Anything blocking you here?

levity:
It's going to be fixed by my PR #989

levity:
Should be fixed now (pending deploy to prod)

PR #989, opened 2022-06-03
Add circle ID to URLs

levity:
## Motivation and Context

Put the circle ID in the URL and use that as the single source of truth for what circle should be displayed on a given page. This removes the need to "guess" the current circle, which makes it easier to bridge old (fetchManifest-based) and new (Zeus-based) pages, and addresses some bugs (#951, #981, et al.).

## Description

- add circle ID to URL for all circle-specific routes
- don't store circle ID in local storage
- rework auth flow so it's completed before any React Router Routes are rendered

## Test and Deployment Plan

- load all circle-specific pages, both from within the app and as the first page in the browser tab
- try loading circle-specific pages and circle-admin-only pages from scratch while logged in and logged out
- try out all steps in login flow (login, logout, cancel signature prompt, use Wallet Connect, etc)
- verify that Dework integration goes to the right callback page


src/components/MainLayout/MainLayout.tsx
```diff
@@ -45,3 +60,41 @@ export default MainLayout;
...
 };
+
+type AuthSteps = 'connect' | 'sign' | 'done';
+
+const RequireAuth = (props: { children: ReactElement }) => {
+  const address = useWalletAuth().address;
+  const web3Context = useWeb3React();
+  const { finishAuth } = useApiBase();
+  const [authStep, setAuthStep] = useState<AuthSteps>('connect');
+  console.log(authStep, web3Context.active, web3Context.account, address); // eslint-disable-line
```
levity, line 71: remove this

src/components/MainLayout/MainLayout.tsx
```diff
@@ -45,3 +60,41 @@ export default MainLayout;
...
+  useEffect(() => {
+    // reset after logging out or signature error
+    if (authStep !== 'connect' && !web3Context.active) {
+      setAuthStep('connect');
+      return;
+    }
+
+    if (authStep === 'connect' && web3Context.active) {
+      setAuthStep('sign');
+      finishAuth({ web3Context }).then(success => {
```
levity, line 82: might need to do `catch` here to handle expired token

src/hooks/useApiBase.ts
```diff
@@ -95,35 +60,40 @@ export const useApiBase = () => {
...
               connectorName,
               address,
               authTokens: { ...authTokens, [address]: token },
             });
 
+            // wrapped in setTimeout so rWalletAuth is updated before this runs
+            return new Promise(
+              (res, rej) =>
+                setTimeout(() => fetchManifest().then(res).catch(rej))
+              // TODO logout if this fails?
```
levity, line 73: could test for 401 response code here, then log out

src/pages/DefaultPage/DefaultPage.tsx
```diff
@@ -57,47 +53,24 @@ const useStyles = makeStyles(theme => ({
...
-        </p>
-      </Wrapper>
-    );
-  }
   // still loading
   if (!myProfile) return null;
 
   if (!selectedCircle) {
     return (
-      <Wrapper>
```
levity, line 100: need to fix this so that if there is some other circle you could be in, just not selected, you get redirected to /circles

topocount (approved):
sweet! This is so great

levity:
thanks for the review @topocount. after this next deploy, i'll run some manual testing on the preview app & if that all goes well i'll merge it.

PR #991, opened 2022-06-03
WIP: Feat/siwe

crabsinger:
copy of https://github.com/coordinape/coordinape/pull/967/commits


levity:
@crabsinger is this intended to replace #967? If so, can we close that one?

crabsinger:
@levity yes, this took the work of #930 and #967 and polished it up a bit. I'm awaiting a review and approval (mainly of the regex domain checking) and then this should be ready to merge.

reeserj:
After the initial connection to Coinbase wallet Attempting to connect to the Coinbase wallet option calls my Frame wallet browser extension rather than bringing up the Coinbase QR code to scan. (Assuming that's due to Frame and Coinbase wallet extensions sharing some code) 
image

The signature also displays the Coinbase wallet that I had previously connected with. (I think after we manually disconnect we need to make sure we clear any browser cache/Cookies in the app)
image

I can connect regularly through coinbase wallet if I go in through incognito mode and turn off my browser extension.

But then if I try and connect through WalletConnect after connecting with Coinbase wallet we get the signature and expected account don't match error 
image

topocount (approved):
LGTM!

PR #972, opened 2022-05-28
fix: Use circle token name instead of hardcoded GIVE

karelianpie:
## Motivation and Context

When a user changes the GIVE token name there were a few places were we were hardcoding "GIVE", it should be dynamically set depending on the circle's token name.

## Description

Change the places where "GIVE" was hardcoded to use the token name from the circle.

## Test and Deployment Plan

Ran the app locally and saw the change for most of them, not totally sure how to test the following:

* api-lib/handleRefundGiveMsg.ts
* api-lib/handleOptOutMsg.ts
* api/hasura/cron/epochs.ts

## Screenshots (if appropriate):

Before
Screen Shot 2022-05-28 at 1 15 28 pm
After
Screen Shot 2022-05-28 at 1 21 51 pm

Before
Screen Shot 2022-05-28 at 1 30 20 pm
After
Screen Shot 2022-05-28 at 1 35 44 pm

After
Screen Shot 2022-05-28 at 1 22 39 pm
After
Screen Shot 2022-05-28 at 1 22 22 pm
After
Screen Shot 2022-05-28 at 1 40 20 pm

## Related Issue

Closes https://github.com/coordinape/coordinape/issues/924

PR #992, opened 2022-06-03
fix: Add `token_name` to `mockCircle`

karelianpie:
## Motivation and Context

Type error when running `yarn typecheck` - https://github.com/coordinape/coordinape/runs/6715380522?check_suite_focus=true

## Description

Added `token_name` to the circle mock

## Test and Deployment Plan

* Ran `yarn typecheck` and saw it pass

## Reviewers

@topocount @levity

## Related Issue

https://github.com/coordinape/coordinape/pull/972#issuecomment-1146286464

PR #993, opened 2022-06-03
Fix type error from #972

topocount:


karelianpie:
My previous PR has the same changes @topocount, https://github.com/coordinape/coordinape/pull/992

crabsinger (approved):

PR #1003, opened 2022-06-06
remove recoil for allocation workflow

CryptoGraffe:
This work removes a lot of dependency on recoil for global state that is now page level state using vanilla react instead of recoil. It also breaks things up and moves stuff around quite a bit.

Test Plan:
1. Test everything to do with epoch statement and opt-in etc. 
2. Test everything to do with teammate selectionm
3. Test everything to do with allocation

- [x] Select Team isn’t getting marked as “completed”, refresh its fine
- [x] Stuff is hanging and then logging me out when saving teammates/alloc, sometimes
- [x] useQuery stuff doesn’t handle errors (Made sep issue #1022 )
- [x] remove console logs
- [x] need E2E tests for allocation (this already existed durrr)
- [x] write unit test for calculation (Made sep issue #1026)
- [x] fix ProfileCard.test.tsx
- [x] unable to run cy:dev tests
- [x] remove magic numbers in styling (deferred some of this because its throwaway work wrt 1.5)
- [x] use design tokens instead of hardcoded numbers  (deferred some of this because its throwaway work wrt 1.5)


CryptoGraffe:
@reeserj number 5 is fixed. number 7 and 4 are unrelated to this change as far as I can tell. Can you re-test?

CryptoGraffe:
@levity this is ready for you

levity (approved):

PR #1016, opened 2022-06-08
Set CI devchain port to 8555

topocount:
This is residual work from separating out our local CI environment from
the local dev environment. We should be able to have two separate
devchain processes running simultaneously to support each environment.

The project tags were removed from the ci.sh file because they
complicate setup and teardown in docker compose commands and dont really
add much

test plan

yarn cy:run

ensure your local devchain isn't queried by the integration tests at
any point.


levity:
> The project tags were removed from the ci.sh file because they complicate setup and teardown in docker compose commands and dont really add much

I added these because when I tried to do `docker compose down` without them, I got errors because it tried to take down the network that my dev containers were still using

package.json
```diff
@@ -94,7 +94,7 @@
     "hardhat:ganache": "./hardhat/scripts/start-ganache.sh",
     "hardhat:test": "yarn --cwd hardhat test",
     "test": "yarn hardhat:ganache --exec craco test --runInBand",
-    "test:ci": "yarn hardhat:ganache --exec ./scripts/ci.sh",
+    "test:ci": "CI=true yarn hardhat:ganache --exec ./scripts/ci.sh",
```
crabsinger, line 97: after the incident with vercel, do we want to change this to something besides CI ? @topocount

levity:
@topocount i'm in favor of keeping the project tags. in addition to helping the cy:dev command exit cleanly, it also creates a separation between the two sets of containers in the Docker dashboard UI. maybe that's still not a ton of benefit, but i also don't see what the downside is.

topocount:
> @topocount i'm in favor of keeping the project tags. in addition to helping the cy:dev command exit cleanly, it also creates a separation between the two sets of containers in the Docker dashboard UI. maybe that's still not a ton of benefit, but i also don't see what the downside is. i'm happy to make the changes here if you agree.

Happy to keep them in in this case. I dont have a docker UI, but i'll add some convenience npm scripts to make CLI life easier.

package.json
topocount: @crabsinger @levity This would be a significant change, and I'd like to discuss this in depth about how we want to evolve environment configs mid to long term.

For starters, Vercel has environment variables (which we now use) to prevent the issue we ran into. The CI variable is meant to emulate the CI environment. That is the intent here, since ganache and cypress also can run in a local dev environment.

Rather than creating a monolithic variable, I'd rather have some canary var, that if detected in a staging or production build, will stop the build and error, like a stopship.

cypress/plugins/index.ts
```diff
@@ -14,6 +14,11 @@ export default (on, config) => {
   // `config` is the resolved Cypress config
   config = dotenvPlugin(config, { debug: true }, true);
   if (config.env.CI || process.env.CI) {
+    config = dotenvPlugin(
+      config,
+      { debug: true, path: '.ci.env', ovverride: true },
```
levity, line 19:
```suggestion
      { debug: true, path: '.ci.env', override: true },
```

levity (approved):

levity:
Looks like CI might be having trouble connecting to the new testchain... perhaps not everyone is using the new port

topocount:
@levity spent some time looking into this today since it's been open for so long and there's some funkiness in how cypress is generating a provider to mint the coins against that's a bit more work than is probably worth doing now. I'll close this for now and come back to this after vaults launch.

PR #1020, opened 2022-06-09
Submit Distribution page Updates

### LINKED ISSUE #813: add Gift Circle / Fixed Payments UI to Submit Distribution page
- when fixed payments are disabled
  - show only one vault selector & amount form
  - show the "fixed payments are disabled"  message in the place of the other form
- if the fixed payment token doesn't match any vault you control
  - form fields are disabled, vault dropdown shows "No vault"
  - button is Export CSV instead of Submit Distribution
- when fixed payments are enabled
  - show Gift Circle form & Fixed Payment form
    - the fixed payment's vault selector is disabled, since it's hardcoded to the fixed payment token type from circle settings
    - the budget amount input is also disabled and fixed to the sum of all users' fixed payments in the circle
  - when the fixed payment token is the same as the gift circle token
    - show only a single submit button, because there's only one transaction necessary
  - if the tokens are different
    - show two submit buttons
      - each one starts a separate transaction flow
    - after submission
      - button becomes a "View on Etherscan" button
      - display summary of distribution instead of form
      - 1 Backlink
        - same as after submission
- form validation for both forms
  - if the vault doesn't have enough funds for the specified amount, show error & disable button
- combined data table
  - show columns for both the gift circle and fixed payments column
  - if they're the same token, show a column of summed values
- backend data change
  - Store in the distribution table the type(s) it was for: gift circle, fixed payments, or both
    - because we can't just infer from the current fixed payment token type in circle settings -- that could change after distribution
    - store the gift amount and fixed amount
- loading the page after 1 or 2 distributions have been submitted
  - same as after submission
  - use the submitted distribution data for the forms and the data table

zashton:
Fixes #813 

- when fixed payments are disabled
  - [x] show only one vault selector & amount form
  - [x] show the "fixed payments are disabled"  message in the place of the other form
- if the fixed payment token doesn't match any vault you control
  - [x] form fields are disabled, vault dropdown shows "No vault"
  - [x] button is Export CSV instead of Submit Distribution
- when fixed payments are enabled
  - [x] show Gift Circle form & Fixed Payment form
  - [x] the fixed payment's vault selector is disabled, since it's hardcoded to the fixed payment token type from circle settings
  - [x] the budget amount input is also disabled and fixed to the sum of all users' fixed payments in the circle
  - when the fixed payment token is the same as the gift circle token
    - [x] show only a single submit button, because there's only one transaction necessary
  - if the tokens are different
    - [x] show two submit buttons
      - [x] each one starts a separate transaction flow
    - after submission
      - [x] button becomes a "View on Etherscan" button
      - [x] display summary of distribution instead of form
      - 1 Backlink
        - [x] same as after submission
- form validation for both forms
  - [x] if the vault doesn't have enough funds for the specified amount, show error & disable button
- combined data table
  - [x] show columns for both the gift circle and fixed payments column
  - [x] if they're the same token, show a column of summed values
- backend data change
  - [x] Store in the distribution table the type(s) it was for: gift circle, fixed payments, or both
    - because we can't just infer from the current fixed payment token type in circle settings -- that could change after distribution
    - [x] store the gift amount and fixed amount
- loading the page after 1 or 2 distributions have been submitted
  - [x] same as after submission
  - [x] use the submitted distribution data for the forms and the data table

- [x] figure out why unit test is failing @zashton 
- [ ] e2e tests to test out the above implementations (circle distribution/ fixed payment distribution/ combined distribution ) @topocount


zashton:
> @zashton do you want to split this out more? I'm happy to take the backend stuff, for example

Yeah could use your help , not sure why the contract calls are throwing errors. Can't work on those unchecked items in the PR desc that happen after submit distribution and the commented out contract calls to get the vault token balance in `DistributionForm.tsx`>`onChangeVault` also throwing contract errors

```
backend data change
 Store in the distribution table the type(s) it was for: gift circle, fixed payments, or both
because we can't just infer from the current fixed payment token type in circle settings -- that could change after distribution
 store the gift amount and fixed amount
 ```
 this should be done just can't test it yet because its post submission

**Comment by topocount** [comment]
> Any part that is unchecked in the desc you would want to work on?

@zashton seems like you're almost done! Happy to do the E2E tests on this, since the feature is such a big lift and there's a lot of logic to test out.

**Comment by zashton** [comment]
> > Any part that is unchecked in the desc you would want to work on?
> 
> @zashton seems like you're almost done! Happy to do the E2E tests on this, since the feature is such a big lift and there's a lot of logic to test out.

that will be great, i'm still trying to troubleshoot an issue contract dust error when doing combined distribution(fixed payment + circle alloc). Other than that i think its done.

File: src/lib/merkle-distributor/index.ts [hunk 1]
```diff
@@ -23,17 +26,36 @@ import { MerkleDistributorInfo, parseBalanceMap } from './parse-balance-map';
  */
 export const createDistribution = (
   gifts: Record<string, number>,
+  fixedGifts: Record<string, BigNumber> | undefined,
```
levity, line 29: can we do this work around fixed/gift distinctions at a higher level of the code? the stuff in `lib/merkle-distributor` is intended to stay general

levity: Yeah, that makes sense. How about this?

```diff
diff --git a/src/lib/merkle-distributor/index.ts b/src/lib/merkle-distributor/index.ts
index f5e64088..5839f6a9 100644
--- a/src/lib/merkle-distributor/index.ts
+++ b/src/lib/merkle-distributor/index.ts
@@ -26,8 +26,16 @@ export const createDistribution = (
   totalAmount: BigNumber,
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
+  const balances = balancesFromGifts(gifts, totalAmount);
+  return createDistributionFromBalances(balances, previousDistribution);
+};
+
+export const balancesFromGifts = (
+  gifts: Record<string, number>,
+  totalAmount: BigNumber
+) => {
   const totalGive = Object.values(gifts).reduce((t, v) => t + v);
-  let balances = Object.keys(gifts).map(address => ({
+  const balances = Object.keys(gifts).map(address => ({
     address,
     earnings: totalAmount.mul(gifts[address]).div(totalGive),
   }));
@@ -39,7 +47,13 @@ export const createDistribution = (
   const topBalance = assertDef(balances.find(x => x.address === topGift[0]));
   topBalance.earnings = topBalance.earnings.add(dust);
   assert(getDust(totalAmount, balances).eq(0), `dust is still not 0`);
+  return balances;
+};

+export const createDistributionFromBalances = (
+  balances: { address: string; earnings: BigNumber }[],
+  previousDistribution?: Partial<MerkleDistributorInfo>
+): MerkleDistributorInfo => {
   // add values from previous epoch
   if (previousDistribution) {
     assert(previousDistribution.claims, 'No claims found');

```

Then you could do this in useSubmitDistribution around line 84:

```js
const balances = balancesFromGifts(gifts, totalAmount);

// loop through balances and add fixed amounts

createDistributionFromBalances(balances, prev);
```

levity: Although-- you could also just leave it as-is if you want. I thought about it more and realized that the existing code is already not really that general, since it handles converting GIVE amounts based on a total.

src/lib/merkle-distributor/index.ts
```diff
@@ -23,17 +26,36 @@ import { MerkleDistributorInfo, parseBalanceMap } from './parse-balance-map';
...
   totalAmount: BigNumber,
+  fixedAmount: BigNumber,
```
levity, line 31: Remove this since it's unused

hasura/migrations/default/1654782109411_alter_table_public_distributions_add_column_distribution_type/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
+-- alter table "public"."distributions" add column "distribution_type" integer
+--  not null default '1';
```
levity, line 4: please add a `drop column` in the down migrations

topocount (commented):
A few QA issues I picked up on:

- [ ] The member page does reflect that a fixed payment token has been chosen until the page is refreshed
- [ ] A deposit isn't reflected in the distributions page until after the page is refreshed
- [ ] The Give distribution submission button is not disabled when the amount is set to zero and doesnt reflect insufficient balance until after it's clicked. (This feels like a bit of a nitpick)

zashton:
> This is some amazing work, @zashton
> 
> A few QA issues I picked up on:
> 
> * [ ]  The member page does reflect that a fixed payment token has been chosen until the page is refreshed
> * [ ]  A deposit isn't reflected in the distributions page until after the page is refreshed
> * [ ]  The Give distribution submission button is not disabled when the amount is set to zero and doesnt reflect insufficient balance until after it's clicked. (This feels like a bit of a nitpick)

issue 2 & 3 should be fixed now

topocount (approved):
ballin'! let's merge this puppy and i'll tackle E2E in a separate PR

topocount:
clutch rebase btw

PR #1030, opened 2022-06-11
changing all references to `window` to be client-side conditionalized

### LINKED ISSUE #1027: Remove or make async/conditional all access to `window`
NextJS code can run server side, and hence doesn't always have access to `window`. Change our dependency on these to be wrapped in useEffect or other async mechanisms.

CryptoGraffe:
Fixes #1027 by making sure we only reference `window`  when we are on the client-side. 

**NOTE:** After deploy, we need to make sure login and landing page (non-app) still work ok.

### DIFF
--- a/src/pages/VaultsPage/DepositModal.tsx
+++ b/src/pages/VaultsPage/DepositModal.tsx
@@ -33,7 +33,7 @@ export default function DepositModal({
   const [max, setMax] = useState<any>();
   const [submitting, setSubmitting] = useState(false);
   const contracts = useContracts();
-  (window as any).contracts = contracts;
+
   const [selectedContracts, setSelectedContracts] = useState<Contracts>();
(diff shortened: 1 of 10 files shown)


src/pages/VaultsPage/DepositModal.tsx
```diff
@@ -33,12 +33,18 @@ export default function DepositModal({
   const [max, setMax] = useState<any>();
   const [submitting, setSubmitting] = useState(false);
   const contracts = useContracts();
-  (window as any).contracts = contracts;
```
levity, line 36: Oops--this was just for debugging and can be removed entirely

levity (changes requested):

src/pages/VaultsPage/DepositModal.tsx
CryptoGraffe: Ok I deleted the related useEffect

levity (approved):

PR #1032, opened 2022-06-13
fix: Change styling only when is select

karelianpie:
## Motivation and Context

Organization Name is missing the grey background for it's text box when creating a circle, this was introduced as part of the work for https://github.com/coordinape/coordinape/pull/914 which required a new styling for the `ApeAutocomplete`

## Description

Only change the `ApeAutocomplete` styling when it's a select (in the map) 

## Test and Deployment Plan

Ran locally and saw the correct style being applied

## Screenshots (if appropriate):

Screen Shot 2022-06-13 at 4 52 31 pm

## Reviewers

@levity @crabsinger 

## Related Issue

Closes https://github.com/coordinape/coordinape/issues/1019

PR #1033, opened 2022-06-13
fix: Add isSelect prop to the select in map

karelianpie:
Follow up on https://github.com/coordinape/coordinape/pull/1032

Add missing `isSelect` prop


karelianpie:
@levity forgot to actually add the prop in the map select in https://github.com/coordinape/coordinape/pull/1032

levity (approved):

Issue #559, opened 2022-02-22
"Coordinape In Your Circle": Reinstate Coordinape member in admin view 


spydermonkee:
**Problem**
We've added a Coordinape user in all circles to enable people to give to Coordinape if they feel inspired as a token of thanks or just to support Coordinape. The problem is that we have not given our Coordinape Humans enough context as to what this is all about and many admins have simply deleted it. Others have ignored. A few have given to Coordinape! 

**Specification**
- If the coordinape user ( user.role == 2) has a non-null `deleted_at` value show the user in a DISABLED state since the admin previously deleted it already and we don't want them to become annoyed that i's back against their will.

Design: 
https://www.figma.com/file/Db8KthuZRjT5CfMvZsjTT2/Coordinape-Design-v7?node-id=2692%3A9277

**NOTE - this is dependent on #558**

reeserj:
@amrro would you want to pick this one up when you finish #558 ?

amrro:
Yes. I please assign it to me.

crabsinger:
https://github.com/coordinape/coordinape/pull/1042 resolved this

PR #1042, opened 2022-06-15
Add actions to enable/disable coordinape user, in Admin Page

### LINKED ISSUE #558: "Coordinape In Your Circle": ENABLE/DISABLE in members list in admin view

amrro:
Fixes #558 #559
____

In order to reinsstate the coordinape user, this creates a dummy coorindape user object and adds to the table in case of: 1) doesn't exist at all, 2) soft-deleted.
In case there are no users yet in the circle, coordinape user won't be shown.

Since `createUser` cannot add user with a `role` that is larger than 1, I used the coordinape's address as indication if this is the coordinape user or not.

Screen Shot 2022-06-16 at 12 38 11 AM

____

### Caveats
- In the seeding data, Coordinape doesn't exist in any circle. Make sure to add it manually beforehand. You can run: 
```gql
mutation coordinape {
  createUser(payload: {address: "0xfad763da9051953fea58c2304395719a3b6ba361", circle_id: 33, name: "Coordinape"}) {
    UserResponse {
      address
      name
    }
  }
}
```
Then update its `role` in the database manuallay to value `2`.


src/pages/AdminPage/AdminPage.tsx
```diff
@@ -201,6 +207,25 @@ const AdminPage = () => {
             : undefined
         }
       />
+      <ActionDialog
```
levity, line 210: you don't have to use this here; a simple `window.confirm` works as well

levity (changes requested):
Thanks for this, @amrro. Please note the spec in #558:

> For the enable, this would require a new hasura mutation that restores the user, mutation would set `deleted_at` back to a null value. 

It's important that the special role for the coordinape user is correctly set. Creating a user with the same address but the role of a normal user is going to confuse other code and analytics, so it's not an acceptable solution. Also, we don't want to create multiple deleted users with the same address if the Coordinape user is enabled & disabled multiple times.

Instead, please create a new mutation that searches for a user with non-null deleted_at and role = 2, and un-deletes them by setting deleted_at to null. All new circles have a Coordinape user created by default, so you can assume one always exists in the circle that you can undelete.

src/lib/zod/index.ts
```diff
@@ -258,6 +258,12 @@ export const updateCircleInput = z
   })
   .strict();
 
+export const restoreCoordinapeInput = z
```
levity, line 261: @topocount @zashton IMO these should be co-located with handlers when they're not used in multiple places; what do you think?

src/pages/AdminPage/components.tsx
```diff
@@ -832,23 +890,33 @@ export const ContributorsTable = ({
...
                       area="wide"
                       align="left"
+                      disabled={u.deleted_at !== null}
                     >
                       <UserName user={u} />
                     </Table.Cell>
 
-                    <Table.Cell key={`address-${u.id}`}>
+                    <Table.Cell
+                      key={`address-${u.id}`}
```
levity, line 899: i overlooked this in the previous review... all these `key` properties on Cell are unnecessary. Only the top-level `Row` tag needs one

src/lib/gql/mutations.ts
```diff
@@ -415,6 +415,26 @@ export async function updateTeammates(circleId: number, teammates: number[]) {
...
+            circle_id: circleId,
+          },
+        },
+        {
+          success: true,
+        },
+      ],
+    },
+    {
+      operationName: 'resotre_coordinape',
```
levity, line 433:
```suggestion
      operationName: 'restore_coordinape',
```

src/lib/gql/mutations.ts
```diff
@@ -415,6 +415,26 @@ export async function updateTeammates(circleId: number, teammates: number[]) {
   return updateTeammates;
 }
 
+export async function restoreCoordinpeUser(circleId: number) {
```
levity, line 418:
```suggestion
export async function restoreCoordinapeUser(circleId: number) {
```

src/hooks/useApiAdminCircle.ts
```diff
@@ -97,16 +97,24 @@ export const useApiAdminCircle = (circleId: number) => {
     [circleId]
   );
 
+  const restoreCoordinape = useRecoilLoadCatch(
+    () => async (circleId: number) => {
+      await mutations.restoreCoordinpeUser(circleId);
```
levity, line 102:
```suggestion
      await mutations.restoreCoordinapeUser(circleId);
```

hasura/metadata/actions.yaml
```diff
@@ -112,6 +112,15 @@ actions:
       value_from_env: HASURA_EVENT_SECRET
   permissions:
   - role: user
+- name: restoreCoordinape
+  definition:
+    kind: synchronous
+    handler: '{{HASURA_API_BASE_URL}}/actions/restoreCoordinape'
```
levity, line 118: we recently changed the way we add handlers (see #1046 for context); please follow the new pattern

src/pages/AdminPage/components.tsx
```diff
@@ -887,14 +964,28 @@ export const ContributorsTable = ({
...
-                          ? () => setDeleteUserDialog(u)
-                          : undefined
-                      )}
+                      {u.role === USER_ROLE_COORDINAPE
+                        ? renderCoordinapeActions(u.deleted_at === null, () => {
+                            const shouldEnable = u.deleted_at !== null;
+                            const confirm = window.confirm(
+                              `${
+                                shouldEnable ? 'Enable' : 'Disable'
+                              } Coordinape in this ${circle.name}?`
```
levity, line 973:
```suggestion
                              } Coordinape in this circle?`
```

hasura/metadata/actions.graphql
```diff
@@ -103,7 +107,6 @@ input CreateUserInput {
   non_receiver: Boolean
   role: Int
   starting_tokens: Int
```
levity, line 109:
```suggestion
  starting_tokens: Int
  fixed_payment_amount: Float
```

hasura/metadata/actions.graphql
```diff
@@ -125,7 +128,6 @@ input AdminUpdateUserInput {
   fixed_non_receiver: Boolean
   non_receiver: Boolean
   role: Int
```
levity, line 130:
```suggestion
  role: Int
  fixed_payment_amount: Float
```

src/pages/AdminPage/components.tsx
```diff
@@ -291,19 +295,64 @@ export const TableLink = styled(RouterLink, {
...
+        flexDirection: 'column',
+      }}
+    >
+      <Text size="large" css={{ my: '$md' }}>
+        Why is Coordinape in your circle?
+      </Text>
+      <p>
+        We&apos;re experimenting with the gift circle mechanism as our revenue
+        model. By default, Coordinape appears in your circle and any user can
+        allocate to Coordinape. To remove the Coordinape user, click the trash
```
levity, line 313: Change this from "To remove..." to "To show or hide the Coordinape user, use the links on the right side of this row."

src/pages/AdminPage/components.tsx
```diff
@@ -832,23 +890,33 @@ export const ContributorsTable = ({
...
                       align="left"
+                      disabled={u.deleted_at !== null}
                     >
                       <UserName user={u} />
                     </Table.Cell>
 
-                    <Table.Cell key={`address-${u.id}`}>
+                    <Table.Cell
+                      key={`address-${u.id}`}
+                      disabled={u.deleted_at !== null}
```
levity, line 900: I'm bikeshedding here... but you could have avoided adding all these `disabled` lines and put it in the Row tag once instead.

levity (changes requested):

src/pages/AdminPage/components.tsx
amrro: Pardon me. I tried to add it to the whole row. But we have a single column that is not disabled. And a child cannot have higher opacity than its parents, and workaround would make it complex.

levity: Ah, sorry, I overlooked the non-disabled cell. How about something like:

```
<Row css={isCoordinapeUser ? {
  '> td': { opacity: 0.3 },
  '> td.normal': { opacity: 1 }
}: null}>
  <Cell>a</Cell>
  <Cell className="normal">b</Cell>
</Row>
```

spydermonkee:
@amrro - Hey there!
What's the latest status on this PR? Is it still waiting for review, or are there still requested changes that need to be addressed?

api/hasura/actions/_handlers/restoreCoordinape.ts
```diff
@@ -0,0 +1,64 @@
...
+      update_users_by_pk: [
+        {
+          pk_columns: { id: existingCoordinape.id },
+          _set: { deleted_at: null },
+        },
+        { __typename: true },
+      ],
+    },
+    {
+      operationName: 'resotre_coordinape',
```
levity, line 55:
```suggestion
      operationName: 'restore_coordinape',
```

levity:
> @amrro - Hey there! What's the latest status on this PR? Is it still waiting for review, or are there still requested changes that need to be addressed?

I'll review this now

levity:
I've merged this into a copy of your branch in our repo, so that we can run CI on it. (A fix so that CI can run on fork branches is coming soon.)

PR #1046, opened 2022-06-16
Unite actions behind a central hander

topocount:
## Motivation and Context

Our builds our slow right now (circa 15-22 minutes).
[This preview build](https://vercel.com/coordinape/coordinape/7QCDu4WKwF5fFpwJ3iwyy61ekN4B) completes in 10 minutes, without 
the benefit of the build cache for the new handler bundle, even, and ~5 minutes with the benefit of the cache.

There are still further optimizations to be made as well,
as far as bundling and routing different types of requests
into a single endpoint.

## Test Plan

E2E tests are passing so it appears that all endpoints are functioning as expected.

## Merge Plan

merge after #1047 because it fixes an issue that will cause all endpoints to crash after combining them all.

### DIFF
--- a/api-lib/types.d.ts
+++ b/api-lib/types.d.ts
@@ -73,3 +73,7 @@ export interface EventTriggerPayload<
     name: string;
   };
 }
+
+export interface ActionPayload<G extends keyof GraphQLTypes> {
+  action: { name: G };
+}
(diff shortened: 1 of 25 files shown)


levity (commented):
i think we shouldn't return 200 when no handler's configured? it could mask errors. otherwise this looks great

PR #1047, opened 2022-06-16
Goerli prep and Vaults endpoint bugfix

topocount:
## Motivation and Context

Using hardhat dependencies via npm link or npm imports in our backend as local dependencies doesn't work in the vercel serverless environment.

## Description

This changes those imports to a relative path, which is a bit course and inelegant, but bypasses the complexity of configuring a local npm dependency. This can be reverted if/when we publish our hardhat dep on the registry or migrate to a monorepo.

## Test and Deployment Plan

Make sure vault deployments work in staging!

### DIFF
--- a/api-lib/config.ts
+++ b/api-lib/config.ts
@@ -61,6 +61,8 @@ export const HARDHAT_GANACHE_PORT: number = getEnvValue(
 
 export const ETHEREUM_RPC_URL: string = getEnvValue('ETHEREUM_RPC_URL');
 
+export const GOERLI_RPC_URL: string = getEnvValue('GOERLI_RPC_URL');
+
 // This is only set to 'production' when deployed from main branch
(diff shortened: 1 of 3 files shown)


levity (approved):
glad this worked!

api-lib/config.ts
```diff
@@ -61,6 +61,8 @@ export const HARDHAT_GANACHE_PORT: number = getEnvValue(
 
 export const ETHEREUM_RPC_URL: string = getEnvValue('ETHEREUM_RPC_URL');
 
+export const GOERLI_RPC_URL: string = getEnvValue('GOERLI_RPC_URL');
```
karelianpie, line 64: We should add this to the `.env.example` as it was failing to initialise without it, created the PR https://github.com/coordinape/coordinape/pull/1055 to add it

Issue #978, opened 2022-05-30
Integration Tests: Vault Withdrawal


topocount:
A circle admin can withdraw their funds from the vault.

See #960 for more detauls

## Things to Validate

- [ ] Check the local transaction histories and ensure they appear correctly for the withdrawal.
- [ ] withdrawal amounts are correct in the UI
- [ ] The activity is logged correctly in the Vault History Page

Issue #490, opened 2022-01-27
Build Withdraw from Vault


spydermonkee:
### Specification

Should send `ApeRouter.delegateWithdrawal` or `.removeTokens` depending on whether the vault has a Yearn vault token or a simple token

This design should mirror the deposit design #457. (However we allow users to deposit from a different wallet, they should also be able to withdraw to a different wallet.)
https://www.figma.com/file/wXF7xGVv1j2SqwWfgbamS8/Coordinape-Design-v8?node-id=2136%3A23479

Pending:
- [ ] decision on whether to support withdrawing to arbitrary wallets (to mirror being able to deposit from arbitrary wallets), and if so, getting smart contract changes for that

PR #1053, opened 2022-06-17
Implement Vault Withdraw modal

topocount:

withdraw

## Related Issues

#490 #978

### DIFF
--- a/src/pages/VaultsPage/VaultRow.tsx
+++ b/src/pages/VaultsPage/VaultRow.tsx
@@ -21,6 +21,7 @@ export function VaultRow({
   css?: CSS;
 }) {
   const [modal, setModal] = useState<ModalLabel>('');
+  const [userIsOwner, setUserIsOwner] = useState<boolean>(false);
   const [balance, setBalance] = useState(0);
   const closeModal = () => setModal('');
   const contracts = useContracts();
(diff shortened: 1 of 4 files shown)


src/pages/VaultsPage/VaultRow.tsx
```diff
@@ -45,6 +46,15 @@ export function VaultRow({
...
       ) : null}
+      {modal === 'withdraw' ? (
+        <WithdrawModal
+          visible={true}
+          onClose={closeModal}
+          vault={vault}
+          balance={balance}
+          onWithdraw={updateBalance}
+        />
+      ) : null}
```
topocount, line 57: should we change the modal render to switch/case?

levity: Good idea

src/pages/VaultsPage/WithdrawModal.tsx
```diff
@@ -0,0 +1,102 @@
...
+        css={{
+          position: 'relative',
+          display: 'flex',
+          flexDirection: 'column',
+          alignItems: 'center',
+          backgroundColor: 'white',
+          width: '100%',
+          padding: '0 0 $lg',
+          overflowY: 'auto',
+          maxHeight: '100vh',
```
topocount, line 74: get rid of maxHeight

src/pages/VaultsPage/VaultRow.tsx
```diff
@@ -34,17 +34,34 @@ export function VaultRow({
...
 
-  useBlockListener(updateBalance, [vault.id]);
+  const updateOwner = async () => {
+    const currentVault = contracts?.getVault(vault.vault_address);
+    if (!currentVault || !contracts) {
+      setUserIsOwner(false);
+      return;
+    }
+    const [ownerAddress, userAddress] = await Promise.all([
+      currentVault.owner(),
```
levity, line 44: even though updating the owner on every block is strictly speaking the correct thing to do, i think it's safe to assume it will change so infrequently (and can be handled by a page reload if it does) that it's a better use of resources to only query this once per vault.

Otherwise everything looks good here! 👍

src/pages/VaultsPage/VaultRow.tsx
```diff
@@ -36,15 +36,32 @@ export function VaultRow({
...
+      }
+      const [ownerAddress, userAddress] = await Promise.all([
+        currentVault.owner(),
+        contracts.getMyAddress(),
+      ]);
+      setUserIsOwner(ownerAddress.toLowerCase() === userAddress.toLowerCase());
+    };
+    updateOwner();
+  }, [contracts, vault.id]);
+
```
topocount, line 54: @levity it's easier to look at the net diff for your remaining comment than the commit diff

levity (approved):

PR #1054, opened 2022-06-17
Help Button

### LINKED ISSUE #1052: Floating Help Button on all pages
image

CryptoGraffe:
Implements a floating Help Button in the bottom right of screen. Based on work from the onboarding. 

Fixes #1052 

Test plan:
-Make sure it works ok on desktop and mobile
-Make sure the links work
-Check it out on a few different pages to make sure z-index and sizing/overlap stuff is ok


src/stitches.config.ts
```diff
@@ -85,6 +85,7 @@ export const colors = {
   surface: figmaColors.grey2,
   border: figmaColors.grey4,
   focusedBorder: figmaColors.grey8,
+  dividingLine: figmaColors.grey8,
```
levity, line 88: bikeshedding a bit here but maybe we should rename `focusedBorder` to `heavyBorder` and use that here instead of creating a new name

CryptoGraffe: The usage of "focusedBorder" is pretty all over the place. Nav link active underlines, external link icon color, users list on map scroll bar thumb, checkbox border, and then "focused text area border" , the only place where the name lives up. I think this is alright to rename but I don't love the kind of fake semantics here.

CryptoGraffe: I'm actually going to use `borderMedium` since this is room for a darker one and there is a patten here with `secondary` and `secondaryDark`

levity (approved):

PR #1092, opened 2022-06-29
Modify backend to store only offchain Vault Tx Info

topocount:
## Motivation and Context

We only want to store application info in our DB that isn't logged
onchain.

I added a deployment_block column to the vault table as well, for event filtering
optimization in the frontend.

## Test and Deployment Plan

Create a vault in the UI or in graphiql and then run this mutation
```gql
mutation MyMutation {
  createVaultTx(payload: {tx_hash: "0xa617c7146237bfdd510c4b24e96c9240bff7b624a4f4e43108cbde5ea0bc254a", tx_type: "Deposit", vault_id: 2}) {
    id
  }
}
```


src/lib/zod/index.ts
```diff
@@ -371,6 +419,22 @@ type InputSchema<T extends z.ZodRawShape> =
...
+    input: z.object({ payload: z.union(inputSchema) }),
+    action: z.object({ name: z.string() }),
+    session_variables: z.union([
+      HasuraAdminSessionVariables,
+      HasuraUserSessionVariables,
+    ]),
+    request_query: z.string().optional(),
+  });
+}
+
```
topocount, line 437: this needs to be deleted. Was a failed experiment

api/hasura/actions/_handlers/createVaultTx.ts
```diff
@@ -0,0 +1,69 @@
...
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    session_variables: { hasuraProfileId },
+    input: { payload },
+  } = composeHasuraActionRequestBodyWithSession(
+    VaultLogInputSchema,
+    HasuraUserSessionVariables
+  ).parse(req.body);
+  const actionToLog = VaultLogUnionSchema.parse(payload);
```
zashton, line 26: seems to be missing a validation whether the `hasuraProfileId` has an user account in the `organisation` that created this vault ?

api/hasura/actions/_handlers/createVaultTx.ts
```diff
@@ -0,0 +1,69 @@
...
+  switch (actionToLog.tx_type) {
+    case 'Deposit': {
+      const result = await logVaultTx({
+        ...actionToLog,
+        created_by: hasuraProfileId,
+        tx_type: vault_tx_types_enum.Deposit,
```
zashton, line 32: might be able to shorten the code if you do something like this
```ts
created_by: hasuraProfileId,
tx_type: (() => {
      switch (actionToLog.tx_type) {
        case 'Deposit':
          return vault_tx_types_enum.Deposit;
        case 'Withdraw':
          return vault_tx_types_enum.Withdraw;
        case 'Distribution':
          return vault_tx_types_enum.Distribution;
    }})()
```

api/hasura/actions/_handlers/createVaultTx.ts
```diff
@@ -0,0 +1,69 @@
...
+  composeHasuraActionRequestBodyWithSession,
+  HasuraUserSessionVariables,
+} from '../../../../src/lib/zod';
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    session_variables: { hasuraProfileId },
+    input: { payload },
+  } = composeHasuraActionRequestBodyWithSession(
+    VaultLogInputSchema,
```
zashton, line 23: seems to be missing validation whether `circle_id`,`vault_id` is under the `organisation`

api/hasura/actions/_handlers/createVaultTx.ts
```diff
@@ -0,0 +1,69 @@
...
+    HasuraUserSessionVariables
```
zashton, line 24: missing validation of `distribution_id` whether it belongs to an `epoch` under the `circle_id`

hasura/metadata/databases/default/tables/public_vault_transactions.yaml
```diff
@@ -2,48 +2,33 @@ table:
...
-    - description
-    - value
-    - date
-    - vault_id
-    - created_by
-    - created_at
+    - tx_type
     - updated_at
+    - vault_id
     filter:
```
zashton, line 32: `vaults` select permissions check whether `users` `deleted_at` `is_null` might want to be consistent here also

hasura/migrations/default/1656710520327_vault_tx_types/up.sql
```diff
@@ -0,0 +1,50 @@
+-- add deployment_block column
+alter table "public"."vaults" add column "deployment_block" bigint
+ not null;
+
+-- It's easier to drop and recreate the table wholesale
+DROP table "public"."vault_transactions";
```
zashton, line 6:
```
DROP table IF EXISTS "public"."vault_transactions";
```

src/hooks/useVaultRouter.ts
```diff
@@ -65,6 +66,13 @@ export function useVaultRouter(contracts?: Contracts) {
         chainId: contracts.chainId,
       }
     );
+    if (txResult?.tx)
```
zashton, line 69: is the db vault trx record just for display ? I think this part of the code will not execute if the user closes the window/tab before the transaction is mined ?

zashton (changes requested):
All working great, just some minor feedback regarding input validations and perms

src/hooks/useVaultRouter.ts
```diff
@@ -79,13 +87,23 @@ export function useVaultRouter(contracts?: Contracts) {
...
+      () => vaultContract.apeWithdraw(shares, underlying),
+      {
+        showError,
+        showInfo,
+        signingMessage: 'Please sign the transaction to withdraw tokens.',
+        chainId: contracts.chainId,
+        description: `Withdraw ${humanAmount} ${symbol}`,
+      }
+    );
+    if (txResult?.tx)
```
zashton, line 100: same feedback as above

api/hasura/actions/_handlers/createVaultTx.ts
topocount: ah, this is one of the few places where I actually like the idea of using an IIFE. Thanks!

src/hooks/useVaultRouter.ts
topocount: Yeah, it's just for display. Not sure of a better way to do this

zashton (approved):
looks good with the added validations

PR #1119, opened 2022-07-06
Fixed payment csv output

### LINKED ISSUE #811: add fixed payment column to CSV output
- only when fixed payments are enabled
- include token type in header column

Depends on #888

zashton:
fixes #811 

### Test Plan

1.  Choose a circle with an ended epoch and go to circle settings and set the fixed payment token type to any value
2.  Go to admin page and set a few members fixed payment token amounts
3.  Click on Export CSV button and check if the downloaded CSV whether the columns and rows are correct

### DIFF
--- a/api-lib/gql/__generated__/zeus/index.ts
+++ b/api-lib/gql/__generated__/zeus/index.ts
@@ -594,6 +594,7 @@ export type ValueTypes = {
     address: string;
     circle_id: number;
     fixed_non_receiver?: boolean | undefined | null;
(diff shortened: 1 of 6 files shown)


levity:
Yeah, I think this is ok for now. Makes sense to me that we would use the
setting at the moment of export. I could see people arguing for the
opposite but at least it’s not totally unreasonable to leave it as-is.

zashton:
we could do dropdown button in future `Export` + `Export with current settings`

PR #1123, opened 2022-07-11
Integrate Vault Transaction Logs

### LINKED ISSUE #476: Vaults List Page: Implement Recent Transactions
### Specification

- store history of deposits, withdrawals, & distributions in backend, & retrieve from that to display data in this table. 
- for distribution transactions, link to the page for that distribution

See schema migrations in #583

https://www.figma.com/file/wXF7xGVv1j2SqwWfgbamS8/App-1.0?node-id=3987%3A14226

topocount:
## Description

Both the VaultRow and All Transactions lists are implemented. Deposit,
Distribution and Withdraw events are captured and logged. Events
that do not have entries in our DB are not yet logged because we
need a better way to capture details from those txs or at least
communicate that the events are incomplete.

## Test and Deployment Plan

- deploy a vault and create deposits, distributions and withdraws
- wait a few seconds
- events should show up inline in the Vault transactions tables

image

## Related Issue

closes #476

Some remaining tasks I need to log:
- [x] Render Loading Components better
- implement E2E tests
- optimize the marshalling sequence a bit better if logs are
  slow to render on Goerli
- force the logs to re-render directly due to user interactions. It
  seems like deposits and withdrawals dont automatically submit a
  refetch request to useVaults yet.

### DIFF
--- a/src/lib/gql/queries.ts
+++ b/src/lib/gql/queries.ts
@@ -1,3 +1,6 @@
+import { order_by } from 'lib/gql/__generated__/zeus';
+import { allVaultFields } from 'lib/gql/mutations';
+
 import {
   IApiFullCircle,
   IApiManifest,
@@ -192,6 +195,85 @@ export const getDiscordWebhook = async (circleId: number) => {
     }
   );
   return circle_private.pop()?.discord_webhook;
+};
+
+export const getVaultAndTransactions = async (address: string | undefined) => {
+  const result = await client.query(
+    {
+      vaults: [
+        { where: { vault_address: { _eq: address } } },
+        {
+          ...allVaultFields,
+          vault_transactions: [
+            { order_by: [{ id: order_by.asc }] },
+            {
+              tx_hash: true,
+              tx_type: true,
+              created_at: true,
+              profile: {
+                address: true,
+                users: [{}, { circle_id: true, name: true }],
+              },
+              distribution: {
+                gift_amount: true,
+                fixed_amount: true,
+                epoch: {
+                  start_date: true,
+                  end_date: true,
+                  number: true,
+                  circle: { name: true },
+                },
+              },
+            },
+          ],
+          protocol: {
+            name: true,
+          },
+        },
+      ],
+    },
+    { operationName: 'getVault' }
+  );
+  const vault = result.vaults.pop();
+  if (!vault) throw new Error(`No Vault for address ${address}`);
+  return vault;
+};
+
+export const getCircle = async (circle_id: number) => {
+  const { circles_by_pk } = await client.query({
+    circles_by_pk: [
+      { id: circle_id },
+      {
+        id: true,
+        name: true,
+        logo: true,
+        default_opt_in: true,
+        is_verified: true,
+        alloc_text: true,
+        team_sel_text: true,
+        token_name: true,
+        vouching: true,
+        min_vouches: true,
+        nomination_days_limit: true,
+        vouching_text: true,
+        only_giver_vouch: true,
+        team_selection: true,
+        created_at: true,
+        updated_at: true,
+        protocol_id: true,
+        organization: {
+          id: true,
+          created_at: true,
+          name: true,
+          updated_at: true,
+        },
+        auto_opt_out: true,
+        fixed_payment_token_type: true,
+      },
+    ],
+  });
+  if (!circles_by_pk) throw new Error(`circle with id ${circle_id} not found`);
+  return circles_by_pk;
 };
 
 export const getFullCircle = async (
--- a/src/lib/gql/queries.ts
+++ b/src/lib/gql/queries.ts
@@ -1,6 +1,3 @@
-import { order_by } from 'lib/gql/__generated__/zeus';
-import { allVaultFields } from 'lib/gql/mutations';
-
 import {
   IApiFullCircle,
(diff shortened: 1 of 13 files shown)


src/pages/VaultsPage/VaultTransactions.tsx
```diff
@@ -1,62 +1,285 @@
...
+    isIdle,
+    data: vault,
+  } = useQuery(['vault', address], async () => {
+    const result = await getVaultAndTransactions(address);
+    return result;
+  });
 
-  if (!vault && !isLoading && !isIdle) {
+  useEffect(() => {
+    if (vault) getOnchainVaultTransactions(vault).then(setVaultTxList);
```
levity, line 33: you could wrap this call in useQuery, and then you would get caching and be able to avoid `useEffect`, `useState`, & `then`. same applies to `VaultRow`. this will make clicking through from VaultRow to the full list of transactions immediate too, so long as you use the same cache key

src/pages/VaultsPage/VaultTransactions.tsx
```diff
@@ -1,62 +1,285 @@
...
+  const provider = getProviderForChain(chain_id);
+  const contracts = new ContractsReadonly(chain_id, provider);
+  const depositEvents = getDepositEvents(contracts, vault);
+  const withdrawEvents = getWithdrawEvents(contracts, vault);
+  const distributionEvents = getDistributionEvents(contracts, vault);
+  const eventResults = await Promise.all([
+    depositEvents,
+    withdrawEvents,
+    distributionEvents,
+  ]);
```
levity, line 65:
```suggestion
  const eventResults = await Promise.all([
    getDepositEvents(contracts, vault),
    getWithdrawEvents(contracts, vault),
    getDistributionEvents(contracts, vault),
  ]);
```

i like this better because i was initially misled into thinking the calls were sync, not async; and the const names don't really add anything since they're only referenced shortly after and are very similar to the method names

src/utils/provider.ts
```diff
@@ -72,3 +78,20 @@ export function makeExplorerUrl(chainId: number, txHash: string | undefined) {
...
+  switch (chainId) {
+    case 1: // mainnet
+      return new InfuraProvider('homestead', INFURA_PROJECT_ID);
+    case 5: // Goerli
+      return new InfuraProvider('goerli', INFURA_PROJECT_ID);
+    case 1337:
+    case 1338:
+      return new JsonRpcProvider(
+        'http://localhost:' + HARDHAT_GANACHE_PORT,
+        1338
```
levity, line 92:
```suggestion
        chainId
```

src/pages/VaultsPage/VaultTransactions.tsx
```diff
@@ -1,62 +1,285 @@
...
+    deployment_block
+  );
+
+  const distributions: RawDistributionTx[] = [];
+  for (const event of distroEvents) {
+    const block = await event.getBlock();
+    const txDetails = vault_transactions.find(
+      tx => tx.tx_hash === event.transactionHash
+    );
+    const { name: circleName, id: circleId } = await getCircle(
```
levity, line 246: you could avoid an extra request here by adding these two fields to the VaultAndTransactions query. yes, then you're depending on the vault_transactions data rather than decoding the circle id from the event, but these should be the same anyway

src/lib/gql/queries.ts
```diff
@@ -194,6 +197,85 @@ export const getDiscordWebhook = async (circleId: number) => {
...
+      ],
+    },
+    { operationName: 'getVault' }
+  );
+  const vault = result.vaults.pop();
+  if (!vault) throw new Error(`No Vault for address ${address}`);
+  return vault;
+};
+
+export const getCircle = async (circle_id: number) => {
```
levity, line 242: remove this if you go with my suggestion below about not needing this

src/lib/gql/queries.ts
```diff
@@ -194,6 +197,85 @@ export const getDiscordWebhook = async (circleId: number) => {
   return circle_private.pop()?.discord_webhook;
 };
 
+export const getVaultAndTransactions = async (address: string | undefined) => {
```
levity, line 200: probably better to move this to its own file under `VaultsPage`. fits better with the pattern we've been establishing of queries being really fine-tuned to the needs of their calling pages, not to be reused in different contexts

hasura/metadata/databases/default/tables/public_users.yaml
```diff
@@ -25,6 +25,13 @@ array_relationships:
       table:
         name: burns
         schema: public
+- name: circle_api_keys
```
levity, line 28: seems like these changes ought to be in main already... ?

levity (changes requested):

src/lib/zod/index.ts
```diff
@@ -297,16 +297,64 @@ export const createVaultInput = z
...
+  .strict();
+
+const VaultLogEnum = z.enum([
+  'Deposit',
+  'Withdraw',
+  'Distribution',
+  'CircleBudget',
+]);
+
+export const VaultLogInputSchema = z.object({
```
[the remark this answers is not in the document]
topocount: This file is apparently only consumed by the backend. I feel like it should be moved to api-lib

levity (approved):

PR #1126, opened 2022-07-11
Vaults UX Improvements

teeolendo:
## Motivation and Context
Minor bug fixes around vault UX

## Description

- Changed the useVaults query to use chainID
~- Changed distribution page to show the summary page if trx_hash is set. Previously, a rejected on-chain transaction would show up as completed.~
- Changed the CreateForm to disable the loading icon if a vault is not created successfully

## Test and Deployment Plan

None are required.

### DIFF
--- a/src/pages/VaultsPage/CreateForm.tsx
+++ b/src/pages/VaultsPage/CreateForm.tsx
@@ -206,4 +206,5 @@ const AssetButton = styled(Button, {
     backgroundColor: '$secondaryText !important',
     '> span': { color: 'white !important' },
   },
+  length: 0,
 });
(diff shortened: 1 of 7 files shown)


src/pages/VaultsPage/CreateForm.tsx
```diff
@@ -206,4 +206,5 @@ const AssetButton = styled(Button, {
     backgroundColor: '$secondaryText !important',
     '> span': { color: 'white !important' },
   },
+  length: 0,
```
levity, line 209:
```suggestion
```

levity:
> Changed distribution page to show the summary page if trx_hash is set. Previously, a rejected on-chain transaction would show up as completed.

@teeolendo i don't see a change in this PR for this... ?

teeolendo:
> > Changed distribution page to show the summary page if trx_hash is set. Previously, a rejected on-chain transaction would show up as completed.
> 
> @teeolendo i don't see a change in this PR for this... ?

@levity you can ignore that comment for now. when I pulled from main, I saw the page had changed significantly so I left it out that change. Gonna circle back on a future PR this week.

levity (approved):

PR #1135, opened 2022-07-14
Preview values for allocation table

### LINKED ISSUE #1132: table on distributions page doesn't show preview values anymore
On the distributions page, when distributions have not yet been sent and the form is filled out, the table should show the values that will be used for the distribution when it's created. i.e. when I put 1000 DAI as a total amount for the gift circle, and someone has received half the GIVE for that epoch, one of the columns should show 500 DAI.

This was the behavior before #1020 was merged

zashton:
fixes #1132


levity:
The table needs to work in both cases--before & after a distribution

levity (approved):

Issue #1138, opened 2022-07-14
Allow editing current epoch


levity:
Follow-up to #1114 

The current epoch should also be editable, the same way that upcoming epochs are, except that you can't change the start date.

The edit button should go at the top right of the CurrentEpochPanel, pushing the Nominations & Allocations cards down.
<img width="605" alt="Coordinape" src="https://user-images.githubusercontent.com/193667/179033924-d9406599-8c11-4726-86c5-0a4f84069dce.png">

spydermonkee:
@levity , @sshmm - This seems like a high priority to correct since it's a significant product behavior regression. @sshmm , do you have bandwidth to prioritize this? @levity, do you have any thoughts about this?

levity:
> @levity , @sshmm - This seems like a high priority to correct since it's a significant product behavior regression. @sshmm , do you have bandwidth to prioritize this? @levity, do you have any thoughts about this?

If @sshmm has bandwidth today/tomorrow it should be a pretty quick fix, since it's very similar to what he just did for upcoming epochs. Otherwise I'll work on it tomorrow.

sshmm:
@spydermonkee I will start working on it in less than an hour

spydermonkee:
Awesome, thanks a lot @sshmm and @levity!

PR #1140, opened 2022-07-14
Allow Editing Current Epoch

sshmm:
## Motivation and Context

implements #1138 - Fixes an issue regarding the Editing Form where the form remains open if the user switched to another circle.

## Description
1- Add Edit button to current Epoch if the user is Admin.
2- Disable editing start date and start time in EpochForm for current Epoch.
3- The validation is between current epoch and future epochs if not empty
3- Bypass comparing between start date and current time for current epoch.

## Test and Deployment Plan
1- Tested editing current epoch with and without Available upcoming epochs.
2- Retested adding new epochs and editing next epochs.

Admin
image
image

Regular user
image

### DIFF
--- a/src/pages/HistoryPage/CurrentEpochPanel.tsx
+++ b/src/pages/HistoryPage/CurrentEpochPanel.tsx
@@ -3,7 +3,7 @@ import { NavLink } from 'react-router-dom';
 import { CSS } from 'stitches.config';
 
 import { paths } from 'routes/paths';
-import { Box, Panel, Text, Button } from 'ui';
+import { Box, Panel, Text, Button, Flex } from 'ui';
 import Medal from 'ui/icons/Medal.svg';
 import PlusInCircle from 'ui/icons/PlusInCircle.svg';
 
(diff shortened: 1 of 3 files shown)


crabsinger (approved):
lgtm

CryptoGraffe:
If you choose repeat and then get into an error state, choosing no-repeat and entering a valid day does not clear error state 
image

Also clicking the labels for the radio buttons selects no-repeat instead of the associated radio button. 

Both of these could be handled in separate tickets since this is somewhat urgent to get deployed.

sshmm:
> If you choose repeat and then get into an error state, choosing no-repeat and entering a valid day does not clear error state image
> 
> Also clicking the labels for the radio buttons selects no-repeat instead of the associated radio button.
> 
> Both of these could be handled in separate tickets since this is somewhat urgent to get deployed.

It validates on submit not on change and it will clear and submit or show new errors on clicking save

CryptoGraffe:
@sshmm ok this is ok for now but kinda weird experience. maybe would be best to clear error on edit if we can't do client side validation. not critical for this ticket

sshmm:
> @sshmm ok this is ok for now but kinda weird experience. maybe would be best to clear error on edit if we can't do client side validation. not critical for this ticket

You are right but the validation is done on the ient but you have to click submit first.
I can work on this tomorrow (today for me UTC+2) to make the experience better but may be in another issue to merge this one first

levity:
Thanks for the timely support on this, everyone!

spydermonkee:
Yes, thank you @sshmm and @CryptoGraffe (and @levity)!

PR #1144, opened 2022-07-15
Fix adding a new circle as a new user

sshmm:
## Motivation and Context

When a user who is not a part of a circle tries to create a new circle, they create button is not responsive

## Description

There was an issue regarding the input field of protocol(org) name
image
The control props was not added to the input field in the condition of a user is not part of an org.
image

## Test and Deployment Plan
Use a new wallet and try to add a circle

## Additional info
The form should be done by react hook-form but this is a quick fix for a current issue

### DIFF
--- a/src/pages/CreateCirclePage/CreateCirclePage.tsx
+++ b/src/pages/CreateCirclePage/CreateCirclePage.tsx
@@ -9,7 +9,7 @@ import CreateCircleForm from 'forms/CreateCircleForm';
 import { useApiWithProfile } from 'hooks';
 import { useMyProfile } from 'recoilState/app';
 import * as paths from 'routes/paths';
-import { Box, Button, Panel, Text, TextField, Tooltip } from 'ui';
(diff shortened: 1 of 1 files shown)


levity:
@crabsinger does this overlap with your fix?

sshmm:
> @crabsinger does this overlap with your fix?

#1147 is addressing a different issue if this is the one you are talking about

levity (approved):
LGTM as long as when you create a circle from an existing org, that still works

crabsinger:
@sshmm this still seems to be erroring for me (although it may be a separate issue than the one you fixed)...

On your branch in the Vercel preview, I go to https://coordinape-hg277k04e-coordinape.vercel.app/new-circle with a new address, and try to create a Circle, and I get this error:
Screen Shot 2022-07-18 at 10 00 11 AM

crabsinger (changes requested):
Doesn't seem to fully fix new addresses creating circles...

crabsinger (approved):
actually - seems this does resolve the bug!!!

PR #1145, opened 2022-07-15
Add 'Create Circle in New Organization' button to the overview page

wingfeatherwave:
Screen Shot 2022-07-15 at 1 19 43 PM

Resolves 1143

Add 'Create Circle in New Organization' button to the overview page, also add Overview header.  some flex positioning

Pretty simple PR 🎉


crabsinger (approved):
lgtm. but i think theres some bug or weirdness in the state when I go back and select "Add Circle" with an existing org, the Org doesn't prepopulate. But doesn't need to block this.

crabsinger (approved):

PR #1147, opened 2022-07-15
Fix state bug with create circle form

crabsinger:
This seems to resolve bug, but not sure if the "right" way to fix this issue.


wingfeatherwave (approved):
These work for me locally.... nice!

PR #1150, opened 2022-07-18
Distribution page UI design

### LINKED ISSUE #813: add Gift Circle / Fixed Payments UI to Submit Distribution page

### LINKED ISSUE #1122: Balance available doesn't update on distributions screen
**Summary:**

After a deposit the Available amount reflects the vault balance before
submitting a distribution

The balance should reflect the amount available in the vault after the
distribution is created

3. **Steps to recreate:**

  - create a vault in the org containing the sports circle
  - deposit some funds
  - create a distribution for the epoch in the sports circle
  - observe that the balance doesn't update until the page refreshes

Screenshots/video recordings of the issue are very helpful as well. Checking the browser's console for errors also provides developers with helpful debugging information. See [this guide](https://balsamiq.com/support/faqs/browserconsole/) for details on opening the develop console.

zashton:
fixes #813 
fixes #1122 

image

https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=4680%3A11831

### DIFF
--- a/src/lib/merkle-distributor/index.ts
+++ b/src/lib/merkle-distributor/index.ts
@@ -31,15 +31,13 @@ export const createDistribution = (
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
   let balances: Balance[] = [];
-  if (Object.keys(gifts).length) {
-    const totalGive = Object.values(gifts).reduce((t, v) => t + v);
-    Object.keys(gifts).map(address => {
-      balances.push({
-        address,
-        earnings: giftAmount.mul(gifts[address]).div(totalGive),
-      });
+  const totalGive = Object.values(gifts).reduce((t, v) => t + v, 0);
+  Object.keys(gifts).map(address => {
+    balances.push({
+      address,
+      earnings: giftAmount.mul(gifts[address]).div(totalGive),
     });
-  }
+  });
 
   if (fixedGifts) {
     Object.keys(fixedGifts).map(address => {
--- a/src/lib/merkle-distributor/index.ts
+++ b/src/lib/merkle-distributor/index.ts
@@ -30,11 +30,16 @@ export const createDistribution = (
   giftAmount: BigNumber,
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
-  const totalGive = Object.values(gifts).reduce((t, v) => t + v);
-  let balances = Object.keys(gifts).map(address => ({
-    address,
-    earnings: giftAmount.mul(gifts[address]).div(totalGive),
-  }));
+  let balances: Balance[] = [];
+  if (Object.keys(gifts).length) {
+    const totalGive = Object.values(gifts).reduce((t, v) => t + v);
+    Object.keys(gifts).map(address => {
+      balances.push({
+        address,
+        earnings: giftAmount.mul(gifts[address]).div(totalGive),
+      });
+    });
+  }
 
   if (fixedGifts) {
     Object.keys(fixedGifts).map(address => {
@@ -54,7 +59,10 @@ export const createDistribution = (
 
   // Failing this means we did bad math
   assert(dust.lt(20), `panic: dust too high: ${dust.toString()}`);
-  const topGift = assertDef(maxBy(Object.entries(gifts), x => x[1]));
+  const topGift =
+    Object.keys(gifts).length === 0 && fixedGifts
+      ? assertDef(maxBy(Object.entries(fixedGifts), x => x[1]))
+      : assertDef(maxBy(Object.entries(gifts), x => x[1]));
   const topBalance = assertDef(balances.find(x => x.address === topGift[0]));
   topBalance.earnings = topBalance.earnings.add(dust);
   assert(getDust(totalAmount, balances).eq(0), `dust is still not 0`);
--- a/src/lib/merkle-distributor/index.ts
+++ b/src/lib/merkle-distributor/index.ts
@@ -30,7 +30,7 @@ export const createDistribution = (
   giftAmount: BigNumber,
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
-  let balances: Balance[] = [];
+  const balances: Balance[] = [];
   const totalGive = Object.values(gifts).reduce((t, v) => t + v, 0);
   Object.keys(gifts).map(address => {
     balances.push({
--- a/src/lib/merkle-distributor/index.ts
+++ b/src/lib/merkle-distributor/index.ts
@@ -31,7 +31,7 @@ export const createDistribution = (
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
   const totalGive = Object.values(gifts).reduce((t, v) => t + v, 0);
-  const balances = Object.keys(gifts).map(address => ({
+  let balances = Object.keys(gifts).map(address => ({
     address,
     earnings: giftAmount.mul(gifts[address]).div(totalGive),
   }));
--- a/src/lib/merkle-distributor/index.ts
+++ b/src/lib/merkle-distributor/index.ts
@@ -30,14 +30,11 @@ export const createDistribution = (
   giftAmount: BigNumber,
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
-  const balances: Balance[] = [];
   const totalGive = Object.values(gifts).reduce((t, v) => t + v, 0);
-  Object.keys(gifts).map(address => {
-    balances.push({
-      address,
-      earnings: giftAmount.mul(gifts[address]).div(totalGive),
-    });
-  });
+  const balances = Object.keys(gifts).map(address => ({
+    address,
+    earnings: giftAmount.mul(gifts[address]).div(totalGive),
+  }));
 
   if (fixedGifts) {
     Object.keys(fixedGifts).map(address => {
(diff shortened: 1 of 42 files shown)


src/lib/merkle-distributor/index.ts
```diff
@@ -30,11 +30,16 @@ export const createDistribution = (
   giftAmount: BigNumber,
   previousDistribution?: Partial<MerkleDistributorInfo>
 ): MerkleDistributorInfo => {
-  const totalGive = Object.values(gifts).reduce((t, v) => t + v);
-  let balances = Object.keys(gifts).map(address => ({
-    address,
-    earnings: giftAmount.mul(gifts[address]).div(totalGive),
-  }));
+  let balances: Balance[] = [];
```
levity, line 33: wrapping this whole thing in an if-clause and using `push` isn't necessary--all you need to do is add a default value to `reduce`:
```js
  const totalGive = Object.values(gifts).reduce((t, v) => t + v, 0);
```

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
-            {isCombinedDistribution() ? (
-              <span>
+            </Box>
+          </TwoColumnLayout>
+        </Panel>
+        {(fixedDist || circleDist) && <Summary distribution={circleDist} />}
+        <Flex css={{ justifyContent: 'center', mt: '$xl', mb: '$xl' }}>
+          {!circleDist ? (
+            isCombinedDistribution() ? (
+              <Text css={{ fontSize: '$small', lineHeight: '$xtaller' }}>
```
levity, line 407: please use padding or fixed-height for the div here, instead of adding this one-off token `$xtaller` to the stitches theme

src/stitches.config.ts
```diff
@@ -165,6 +165,7 @@ export const {
       tall: 1.625,
       tall2: 1.75,
       taller: '2',
+      xtaller: '3.5',
```
levity, line 168: remove

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
   );
 }
+
+const Summary = ({
+  distribution,
+}: {
+  distribution: EpochDataResult['distributions'][0] | undefined;
+}) => {
+  return (
+    <Flex
```
levity, line 631: if `!distribution`, should we just return null instead of an empty Flex?

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
+            ) : (
+              <Button
+                type="button"
+                color="primary"
+                outlined
+                size="large"
+                fullWidth
+                onClick={async () => {
+                  // use the authed api to download the CSV
+                  if (epoch.number) {
```
levity, line 600: if this doesn't work unless an epoch number is present, we probably just shouldn't show the button at all if it's missing. i see it was like this already before your changes, i should have caught it earlier

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
+                          )}
+                        </>
+                      )}
+                    />
+                  </FormControl>
+                </Box>
+                <Box css={{ width: '100%' }}>
                   <Controller
-                    name="selectedVaultId"
+                    name={'amount'}
```
levity, line 520:
```suggestion
                    name="amount"
```

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
                 fullWidth
               >
                 {sufficientGiftTokens
                   ? submitting
-                    ? 'Submitting...'
-                    : 'Submit Distribution'
-                  : 'Insufficient Tokens'}
+                    ? `Submitting...`
+                    : `Submit ${giftVaultSymbol} Vault Distribution`
+                  : `Insufficient Tokens`}
```
levity, line 424:
```suggestion
                    ? 'Submitting...'
                    : `Submit ${giftVaultSymbol} Vault Distribution`
                  : 'Insufficient Tokens'}
```

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
+                      const v = findVault({ symbol: giftVaultSymbol });
+                      if (v) return v.decimals;
+                      return 0;
+                    })()}
+                    type="number"
+                    error={!!error}
+                    value={circleDist ? circleDist.gift_amount : value}
+                    disabled={submitting || !!circleDist || vaults.length === 0}
+                    max={Number(maxGiftTokens)}
+                    prelabel={'Budget Amount'}
```
levity, line 381:
```suggestion
                    prelabel="Budget Amount"
```

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
+                      )}
+                    </>
+                  )}
+                />
+              </FormControl>
+            </Box>
+            <Box css={{ width: '100%' }}>
               <Controller
-                name="selectedVaultId"
+                name={'amount'}
```
levity, line 362:
```suggestion
                name="amount"
```

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -292,259 +300,374 @@ export function DistributionForm({
...
-            <FormControl fullWidth>
-              <Box css={vaultInputStyles}>Select Vault</Box>
+        <Panel css={{ padding: '16px', minHeight: '147px' }}>
+          <Text h2 css={headerStyle}>
+            Gift Circle
+          </Text>
+          <TwoColumnLayout css={{ pt: '$md' }}>
+            <Box css={{ width: '100%' }}>
+              <FormControl fullWidth>
+                <Controller
```
levity, line 312: just FYI, react-hook-form also has a `useController` hook that you can use instead of this component. I think it's nicer because it produces less nesting. no need to change it in this PR though.

src/pages/DistributionsPage/AllocationsTable.tsx
```diff
@@ -51,53 +66,106 @@ export const AllocationsTable = ({
...
+          justifyContent: 'space-between',
+          mt: '$lg',
+        }}
+      >
+        <Link
+          css={{ color: '$primary' }}
+          target="_blank"
+          href="https://docs.coordinape.com/get-started/compensation/paying-your-team"
+        >
+          Paying Your Team Documentation
```
levity, line 165:
```suggestion
          Documentation: Paying Your Team 
```

levity (changes requested):
Looking good so far! See below.

src/pages/DistributionsPage/DistributionForm.tsx
zashton: created an issue for this
https://github.com/coordinape/coordinape/issues/1168

levity:
@zashton there are still a few more open comments from my review; maybe you missed them because DistributionForm's diff is collapsed by default in Github?

src/pages/DistributionsPage/DistributionForm.tsx
zashton: it would look like this if its a non combined distribution without the Flex box

image

src/pages/DistributionsPage/DistributionForm.tsx
zashton: i've updated it to not show but had to leave the `if (epoch.number)` inside the onclick not sure why typescript is showing an issue

src/pages/DistributionsPage/DistributionForm.tsx
zashton: > @zashton there are still a few more open comments from my review; maybe you missed them because DistributionForm's diff is collapsed by default in Github?

yeah my bad for missing, some of the review got hidden by default. Have added the changes

PR #1152, opened 2022-07-18
Default name to '', but not id.

crabsinger:
bug debugging


crabsinger:
Doesn't seem to work - still getting this error:

PR #1153, opened 2022-07-19
Fix/create circle bug

crabsinger:
## Motivation and Context

Creating a new circle as a new user throws this error:

image


src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -42,7 +42,7 @@ export const SummonCirclePage = () => {
 
   const org = protocols.find(p => p.id === Number(params.get('org')));
   const source = {
-    protocol_id: org?.id || '',
+    protocol_id: org?.id || undefined,
```
levity, line 45: `|| undefined` is redundant here

levity: `undefined !== null` but they're both falsy

PR #1166, opened 2022-07-21
Fix api-user permissions for historical token gifts

META-DREAMER:
Allows API keys with read epochs + read allocations permission to read historical token gifts.

## Motivation and Context

Users of the API were not able to get historical allocation data.

## Description

Adds Hasura permissions for api-user role to read from token_gifts table.

## Test and Deployment Plan

Fetch token gifts using api-user role.


zashton (approved):
looks good to me !

PR #1167, opened 2022-07-22
Opt In Quality of Life Improvements

### LINKED ISSUE #1162: Member Self Opt Out Fix
### **Problem**
On the allocation page, members are accidentally opting themselves out since it is the first option on the page. 

### **Solution**
Opt-In should be the first option on the allocation page and preselected for the user. 

### **Specification**
On the Allocation page, the options need to be flipped. It should appear as follows. 

**Example**
Should you receive GIVV distributions in the Apes Circle during this epoch?
(Selected) Opt-In
I want to be eligible to receive GIVV

Opt-Out
I am paid sufficiently via other channels.

### **Acceptance Criteria** 
Scenario: User wants to allocate
Given: The user is on the allocate page
Then: Opt-In should be selected

### **Stakeholders**
@spydermonkee

topocount:
These changes make it easier for users to opt in and for circle creators
to onboard a circle faster.

First, created users used to be opted out upon creation, and had to opt
in during their first epoch to begin receiving GIVE. Now they are
opted in by default in our app, and there is no longer this chicken and
egg issue of users opting in and having no one to GIVE to because others
are not opted in.

Second, the UI now has the Opt-in radio button selected on every page
load. This makes it easier to opt in, and the user is now required to
explicitly opt out before receiving give.

Test plan:

E2E tests should still pass and unit tests still pass

Related Issue:
closes #1162

### DIFF
--- a/src/pages/AllocationPage/AllocationEpoch.tsx
+++ b/src/pages/AllocationPage/AllocationEpoch.tsx
@@ -133,7 +133,6 @@ const AllocationEpoch = ({
   useEffect(() => {
     if (myUser) {
       setEpochBio(myUser?.bio ?? '');
-      setNonReceiver(myUser.non_receiver);
     }
   }, [myUser]);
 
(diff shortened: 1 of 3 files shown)


src/pages/AllocationPage/AllocationEpoch.tsx
```diff
@@ -133,7 +133,6 @@ const AllocationEpoch = ({
   useEffect(() => {
     if (myUser) {
       setEpochBio(myUser?.bio ?? '');
-      setNonReceiver(myUser.non_receiver);
```
levity, line 136: Wait. Does this mean if I've opted out and I return to this page, it will show me as opted in?

PR #1171, opened 2022-07-25
App UI Icons

smombartz:
Created figma folder with App UI icons. As per conversation with Graffe.


CryptoGraffe (changes requested):
@smombartz Can you drop the extra `icon` nested folder and also name these consistently, e.g. `WalletConnect` is capitalized and `coinbase wallet` has a space in it whereas other ones are lowercase and use dashes for multi-word. `discrod` is misspelled.

CryptoGraffe:
Thanks for doing this, it's gonna be great when we get it all hooked up

CryptoGraffe:
> @smombartz Can you drop the extra `icon` nested folder and also name these consistently, e.g. `WalletConnect` is capitalized and `coinbase wallet` has a space in it whereas other ones are lowercase and use dashes for multi-word. `discrod` is misspelled.

FWIW I'm asking for this now because I'm assuming these names are in Figma and it would be great to get it right now instead of accidentally reintroduce diff names later

smombartz:
> > @smombartz Can you drop the extra `icon` nested folder and also name these consistently, e.g. `WalletConnect` is capitalized and `coinbase wallet` has a space in it whereas other ones are lowercase and use dashes for multi-word. `discrod` is misspelled.
> 
> FWIW I'm asking for this now because I'm assuming these names are in Figma and it would be great to get it right now instead of accidentally reintroduce diff names later

totally! good catch. i was moving fast and we definitely want them to be consistent!!

smombartz:
@CryptoGraffe made the requested changes. should all be nice and tidy now.

CryptoGraffe (approved):

PR #1173, opened 2022-07-26
Vault transactions etherscan fix

zashton:
Fixes

https://www.notion.so/coordinape/226b9363ee0d45cead60e5365ae65d10?v=2137f63c15394b32b90fb80dd09a5244&p=89cecf1fe5b64377ae2f0065d40f0808&pm=s

https://www.notion.so/coordinape/226b9363ee0d45cead60e5365ae65d10?v=2137f63c15394b32b90fb80dd09a5244&p=dccbe9c08ae148ea9b1b7db3b45c84ab&pm=s

### DIFF
--- a/src/pages/VaultsPage/VaultRow.tsx
+++ b/src/pages/VaultsPage/VaultRow.tsx
@@ -133,7 +133,7 @@ export function VaultRow({
           'Loading...'
         ) : vaultTxList?.length ? (
           <TransactionTable
-            chain_id={vault.chain_id}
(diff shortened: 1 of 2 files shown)


levity (changes requested):
please avoid using `snake_case` in JS -- use `camelCase` instead. i know it's mixed because of the fields that we get from Hasura via Zeus, but let's limit the use of snake case to those fields and stick with camel case everywhere else

levity (approved):

PR #1185, opened 2022-07-28
Claims Page

crabsinger:
## Motivation and Context

TODO:
- [ ] test wrapped amounts are dispalyed correctly
- [x] Verify mutations and permissions are correct

Update Claims Page to properly group, sort, and batch claims.

image

Screen Shot 2022-08-03 at 2 43 54 PM

Fixes:
- grouping
- etherscan link
- amount show
- date sorting

### DIFF
--- a/src/pages/ClaimsPage/ClaimsPage.tsx
+++ b/src/pages/ClaimsPage/ClaimsPage.tsx
@@ -189,15 +189,17 @@ export default function ClaimsPage() {
                 </Flex>
               </td>
               <td>
-                {formatEpochDates(
-                  claimedClaimsGroupByRow[claimsRowKey(distribution, txHash)]
-                )}
+                <Text>
+                  {formatEpochDates(
+                    claimedClaimsGroupByRow[claimsRowKey(distribution, txHash)]
+                  )}
+                </Text>
               </td>
               <td>
                 <Flex css={{ justifyContent: 'flex-end' }}>
                   <Flex
                     css={{
-                      minWidth: '10vw',
+                      minWidth: '11vw',
                       justifyContent: 'flex-end',
                       gap: '$md',
                       mr: '$md',
@@ -215,10 +217,10 @@ export default function ClaimsPage() {
                       color="primary"
                       outlined
                       css={{
-                        fontWeight: '$normal',
+                        fontWeight: '$medium',
                         minHeight: '$xs',
                         px: '$sm',
-                        minWidth: '5vw',
+                        minWidth: '15vw',
                         borderRadius: '$2',
                       }}
                       as="a"
@@ -287,10 +289,10 @@ const ClaimRow = ({
               color="primary"
               outlined
               css={{
-                fontWeight: '$normal',
+                fontWeight: '$medium',
                 minHeight: '$xs',
                 px: '$sm',
-                minWidth: '5vw',
+                minWidth: '11vw',
                 borderRadius: '$2',
               }}
               onClick={onClickClaim}
--- a/src/pages/ClaimsPage/ClaimsPage.tsx
+++ b/src/pages/ClaimsPage/ClaimsPage.tsx
@@ -16,7 +16,12 @@ import { makeExplorerUrl } from 'utils/provider';
 import { getClaims, QueryClaim } from './queries';
 import { useClaimAllocation } from './useClaimAllocation';
 
-const currentClaims = (claims: QueryClaim[]) =>
+// claimRows: reduce all claims into one row per group of {vault, circle,
+// isClaimed} for display in UI (where we group all claims into one row per set
+// of claimable claims together. If you can claim them all together, display
+// them as one row. If they were claimed all together, display them as one row
+// with link to the claim on etherscan)
+const claimRows = (claims: QueryClaim[]) =>
   claims
     .sort(c => c.id)
     .reduce(
@@ -27,8 +32,7 @@ const currentClaims = (claims: QueryClaim[]) =>
               curr.distribution.vault.vault_address &&
             c.distribution.epoch.circle?.id ===
               curr.distribution.epoch.circle?.id &&
-            c.distribution.vault.simple_token_address ===
-              curr.distribution.vault.simple_token_address
+            c.txHash === curr.txHash
         ).length > 0
           ? finalClaims
           : [...finalClaims, curr],
@@ -128,7 +132,7 @@ export default function ClaimsPage() {
             { title: 'Epochs', css: styles.th },
             { title: 'Rewards', css: styles.thLast },
           ]}
-          data={currentClaims(claims.filter(c => !c.txHash))}
+          data={claimRows(claims.filter(c => !c.txHash))}
           startingSortIndex={2}
           startingSortDesc
           sortByColumn={() => {
@@ -158,7 +162,7 @@ export default function ClaimsPage() {
             { title: 'Epochs', css: styles.th },
             { title: 'Rewards', css: styles.thLast },
           ]}
-          data={currentClaims(claims.filter(c => c.txHash))}
+          data={claimRows(claims.filter(c => c.txHash))}
           startingSortIndex={2}
           startingSortDesc
           sortByColumn={() => {
--- a/src/pages/ClaimsPage/ClaimsPage.tsx
+++ b/src/pages/ClaimsPage/ClaimsPage.tsx
@@ -17,13 +17,13 @@ import { getClaims, QueryClaim } from './queries';
 import { useClaimAllocation } from './useClaimAllocation';
 
 // claimRows: reduce all claims into one row per group of {vault, circle,
-// isClaimed} for display in UI (where we group all claims into one row per set
-// of claimable claims together. If you can claim them all together, display
-// them as one row. If they were claimed all together, display them as one row
+// txHash}, for representing a group of claims into one row per set
+// of batch claimable claims. If you can claim them all together, display
+// them all together. If they were claimed in same tx, display them as one row
 // with link to the claim on etherscan)
 const claimRows = (claims: QueryClaim[]) =>
   claims
-    .sort(c => c.id)
+    .sort(c => -c.id)
     .reduce(
       (finalClaims, curr) =>
         finalClaims.filter(
@@ -86,9 +86,18 @@ export default function ClaimsPage() {
       </SingleColumnLayout>
     );
 
-  const claimsGroupByVault = groupBy(
-    claims.sort(c => c.id),
-    c => c.distribution.epoch.circle?.id && c.distribution.vault.vault_address
+  const claimedClaimsGroupByRow = groupBy(
+    claims.filter(c => c.txHash).sort(c => -c.id),
+    c => {
+      return `${c.distribution.vault.vault_address} ${c.distribution.epoch.circle?.id} ${c.txHash}`;
+    }
+  );
+
+  const unclaimedClaimsGroupByRow = groupBy(
+    claims.filter(c => !c.txHash).sort(c => c.id),
+    c => {
+      return `${c.distribution.vault.vault_address} ${c.distribution.epoch.circle?.id}`;
+    }
   );
 
   const processClaim = async (claimId: number) => {
@@ -145,7 +154,7 @@ export default function ClaimsPage() {
               key={id}
               onClickClaim={() => processClaim(id)}
               claiming={claiming[id]}
-              claimsGroupByVault={claimsGroupByVault}
+              claimsGroupByRow={unclaimedClaimsGroupByRow}
             />
           )}
         </ClaimsTable>
@@ -181,7 +190,9 @@ export default function ClaimsPage() {
               </td>
               <td>
                 {formatEpochDates(
-                  claimsGroupByVault[distribution.vault.vault_address]
+                  claimedClaimsGroupByRow[
+                    `${distribution.vault.vault_address} ${distribution.epoch.circle?.id} ${txHash}`
+                  ]
                 )}
               </td>
               <td>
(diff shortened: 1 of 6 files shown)


levity:
The tricky thing here is that I think we should show the amount the user actually received, which is not necessarily either the new amount or the total amount. Consider this sequence of events for a user in a circle:

Distributed for Epoch 1: 500 DAI
Claimed: 500 DAI
Distributed for Epoch 2: 600 DAI (total 1100 DAI)
Distributed for Epoch 3: 700 DAI (total 1800 DAI)
Claimed: **1300 DAI**

The logic we need to follow for the Claims page is:
- Group the user's claims by (circle, token)
- Sort within each of those groups in descending order by ID or timestamp
- Next, group claims by tx hash: group claims that do not have a tx hash with the closest later one that does
  - This would group Epoch 2 with Epoch 3 in the above example
- Now the claimed amount for the tx is the sum of new_amount for all claims in its group

crabsinger:
Ok, that was a very helpful added context @levity 

So let's make this a bit more complex adding fixed payments as well to be sure we have clarity on the logic we expect.

### Claims Table

Assuming we have a list of available claims and tx claiming like:

| ID | Circle ID | Token | Epoch    | Distribution Type | Amount | Tx Hash          |
|----|-----------|-------|----------|-------------------|--------|------------------|
| 1  | 1         | DAI   | Epoch 1  | Circle            | 500    | Claimed 500 DAI  |
| 2  | 1         | DAI   | Epoch 2  | Circle            | 600    |                  |
| 3  | 1         | DAI   | Epoch 3  | Fixed Payment     | 700    | Claimed 1300 DAI |
| 4  | 1         | DAI   | Epoch 4  | Circle            | 100    |                  |
| 5  | 1         | DAI   | Epoch 4  | Fixed Payment     | 175    | Claimed 275 DAI  |
| 6  | 1         | USDC  | Epoch 2  | Circle            | 250    | Claimed 250 USDC |
| 7  | 2         | USDC  | Epoch 10 | Circle            | 80     | Claimed 80 USDC  |
| 8  | 3         | ETH   | Epoch 12 | Circle            | 9      | Claimed 9 Eth    |

In this case, we'd want to:
* group claims with ID 2 and 3, into a group with one tx
* group claims with ID 4 and 5, into a group, with one tx
* keep everything else in separate groups

Does that sound right to you?

So Fixed Payments vs Circle payments don't change anything or matter for grouping.

### The logic we need to follow for the Claims page is:

- Group the user's claims by (circle, token)
- Sort the groups by X?
- Sort within each group in descending order by ID or timestamp
- Group again by tx hash
For each claim that does not have a tx hash, group it with the closest later one that does

I'm not seeing any design mockups for grouped txs in https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=2603%3A3922 so I wonder if that's something we want to do now, or maybe we could just do something simpler with these all being rows, and somehow indicating a previous claim was wrapped into a later one, but still using a row view without too much grouping logic... cc @spydermonkee

levity:
> * group claims with ID 2 and 3, into a group with one tx
> * group claims with ID 4 and 5, into a group, with one tx
> * keep everything else in separate groups
> 
> Does that sound right to you?

Yes, exactly.

> I'm not seeing any design mockups for grouped txs

The way we've extended the design for groups is pretty simple: the value in the Date column covers the entire date range of the group. So for the group with IDs 2 and 3, if the start of Epoch 2 was June 1, and the end of Epoch 3 was June 30, it would be:

> Epochs 2-3 ... June 1-30, 2022

The other thing I'm noticing in [the mockup](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=4851%3A15456) is that there are two buttons per row when an epoch has two tokens. The page is definitely not implemented this way now--it's just one row per (token, circle) pair. I would support continuing like that, since it's simpler to implement. (e.g. we don't have to handle edge cases where users are claiming the two tokens at different times)

crabsinger:
> The way we've extended the design for groups is pretty simple: the value in the Date column covers the entire date range of the group. So for the group with IDs 2 and 3, if the start of Epoch 2 was June 1, and the end of Epoch 3 was June 30, it would be:
> 
> > Epochs 2-3 ... June 1-30, 2022

So we'd have one row per claim tx - even if it is pulling from multiple distributions or "available claims". I'm curious - haven't fully thought out - if this will provide adequate information to the users about the source of the funds and the total amounts. Like from the example above. If the row is:

```
DAI | Epochs 2-3 | Claimed 1300 DAI
```

Will a User be able to understand how much is from a Fixed Payment vs a Gift Circle vs something else? And is this ok or not... @levity  does that make sense?

levity:
> I'm curious - haven't fully thought out - if this will provide adequate information to the users about the source of the funds and the total amounts. Like from the example above. If the row is:
> 
> ```
> DAI | Epochs 2-3 | Claimed 1300 DAI
> ```
> 
> Will a User be able to understand how much is from a Fixed Payment vs a Gift Circle vs something else? And is this ok or not... @levity does that make sense?

I think this is ok for now. If we get feedback from people that they need more detail, we can improve UI and docs.

crabsinger:
@levity ok, got the tables displaying correctly, but uncovered a few other bugs that impact this:

- claiming a claim sets a txHash for ALL USERS claims...
- having two vaults for gift circle and fixed payments has bugs, so I need to figure that out so I can test the grouping logic by token.

I could ship this first part without those others if we want... needs a bit of cleanup but mostly there.

crabsinger:
Hasura Permissions on Claims Notes:

- Insert: Should only be able to create a claim of a given distribution ID if you are the owner (creator) of that distribution.

- Update: you can only update your own claims.

- When you're claiming for a group, update all claims in that group with the same txHash.

crabsinger:
Ok, claims are properly updated now - all claims will be updated by to the one tx that claims thegroup.

src/pages/ClaimsPage/ClaimsPage.tsx
```diff
@@ -23,10 +23,12 @@ const currentClaims = (claims: QueryClaim[]) =>
       (finalClaims, curr) =>
         finalClaims.filter(
           c =>
-            c.distribution.vault.vault_address ==
+            c.distribution.vault.vault_address ===
               curr.distribution.vault.vault_address &&
             c.distribution.epoch.circle?.id ===
-              curr.distribution.epoch.circle?.id
+              curr.distribution.epoch.circle?.id &&
+            c.distribution.vault.simple_token_address ===
```
levity, line 30: this additional simple_token_address check isn't necessary; you can't have a case where vault_address is the same and either of the token addresses differ

.env.example
```diff
@@ -8,6 +8,7 @@ IMAGES_AWS_BUCKET=coordinape
 IMAGE_DIR=assets/static/images/
 # separate allowed domains with commas
 SIWE_ALLOWED_DOMAINS=localhost:3000,
+DEV_SEED_PHRASE="test test test test test test test test test test test junk"
```
crabsinger, line 11: left here, but hopefully https://github.com/coordinape/coordinape/pull/1225 is merged ahead of this and can be removed

src/pages/ClaimsPage/ClaimsPage.tsx
```diff
@@ -80,9 +86,18 @@ export default function ClaimsPage() {
       </SingleColumnLayout>
     );
 
-  const claimsGroupByVault = groupBy(
-    claims.sort(c => c.id),
-    c => c.distribution.epoch.circle?.id && c.distribution.vault.vault_address
+  const claimedClaimsGroupByRow = groupBy(
```
crabsinger, line 89: could clean this up into one function handling both cases

crabsinger:
@levity @zashton this is ready for review, it needs a thororough QA again from me tomorrow but would otherwise ready for review.

I would especially like review of the mutation changes and permission logic @levity - I saw a weird bug locally where it seemed txHash was overwritten... but haven't been able to recreate the issue happening.

src/pages/ClaimsPage/queries.ts
```diff
@@ -80,12 +82,20 @@ export const getClaims = async (
     );
 
     const unwrappedAmount = getUnwrappedAmount(claim.amount, pricePerShare);
+    const unwrappedNewAmount = getUnwrappedAmount(
```
zashton, line 85: the pricePerShare will increase overtime when the vault profits. So the realtime unwrapped amount will also be constantly increasing, will user be confused if this number is shown ?

levity:
It’s a great question. I think we will have to educate users to expect this
behavior. At least the number doesn’t go down :) for now I don’t think we
need to change this.

src/pages/ClaimsPage/ClaimsPage.tsx
```diff
@@ -248,9 +268,7 @@ const ClaimRow = ({
       </td>
       <td>
         <Text>
-          {formatEpochDates(
-            claimsGroupByVault[distribution.vault.vault_address]
-          )}
+          {formatEpochDates(claimsGroupByRow[claimsRowKey(distribution)])}
```
levity, line 271: this would be a bit easier to read, maybe a tiny bit faster too, if you pushed the generation of `claimsGroupByRow[claimsRowKey(distribution)]` into a const above rather than running it twice.

actually, wait, you can just calculate it in the parent and pass it into ClaimRow as a prop, instead of passing in the entire `claimsGroupByRow`

levity (changes requested):

src/pages/ClaimsPage/ClaimsPage.tsx
levity: i think you could also just have one data structure, now that you're also grouping by txHash?

```
c1 = claims from hasura
c2 = claimRows(c1)
c3 = groupBy(c1.sort(c => -c.id), claimsRowKey) 
```

then when you loop through either `c2.filter(c => c.txHash)` or `c2.filter(c => !c.txHash)`, its corresponding group is already in `c3`

it's now apparent that c2 is actually completely redundant; instead of using that, you could do e.g. `Object.values(c3).filter(group => !group[0].txHash)` to get the unclaimed groups, then sort by the min id of each group to get the rows in order

this is probably a bit bikesheddy; no need to do it this way if what you have now works; but i hope it was at least informative

src/pages/ClaimsPage/ClaimsPage.tsx
```diff
@@ -80,9 +86,18 @@ export default function ClaimsPage() {
...
-    c => c.distribution.epoch.circle?.id && c.distribution.vault.vault_address
+  const claimedClaimsGroupByRow = groupBy(
+    claims.filter(c => c.txHash).sort(c => -c.id),
+    c => {
+      return claimsRowKey(c.distribution, c.txHash);
+    }
+  );
+
+  const unclaimedClaimsGroupByRow = groupBy(
+    claims.filter(c => !c.txHash).sort(c => c.id),
```
levity, line 97:
```suggestion
    claims.filter(c => !c.txHash).sort(c => -c.id),
```

crabsinger:
nitty bug:

We say "3 epochs" when we really mean 3 claims...

We have these claims for this user

```
{
  "data": {
    "claims": [
      {
        "id": 104,
        "txHash": null,
        "new_amount": 13.5,
        "distribution": {
          "vault_id": 10,
          "epoch_id": 74,
          "epoch": {
            "circle_id": 77
          }
        }
      },
      {
        "id": 110,
        "txHash": null,
        "new_amount": 9,
        "distribution": {
          "vault_id": 10,
          "epoch_id": 75,
          "epoch": {
            "circle_id": 77
          }
        }
      },
      {
        "id": 116,
        "txHash": null,
        "new_amount": 0.42,
        "distribution": {
          "vault_id": 10,
          "epoch_id": 75,
          "epoch": {
            "circle_id": 77
          }
        }
        ```

but are showing it as such:

image

Since there's:
1 combined gift circle + fixed payments claim for epoch 1
1 gift circle claim for epoch 2
1 gift circle claim for epoch 2

I think this is a case we can ignore for now, but want to note it here.

File: src/pages/ClaimsPage/ClaimsPage.tsx (hunk 6 above)
Comment (crabsinger) [reply src/pages/ClaimsPage/ClaimsPage.tsx]: i changed a bunch of the sort logic... 

right now I'm sorting so newest is at the top by default

File: src/pages/ClaimsPage/ClaimsPage.tsx [hunk 7]
```diff
@@ -126,22 +108,38 @@ export default function ClaimsPage() {
             { title: 'Epochs', css: styles.th },
             { title: 'Rewards', css: styles.thLast },
           ]}
-          data={currentClaims(claims.filter(c => !c.txHash))}
+          data={unclaimedClaimsRows}
           startingSortIndex={2}
-          startingSortDesc
+          startingSortDesc={false}
```
crabsinger, line 113: weirdly: this seems to sort by newest at the top

crabsinger:
@levity ok, take another look - paired with @META-DREAMER and refactored this and cleaned it up a BUNCH!

I can do some more testing but I think its looking good to me at this point

crabsinger:
@levity I have NOT yet done the goodHarvest and tested the wrapped amount stuff yet, fyi

src/pages/ClaimsPage/ClaimsPage.tsx
levity: In order for sorting to work in this table component, you have to define `sortByColumn`. It's a function that takes a column index and returns a function that produces a value to sort by. Right now it's just a placeholder `() => c => c` which means "no matter what column this is, just sort by the entire data object for each row". You can see a working example [here](https://github.com/coordinape/coordinape/blob/main/src/pages/DistributionsPage/AllocationsTable.tsx#L113)

levity:
As I mentioned in Discord, I'm going to merge this soon so I can fix my branch's conflicts. You can add the additional stuff to a new PR

src/pages/ClaimsPage/utils.ts
```diff
@@ -62,3 +64,14 @@ export const claimRows = (claims: QueryClaim[]) =>
         : [...finalClaims, curr],
     [] as QueryClaim[]
   );
+
+export const createClaimsRows = (claims: QueryClaim[]) => {
```
crabsinger, line 68: @levity nice! i like this approach to getting all the data grouped and organized as needed!

PR #1183, opened 2022-07-27
Add danger zone to circle admin page

sshmm:
## Motivation and Context
implements #966

## Description

1- Add deleted_at column to circles table
2- Add deleteCircle action
3- add danger zone to circle admin page
4- add remove circle modal with remove and cancel buttons
5- on remove circle the circle will be removed and navigation to overview page happens
6- on error show snackbar error

## Test and Deployment Plan

precondition : circle admin
1- Go to circle Admin page and click remove circle button in danger zone
2a - click remove circle button in remove circle modal
Expected:
Navigation to overview page and the circle should not appear from it
2b- click cancel button
Expected:
the modal should be closed
2c- hardcode a wrong circle number and click remove
Expected:
Snackbar should appear with error

## Screenshots (if appropriate):
image

Issue #966, opened 2022-05-26
1.5.6 - Admin Updates - Danger Zone


reeserj:
Dependent on https://github.com/coordinape/coordinape/issues/962

Any data fetching to be rebuilt using Zeus

[See Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=3560%3A11926)

**Subsections**
- Permanently Remove Circle

Build Modal [Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=5859%3A28498)
- **Make clicking remove inactivate the circle.**

sshmm:
@spydermonkee  should clicking delete the circle send a delete action to Hasura? I can't find active column in circles table

levity:
We need to add a deleted_at column

And then to change all the users of circles to include ‘deleted_at is null’
in their queries

It would be better to create a Postgres view of circles that always
includes that filter, and then have all other tables use that view…
migrating the schema to this might be tricky though.

sshmm:
closed by #1262

PR #1196, opened 2022-07-28
Add circles deleted_at column and deleteCircle action

sshmm:
## Motivation and Context
subtask of #966 and #1183

## Description

1- Add deleted_at column for circles table
2- Add deleteCircle action in hasura and handlers


hasura/migrations/default/1658417643392_alter_table_public_circles_add_column_deleted_at/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
+-- alter table "public"."circles" add column "deleted_at" time without time zone
+--  null;
```
crabsinger, line 4: I think this can just be 
```
 alter table "public"."circles" drop column "deleted_at"
 ```

File: hasura/migrations/default/1658850906511_alter_table_public_circles_add_column_deleted_at/down.sql [hunk 2]
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
+-- alter table "public"."circles" add column "deleted_at" timestamp
+--  null;
```
crabsinger, line 4: same drop column

hasura/metadata/databases/default/tables/public_nominees.yaml
```diff
@@ -90,71 +90,3 @@ select_permissions:
               - deleted_at:
                   _is_null: true
   role: user
-event_triggers:
```
crabsinger, line 93: I think this and the cron was disabled by the seed hook: https://github.com/coordinape/coordinape/blob/a9b424a973d73f9ddc8e6ec60ffe716902dc34f7/scripts/enable_triggers.sh#L12
 but you don't actually want to commit these changes

crabsinger (changes requested):
A few things before ready to merge:
- clean up down migration
- remove changes to event_triggers

hasura/metadata/databases/default/tables/public_users.yaml
```diff
@@ -2,181 +2,181 @@ table:
   name: users
   schema: public
 object_relationships:
-- name: circle
```
levity, line 5: please make sure you're using the same hasura version as in package.json, 2.2.0. otherwise it creates these indentation changes. (we'll want to upgrade this version of hasura and check in just indentation changes soon, in a different PR.)

hasura/metadata/databases/default/tables/public_vaults.yaml
```diff
@@ -48,4 +48,7 @@ select_permissions:
             - profile:
                 id:
                   _eq: X-Hasura-User-Id
+            - circle:
```
levity, line 51: you should add this checks further up in the relationships filter, where `circles` is already.

instead of:
```
circles:
  users:
```

you'll end up with something like:
```
circles:
  _and:
    deleted_at:
      _is_null: true
    users:
```

hasura/migrations/default/1658850906511_alter_table_public_circles_add_column_deleted_at/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."circles" add column "deleted_at" timestamp
```
levity, line 1: please remove the other two migrations

hasura/migrations/default/1658850906511_alter_table_public_circles_add_column_deleted_at/down.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."circles" add column "deleted_at" timestamp;
```
levity, line 1:
```suggestion
alter table "public"."circles" drop column "deleted_at" cascade;
```

api/hasura/actions/_handlers/deleteCircle.ts
```diff
@@ -0,0 +1,64 @@
...
+    input: { payload },
+  } = composeHasuraActionRequestBody(deleteCircleInput).parse(req.body);
+
+  const { circle_id } = payload;
+
+  const {
+    circles: [existingCircle],
+  } = await adminClient.query(
+    {
+      circles: [
```
levity, line 22: it's a bit simpler and cleaner, i think, to use `circles_by_pk` here and then throw the 422 error if `deleted_at` isn't blank:

```
circles_by_pk: [
  { id: circle_id },
  { deleted_at: true }
]
```

api/hasura/actions/_handlers/deleteCircle.ts
```diff
@@ -0,0 +1,64 @@
...
+          where: {
+            id: { _eq: circle_id },
+            // ignore soft_deleted circles
+            deleted_at: { _is_null: true },
+          },
+        },
+        { id: true },
+      ],
+    },
+    {
```
levity, line 34: please avoid splitting things like this onto multiple lines when it's not necessary

hasura/metadata/actions.yaml
```diff
@@ -416,5 +426,6 @@ custom_types:
       type: object
       field_mapping:
         id: id
+  - name: DeleteCirlceResponse
```
levity, line 429:
```suggestion
  - name: DeleteCircleResponse
```

levity (changes requested):

hasura/metadata/databases/default/tables/public_users.yaml
levity: please re-export the metadata once you are using the right version; the diffs should become a lot smaller

api/hasura/actions/_handlers/deleteCircle.ts
sshmm: I did the other changes but I don't understand this. you mean putting the braces and brackets in the same line? that changes automatically by lint after saving

levity: The linter won't put it on one line if you have it on multiple lines, only the reverse. In this case, this should all fit on one line:
```
{ operationName: 'deleteCircle_getExistingCircle' }
```

hasura/metadata/cron_triggers.yaml
```diff
@@ -1,35 +1 @@
-- name: check-nominee
```
crabsinger, line 1: @sshmm revert changes to this file pelase.

sshmm: done. I have to do it on multiple commits because VS code formatter add extra spaces when I save the file manually

api/hasura/actions/_handlers/deleteCircle.ts
```diff
@@ -0,0 +1,43 @@
...
+  } = composeHasuraActionRequestBody(deleteCircleInput).parse(req.body);
+
+  const { circle_id } = payload;
+
+  const { circles_by_pk: circle } = await adminClient.query(
+    { circles_by_pk: [{ id: circle_id }, { deleted_at: true }] },
+    { operationName: 'deleteCircle_getCircleStatus' }
+  );
+
+  if (circle?.deleted_at) {
```
levity, line 23:
```suggestion
  if (!circle || circle?.deleted_at) {
```

levity:
Two runs of Cypress [failed](https://github.com/coordinape/coordinape/actions/runs/2777231683) with the same error; @sshmm can you look into this?

sshmm:
@levity I traced the error and I found that the error is in the test itself.
The test adds vault for **Ended Epoch with Gifts for Claims** as it searches for an org that includes **Ended Epoch with for Giffts** and try to distribute for sports circle which in **Ended Epoch with Gifts**
I will try to fix the test tomorrow and I may need your support as it is my first time dealing with cypress

levity:
> @levity I traced the error and I found that the error is in the test itself. The test adds vault for **Ended Epoch with Gifts for Claims** as it searches for an org that includes **Ended Epoch with for Giffts** and try to distribute for sports circle which in **Ended Epoch with Gifts** I will try to fix the test tomorrow and I may need your support as it is my first time dealing with cypress

Nice find! I added a fix, 7a2ec487

PR #1198, opened 2022-07-29
Stitches friendly icon generation from figma export svgs with new `yarn figma-icons` command

CryptoGraffe:
## Motivation and Context

We have a bunch of icon components in different formats/techniques that I've been itching to automate. 

## Description

@smombartz exported a bunch of SVGs from the Design library and added them in a prior PR. This follow-up work uses `svgr` cli in a script to take those SVGs and wrap them in our custom SvgIcon component, creating nice files for each. 

I'm not in love with the directory structure, and none of these are used yet, but I wanted to get this in as I'm away for the next few days and thought that @wingfeatherwave might want to reorganize a bit and figure out how to phase out the old ones. 

If new icons get added, you can run `yarn figma-icons` and then git add them and push them.


levity:
Nice work!

I can't help but wonder if there's a way to do this without having to actually check in the generated files, using the ability to import svgs as text that [we already have](https://github.com/coordinape/coordinape/blob/main/src/pages/AllocationPage/AllocationTeam.tsx#L16). 

Though I see that going from `<svg>` to `<SvgIcon>` would take some extra work, maybe we could create a wrapper function that does that operation in a general way, and then the update script would just do something like:

```js
const generatedFile = "import { generateIcon } from './generateIcon';\n" + 
filesInIconDir.map(filename => 
  `export const SomeIcon = generateIcon('${filename}');`
).join('\n');
```

CryptoGraffe:
So there is a way to do this automatically in webpack with svgr but I was
really struggling with it, and it seemed to make recompile / startup super
slow.  The other thing is you don’t get nice autocomplete when referring to
an icon to import , you have to declare the import manually first and you
redefine the name each time you import it.

The technique you are proposing might work as well but I haven’t figured
out how to do the svg->SvgIcon part.

CryptoGraffe:
Fwiw the biggest struggle was replacing the svg element using the template and getting props passed through. Hence the sed in the script it was just so much easier.

CryptoGraffe:
Also fwiw, the built in stuff we have now is working via SVGR as well

levity (approved):

PR #1199, opened 2022-07-29
circle share token database schema

### LINKED ISSUE #1081: add table for secret circle tokens for joining and viewing information after invite
**Problem**
For our new onboarding work, we need secret links for joining a circle and viewing information about it after you are invited. 

**Solution**
A circle token table that supports both joinable tokens and info-only tokens

**Specification**
What precisely needs to get built? (include any requirements or design files)

CryptoGraffe:
## Motivation and Context

Fixes #1081 sets up ground work for new circle magic links / share links. 

## Description

New "circle share token" concept with two types: "magic" and "welcome" . 

One Token of each type can exist per circle at a time, with the ability to revoke+replace. Creation/revoke are only allowed by circle admins. 

If a prospective new member  gets access to a token (via a link, for example), they can call an API (automatically via the web app) that converts the token into information about the circle - org name, logo, etc.  This API is mostly useful only for magic links now, but would be useful for welcome links in the future if we allow some non-authed api calls of some sort. 

If the token is a "magic" token, they can also use this token to call the new `createUserWithToken` mutation, which will allow them to add themselves to the circle. 

## Test and Deployment Plan

Try making and revoking both types of tokens, use them to call the new actions. 

I tested this pretty thoroughly with a UI , which I'm breaking into a separate PR to make reviewing easier (maybe its easier??) 

## Notes

I already have a bunch of UI built around this but its not done and I figured I would get this in since I think its ready.


api/hasura/actions/_handlers/createUserMutation.ts
```diff
@@ -0,0 +1,99 @@
...
+            circle_id: { _eq: circleId },
+          },
+        },
+        {
+          id: true,
+          deleted_at: true,
+        },
+      ],
+    },
+    {
```
levity, line 22: can we avoid splitting objects into multiple lines where they would fit on one?

api/circle/landing/[token].ts
```diff
@@ -0,0 +1,62 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { adminClient } from '../../../api-lib/gql/adminClient';
+import { errorResponse } from '../../../api-lib/HttpError';
+import { Awaited } from '../../../api-lib/ts4.5shim';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  // TODO: does this have some kind of CORS protection?
```
Comment (crabsinger) [inline api/circle/landing/[token].ts:8]: say more about this - I approached a CORS attack with SIWE where I just checked the origin URL.

api/circle/landing/[token].ts
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    let token: string | undefined;
```
Comment (zashton) [inline api/circle/landing/[token].ts:10]: probably can simplify it this way
```ts
const token = (() => {
      if (typeof req.query.token == 'string') {
        return req.query.token;
      } else if(req.query.token instanceof Array && req.query.token[0]) {
         return req.query.token[0];
      } else {
        throw new Error('empty token provided');
      }
    })();
```

api/circle/landing/[token].ts
```diff
@@ -0,0 +1,62 @@
...
+          logo: true,
+          name: true,
+          organization: {
+            name: true,
+            logo: true,
+          },
+        },
+      },
+    ],
+  });
```
Comment (zashton) [inline api/circle/landing/[token].ts:53]: missing an operation name

api/hasura/actions/_handlers/createUserWithToken.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  console.log('creating user');
+
+  // create the user
+  const mutationResult = await createUserMutation(address, circleId, {
+    name: input.name,
+    circle_id: circleId,
+    // TODO: are these defaulted or do we need to specify values here? ??
+    // give_token_remaining: input.starting_tokens,
+    // starting_tokens:input.starting_tokens,
```
zashton, line 63: both are defaulted to `100` in db

hasura/metadata/databases/default/tables/public_circle_share_tokens.yaml
```diff
@@ -0,0 +1,53 @@
...
+  schema: public
+object_relationships:
+- name: circle
+  using:
+    foreign_key_constraint_on: circle_id
+insert_permissions:
+- permission:
+    backend_only: false
+    check:
+      circle:
```
zashton, line 12: `deleted_at:  _is_null: true` for circle

hasura/metadata/databases/default/tables/public_circle_share_tokens.yaml
```diff
@@ -0,0 +1,53 @@
...
+select_permissions:
+- permission:
+    columns:
+    - circle_id
+    - type
+    - created_at
+    - updated_at
+    - uuid
+    filter:
+      circle:
```
zashton, line 33: same check 
`deleted_at: _is_null: true for circle`
if its relevant

hasura/metadata/databases/default/tables/public_circle_share_tokens.yaml
```diff
@@ -0,0 +1,53 @@
...
+          - profile:
+              id:
+                _eq: X-Hasura-User-Id
+          - role:
+              _eq: 1
+  role: user
+delete_permissions:
+- permission:
+    filter:
+      circle:
```
zashton, line 45: same check
`deleted_at: _is_null: true for circle`
if its relevant

api/circle/landing/[token].ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+async function circleFromToken(token: string) {
+  const { circle_share_tokens } = await adminClient.query({
+    circle_share_tokens: [
+      {
+        where: {
+          uuid: {
+            _eq: token,
```
Comment (zashton) [inline api/circle/landing/[token].ts:36]: now theres a soft delete circle feature so you would need to check
`deleted_at: _is_null: true for circle`

api/hasura/actions/_handlers/createUserWithToken.ts
```diff
@@ -0,0 +1,71 @@
...
+  const address = await getAddress(hasuraProfileId);
+
+  console.log('getting tokens');
+  // get the circleId from the token and make sure its magic
+  const { circle_share_tokens } = await adminClient.query({
+    circle_share_tokens: [
+      {
+        where: {
+          uuid: {
+            _eq: input.token,
```
zashton, line 37: might want to validate if the circle is already soft deleted

src/lib/zod/index.ts
```diff
@@ -13,6 +13,10 @@ const PERSONAL_SIGN_REGEX = /0x[0-9a-f]{130}/;
 
 export const sha256HashString = z.string().length(64);
 
+export const circleLandingInput = z.object({
```
zashton, line 16: this doesn't seem to be used anywhere

zashton (changes requested):
looks good though just some extra validations needed because of the new delete circle feature

api/hasura/actions/_handlers/createUserMutation.ts
CryptoGraffe: Sorry i just follow whatever the linter does, maybe we can change it?

CryptoGraffe: @levity actually i can't really figure out what you mean here, maybe the source changed?

api/circle/landing/[token].ts
Comment (CryptoGraffe) [reply api/circle/landing/[token].ts:8]: I think this is a non-issue, the API routes are CORS protected by default i believe

api/circle/landing/[token].ts
Comment (CryptoGraffe) [reply api/circle/landing/[token].ts:10]: to me this is less simple haha, idk!

Comment (CryptoGraffe) [reply api/circle/landing/[token].ts:10]: i simplified it slightly, i think its ok

api/circle/landing/[token].ts
Comment (CryptoGraffe) [reply api/circle/landing/[token].ts]: addressed

api/circle/landing/[token].ts
Comment (CryptoGraffe) [reply api/circle/landing/[token].ts]: addressed

api/hasura/actions/_handlers/createUserWithToken.ts
CryptoGraffe: addressed

hasura/metadata/databases/default/tables/public_circle_share_tokens.yaml
CryptoGraffe: addressed

hasura/metadata/databases/default/tables/public_circle_share_tokens.yaml
CryptoGraffe: got it

hasura/metadata/databases/default/tables/public_circle_share_tokens.yaml
CryptoGraffe: done

src/lib/zod/index.ts
CryptoGraffe: deleted

zashton (approved):

zashton (approved):

PR #1201, opened 2022-07-29
Sascha app icons

smombartz:
added new icons and fixed typo


CryptoGraffe:
This is redundant with #1205

PR #1204, opened 2022-07-29
BugFix: stringify distribution number inputs

topocount:
This any type was passing a number into an ethers utility function that
only handles strings.

Added correct typings and stringified the value where appropriate

test plan:

createVault E2E Tests should pass reliably again


crabsinger (approved):
This wasn't only affecting tests - it also seemed to be erroring for me in local dev, but this branch works in local dev, so looks like fix is GOOD!!!

crabsinger:
rerunning the job. not sure if its a spurious failure or meaningful...

zashton (approved):
tested working, the failed test doesn't seem related to this change so all good.

PR #1207, opened 2022-07-30
Create Vault Deployment Recovery Feature

topocount:
## Motivation and Context

Sometimes a user might exit the app before a vault deployment transaction is 
mined. This is problematic because a user spent funds on that deploy, but it
will not be reflected in our app. This recovery cron task ensures vaults
deployed within our app that are successfully mined will be reflected in our
backend whether or not the user had their page open throughout the whole time.

Moreover, this PR lays the groundwork for recovering backend state for untracked
distributions created in our app and untracked claims, which are both much simpler
to recover than a vault.

## Description

This implementation still probably needs more testing, but would probably work
if deployed without further modification. The backend business logic for
recovering a vault is pretty well commented as far as describing the
disambiguation necessary to recover a deployed vault.

We cannot yet recover vaults deployed outside of our app because we need to know
which protocol/org the vault should be assigned to.

The cron task runs every 2 minutes and only begins querying txs that were
submitted at least 5 minutes ago. This means a successfully submitted tx
should show up in the backend no more than 7 minutes after mining if the
browser window isn't open.

## Test and Deployment Plan

This is probably easiest to test on goerli, unless you configure your local
devchain to delay mining blocks.

Basically, just deploy a vault and close your window after submitting the tx.
Observe the entry created in pending_vault_tx table.

The tx must be at least five minutes old before the cron recovery task picks it
up. You can adjust this in the handler if you want it to take less time.

Wait/watch for the cron task to run in hasura and check that the vault shows up
after the task runs

## Reviewers

@levity
@zashton
@crabsinger


levity:
reminder to self: modify permissions once aa3a6873 (#1196) is merged

levity (approved):

levity (approved):

Issue #887, opened 2022-05-10
Circle Level API Keys


META-DREAMER:
**Problem**
Users / devs can't access circle data via API without using their personal auth token and in a way that lets them control access to specific tables. 

**Solution**
A new api-user role with circle level API keys

**Specification**

Deliverables & deadlines of the project

- Create Hasura database table to store API keys:
    - circle_id: Circle that the API key has access to
    - created_by: The circle admin that created the API key
    - value: The actual API key string
    - allow_writes: If true, API key has read+write access. Otherwise its readonly
    - description: String describing the API key’s purpose.
- Set permissions for these API keys under a new “api-user” role to ensure it only reveal appropriate data (scoped to specific circles, not users, and **only exposing data that is sensitivity 2 or lower** from here: [](https://www.notion.so/ff0971d3bb4c4f4f8edad62ba5af8662))
- Allow circle admins to specify **Readonly** vs **Read + Write** for API keys they generate.
- Update auth handler to validate API keys and assign the api-user role.
- Create documentation to help users create and use these API keys

**Stakeholders**
Product team to give input on data visibility / privacy

META-DREAMER:
closed in #1212

PR #1212, opened 2022-08-02
Circle Write APIs

META-DREAMER:
## Description

Added permissions + validation logic in the action handlers to allow the api-user role to call them. For actions that require the acting users ID, added a param to let API users specify which user they are taking the action on behalf of (note that we have no way to enforce that API users actually authenticated or validated that user wanted to take that action, essentially giving API keys "admin rights" on updating allocations or vouches on behalf of other users)

Removed the update permissions on the raw Hasura tables for update_circle since we want to force people to use the updateCircle action.

## Test and Deployment Plan

Call the updateAllocations, updateCircle, and vouch actions with an API key that doesn't have the appropriate permissions and ensure the request is blocked.

Call updateAllocations or vouch with an API key with the appropriate permissions, but act as a user that belongs to a different circle and ensure the request is blocked.

Call the mutations with an API key with the appropriate permissions and valid input to ensure the request succeeds.

CleanShot 2022-08-01 at 19 11 38
CleanShot 2022-08-01 at 19 07 21

## Related Issue

#887


api-lib/circleAdmin.ts
```diff
@@ -39,6 +48,21 @@ const middleware =
         await handler(req, res);
         return;
       }
+
+      if (sessionVariables.hasuraRole === 'api-user') {
```
crabsinger, line 52: I'd think it'd be nice to have this be a constant rather than a string. `API_USER_ROLE` maybe?

META-DREAMER: its strongly typed, so if the string was anything invalid it would throw a type error, so don't think it would be necessary in this case

CryptoGraffe:
This is the first time using the API. Some note on the API page about how to use the key as a bearer token would be nice, I had to launch the explorer and kind of mentally reverse engineer that to figure it out.

CryptoGraffe:
When trying to run updateAllocations with an API key w/o the proper perms I got this error, which seems a bit more detailed and confusing than I'd expect:

```
{
  "errors": [
    {
      "extensions": {
        "details": [
          {
            "path": [],
            "code": "invalid_union",
            "unionErrors": [
              {
                "name": "ZodError",
                "issues": [
                  {
                    "received": "undefined",
                    "path": [
                      "session_variables",
                      "x-hasura-user-id"
                    ],
                    "code": "invalid_type",
                    "message": "Required",
                    "expected": "string"
                  },
                  {
                    "path": [
                      "session_variables",
                      "x-hasura-role"
                    ],
                    "code": "invalid_literal",
                    "message": "Invalid literal value, expected \"user\"",
                    "expected": "user"
                  },
                  {
                    "received": "undefined",
                    "path": [
                      "session_variables",
                      "x-hasura-address"
                    ],
                    "code": "invalid_type",
                    "message": "Required",
                    "expected": "string"
                  }
                ]
              },
              {
                "name": "ZodError",
                "issues": [
                  {
                    "received": "undefined",
                    "path": [
                      "input",
                      "payload",
                      "user_id"
                    ],
                    "code": "invalid_type",
                    "message": "Required",
                    "expected": "number"
                  },
                  {
                    "path": [
                      "session_variables"
                    ],
                    "code": "custom",
                    "message": "Provided API key does not have the required permissions: update_pending_token_gifts"
                  }
                ]
              }
            ],
            "message": "Invalid input"
          }
        ],
        "code": "422"
      },
      "message": "Invalid input: Invalid input (undefined)"
    }
  ]
}
```

CryptoGraffe:
I think that error is not about perms, its chained/added-on to the perms error. I get it even with a valid API key. Something with updateAllocations validation, i am sort of throwing garbage data at it but the error isn't helpful . Maybe this is for another new issue:
```
mutation MyMutation {
  updateAllocations(
    payload: {circle_id: 12, allocations: {note: "zip", recipient_id: 1, tokens: 1}}
  ) {
    user_id
  }
}

{
  "errors": [
    {
      "extensions": {
        "details": [
          {
            "path": [],
            "code": "invalid_union",
            "unionErrors": [
              {
                "name": "ZodError",
                "issues": [
                  {
                    "received": "undefined",
                    "path": [
                      "session_variables",
                      "x-hasura-user-id"
                    ],
                    "code": "invalid_type",
                    "message": "Required",
                    "expected": "string"
                  },
                  {
                    "path": [
                      "session_variables",
                      "x-hasura-role"
                    ],
                    "code": "invalid_literal",
                    "message": "Invalid literal value, expected \"user\"",
                    "expected": "user"
                  },
                  {
                    "received": "undefined",
                    "path": [
                      "session_variables",
                      "x-hasura-address"
                    ],
                    "code": "invalid_type",
                    "message": "Required",
                    "expected": "string"
                  }
                ]
              },
              {
                "name": "ZodError",
                "issues": [
                  {
                    "received": "undefined",
                    "path": [
                      "input",
                      "payload",
                      "user_id"
                    ],
                    "code": "invalid_type",
                    "message": "Required",
                    "expected": "number"
                  }
                ]
              }
            ],
            "message": "Invalid input"
          }
        ],
        "code": "422"
      },
      "message": "Invalid input: Invalid input (undefined)"
    }
  ]
}
```

CryptoGraffe:
It was because I didn't supply the user_id field in updateAllocations

CryptoGraffe:
> It was because I didn't supply the user_id field in updateAllocations

This should be the error returned: 
```    assert('user_id' in input, 'user_id not specified');
```

but zod is killing it first

api/hasura/actions/_handlers/vouch.ts
```diff
@@ -18,25 +19,56 @@ import {
...
-    } = composeHasuraActionRequestBodyWithSession(
-      vouchInput,
-      HasuraUserSessionVariables
-    ).parse(req.body);
+    } = await requestSchema.parse(req.body);
+
+    let voucherProfileId: number | null;
+    // Allow passing in voucher_user_id for api-users to vouch on behalf of a user
+    if (sessionVariables.hasuraRole === 'api-user') {
+      assert('voucher_user_id' in input, 'voucher_user_id not specified');
```
CryptoGraffe, line 54: I think this error handling will get lost because zod will have already thrown

api/hasura/actions/_handlers/vouch.ts
```diff
@@ -18,25 +19,56 @@ import {
...
+      voucherProfileId = input.voucher_user_id;
+
+      // Check if voucher exists in the same circle as the API key
+      // TODO: this uses assert for error handling
```
CryptoGraffe, line 58: others in the group seemed fine w/ using assert, do you have concerns about it? In fact you are using it above. I don't have a strong opinion but coming from other languages it seemed a bit strange. maybe its very normal in node/ts

api/hasura/actions/_handlers/vouch.ts
```diff
@@ -18,25 +19,56 @@ import {
...
-      vouchInput,
-      HasuraUserSessionVariables
-    ).parse(req.body);
+    } = await requestSchema.parse(req.body);
+
+    let voucherProfileId: number | null;
+    // Allow passing in voucher_user_id for api-users to vouch on behalf of a user
+    if (sessionVariables.hasuraRole === 'api-user') {
+      assert('voucher_user_id' in input, 'voucher_user_id not specified');
+      voucherProfileId = input.voucher_user_id;
```
CryptoGraffe, line 55: why is this a profile ID instead of a user ID? the field is called user_id in the input

api/hasura/actions/_handlers/vouch.ts
```diff
@@ -18,25 +19,56 @@ import {
...
+
+      // Check if voucher exists in the same circle as the API key
+      // TODO: this uses assert for error handling
+      const voucher = await getUserFromProfileId(
+        voucherProfileId,
+        sessionVariables.hasuraCircleId
+      );
+      if (!voucher) {
+        throw new Error(
+          'This API key does not belong to the same circle as the voucher'
```
CryptoGraffe, line 65: Seems like this error should be more specific - the circleID is provided, and that is checked against the API key before this, so the intent to use KEY+Circle is correct. What is incorrect here is that the userID (or profile ID?) is not a member of the current circle

CryptoGraffe (commented):
This looks good! nice work. Just a few minor things mostly around error handling and then some confusion around voucher user vs profile id.

api/hasura/actions/_handlers/vouch.ts
META-DREAMER: This is just copy pasted code from somewhere else, didn't write that comment myself

api/hasura/actions/_handlers/vouch.ts
META-DREAMER: Not sure tbh, it's copy pasted to handle the api-user role. Can take a deeper look if you want me to address in this PR

META-DREAMER:
> > It was because I didn't supply the user_id field in updateAllocations
> 
> This should be the error returned:
> 
> ```
> assert('user_id' in input, 'user_id not specified');
> ```
> 
> but zod is killing it first

The code is designed to not actually hit this ever, its just to make TS happy. Its intended for Zod to catch the malformed input first

api/hasura/actions/_handlers/vouch.ts
META-DREAMER: yea, its just to make TS happy, not intended to actually ever fire.

api/hasura/actions/_handlers/vouch.ts
META-DREAMER: renaming to `voucher_profile_id`

api/hasura/actions/_handlers/vouch.ts
CryptoGraffe: lol i think i wrote it tbh

api/hasura/actions/_handlers/vouch.ts
CryptoGraffe: I don't see this change pushed?

api/hasura/actions/_handlers/vouch.ts
CryptoGraffe: this isn't changed either, maybe a push is missing?

CryptoGraffe:
> The code is designed to not actually hit this ever, its just to make TS happy. Its intended for Zod to catch the malformed input first

What is your thought on the zod errors I was getting and their lack of clarity? Did you make a change? I'm thinking i'm missing some commit or something

META-DREAMER:
@CryptoGraffe sorry messed up my commit, and also misunderstood your comments to mean that it wasn't an issue since it was invalid input. Did some work to change the zod schema to use a discriminated union for a cleaner error message. Here's what its like now:

```
mutation MyMutation {
  updateAllocations(
    payload: {circle_id: 12, allocations: {note: "zip", recipient_id: 1, tokens: 1}}
  ) {
    user_id
  }
}

{
  "errors": [
    {
      "extensions": {
        "details": [
          {
            "received": "undefined",
            "path": [
              "input",
              "payload",
              "user_id"
            ],
            "code": "invalid_type",
            "message": "Required",
            "expected": "number"
          },
          {
            "path": [
              "session_variables"
            ],
            "code": "custom",
            "message": "Provided API key does not have the required permissions: update_pending_token_gifts"
          }
        ],
        "code": "422"
      },
      "message": "Invalid input: Required (user_id), Provided API key does not have the required permissions: update_pending_token_gifts (session_variables)"
    }
  ]
}
```

CryptoGraffe (dismissed):

crabsinger (approved):

PR #1215, opened 2022-08-02
Update Claim perms

zashton:
Insert: Should only be able to create a claim of a given distribution ID if you are the owner (creator) of that distribution.

Update: you can only update your own claims.

### DIFF
--- a/hasura/metadata/databases/default/tables/public_claims.yaml
+++ b/hasura/metadata/databases/default/tables/public_claims.yaml
@@ -19,15 +19,8 @@ insert_permissions:
     backend_only: false
     check:
       distribution:
-        epoch:
-          circle:
-            users:
-              _and:
-              - deleted_at:
-                  _is_null: true
-              - profile:
-                  id:
-                    _eq: X-Hasura-User-Id
+        created_by:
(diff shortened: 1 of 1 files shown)


crabsinger (approved):

levity:
Note that as of 056ab91acb8aa7f8fe2012eaed85aba38ddc1e3d, you have to check that circles.deleted_at is also null

zashton:
> Note that as of [056ab91](https://github.com/coordinape/coordinape/commit/056ab91acb8aa7f8fe2012eaed85aba38ddc1e3d), you have to check that circles.deleted_at is also null

i think user should be able to claim the funds even though circle is deleted, but not sure if the UI handles that edge case.

levity:
> > Note that as of [056ab91](https://github.com/coordinape/coordinape/commit/056ab91acb8aa7f8fe2012eaed85aba38ddc1e3d), you have to check that circles.deleted_at is also null
> 
> 
> 
> i think user should be able to claim the funds even though circle is deleted, but not sure if the UI handles that edge case.

Agreed--that's for the update. We want to check circles.deleted_at on insert

PR #1216, opened 2022-08-02
Create relationship from distributions.created_by -> profile.id

crabsinger:
Create fkey relationship from distributions.created_by -> profile.id

### DIFF
--- a/hasura/metadata/databases/default/tables/public_distributions.yaml
+++ b/hasura/metadata/databases/default/tables/public_distributions.yaml
@@ -2,6 +2,15 @@ table:
   name: distributions
   schema: public
 object_relationships:
+- name: distribution_creator
+  using:
+    manual_configuration:
+      column_mapping:
+        created_by: id
+      insertion_order: null
+      remote_table:
+        name: profiles
+        schema: public
 - name: epoch
   using:
     foreign_key_constraint_on: epoch_id
--- a/hasura/metadata/databases/default/tables/public_distributions.yaml
+++ b/hasura/metadata/databases/default/tables/public_distributions.yaml
@@ -2,18 +2,12 @@ table:
   name: distributions
   schema: public
 object_relationships:
-- name: distribution_creator
-  using:
-    manual_configuration:
-      column_mapping:
-        created_by: id
-      insertion_order: null
-      remote_table:
-        name: profiles
-        schema: public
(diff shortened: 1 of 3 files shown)


hasura/metadata/databases/default/tables/public_distributions.yaml
```diff
@@ -2,6 +2,15 @@ table:
   name: distributions
   schema: public
 object_relationships:
+- name: distribution_creator
```
zashton, line 5: i think if you create the foreign key first before adding the relationship, it should look something like that instead

```
object_relationships:
- name: distribution_creator
  using:
    foreign_key_constraint_on: created_by
```

zashton (changes requested):

hasura/metadata/databases/default/tables/public_distributions.yaml
crabsinger: thanks @zashton - added the migrations and changed this.

hasura/metadata/databases/default/tables/public_distributions.yaml
```diff
@@ -5,6 +5,9 @@ object_relationships:
 - name: epoch
   using:
     foreign_key_constraint_on: epoch_id
+- name: profile
```
crabsinger, line 8: I'm just naming this profile as it matches convention... but we could name it something else

hasura/migrations/default/1659467509391_set_fk_public_distributions_created_by/up.sql
```diff
@@ -0,0 +1,5 @@
+alter table "public"."distributions"
+  add constraint "distributions_created_by_fkey"
+  foreign key ("created_by")
+  references "public"."profiles"
+  ("id") on update restrict on delete restrict;
```
crabsinger, line 5: I wasn't sure if these should be `restrict` or something else... restrict seems to make sense since we're doing soft deletes and hard deletes would mess with our data.

zashton (approved):
looks good to me

hasura/migrations/default/1659467509391_set_fk_public_distributions_created_by/up.sql
zashton: yeah restrict is good enough

PR #1219, opened 2022-08-02
🪲 Bugfix for Undelete-able zero in input field

wingfeatherwave:
Number form fields that use `<DeprecatedApeTextField` faced an issue where the initial input value of `0` could not easily be deleted.  
* Fix leading-zero bug
* Set some number only fields to `type="number"`
* Fix vouching input field bug

### DIFF
--- a/cypress/e2e/updateUser.cy.ts
+++ b/cypress/e2e/updateUser.cy.ts
@@ -76,9 +76,9 @@ context('Coordinape', () => {
       });
 
     // enter the fixed payment amount
-    cy.getInputByLabel('Fixed Payment Amount').clear().type('1200').blur();
+    cy.getInputByLabel('Fixed Payment Amount').clear().type('12000').blur();
 
-    cy.contains('Save').click();
+    cy.contains('Save').click({force: true});
     cy.reload(true);
     cy.contains('Kasey', { timeout: 120000 }).should('be.visible');
     // Verify new value in contributors table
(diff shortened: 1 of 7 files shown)


wingfeatherwave:
woof, these cypress tests are annoying me.  It passed locally. is it timing out?

cypress/e2e/updateUser.cy.ts
```diff
@@ -76,9 +76,9 @@ context('Coordinape', () => {
       });
 
     // enter the fixed payment amount
-    cy.getInputByLabel('Fixed Payment Amount').clear().type('1200').blur();
+    cy.getInputByLabel('Fixed Payment Amount').clear().type('12000').blur();
```
zashton, line 79: The test is failing here on this 12000 value, i guess the 0 is still appended to the end of the number.
```suggestion
    cy.getInputByLabel('Fixed Payment Amount').clear().type('1200').blur();
```

@wingfeatherwave

zashton (changes requested):

cypress/e2e/updateUser.cy.ts
levity: i was actually just thinking of removing this test. i don't think it's high-value enough to incur the time cost of Cypress, since the first test in this file already verifies that the form is working.

levity (dismissed):
Approving the diff; haven't QA'ed; assuming you've tested changes already.

zashton (approved):

PR #1222, opened 2022-08-03
Distribution csv sync

zashton:
Fixes
https://www.notion.so/coordinape/VAULTS-QA-dd5a668750314cbc8f9a431fac1132a0?p=5658986b4684436a8efe22b7eb4588d7&pm=s

Syncing the CSV with the allocation to show preview values

### DIFF
--- a/api/hasura/actions/_handlers/allocationCsv.ts
+++ b/api/hasura/actions/_handlers/allocationCsv.ts
@@ -4,7 +4,7 @@ import type { VercelRequest, VercelResponse } from '@vercel/node';
 import { v4 as uuidv4 } from 'uuid';
 
 import { authCircleAdminMiddleware } from '../../../../api-lib/circleAdmin';
-import { formatShortDateTime } from '../../../../api-lib/dateTimeHelpers';
+import { DISTRIBUTION_TYPE } from '../../../../api-lib/constants';
 import { adminClient } from '../../../../api-lib/gql/adminClient';
 import { getEpoch } from '../../../../api-lib/gql/queries';
 import { errorResponseWithStatusCode } from '../../../../api-lib/HttpError';
@@ -20,7 +20,8 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     input: { payload },
   } = composeHasuraActionRequestBody(allocationCsvInput).parse(req.body);
 
-  const { circle_id, epoch_id, epoch } = payload;
+  const { circle_id, epoch_id, epoch, form_gift_amount, gift_token_symbol } =
+    payload;
   const epochObj = await getEpoch(circle_id, epoch_id, epoch);
   if (!epochObj) {
     return errorResponseWithStatusCode(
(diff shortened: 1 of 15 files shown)


api/hasura/actions/_handlers/allocationCsv.ts
```diff
@@ -4,7 +4,8 @@ import type { VercelRequest, VercelResponse } from '@vercel/node';
 import { v4 as uuidv4 } from 'uuid';
```
levity, line 4: @zashton can you please write a jest test for this, to verify that when you add certain inputs, they show up in the CSV? you can see an example of how to mock the request & response objects as well as the hasura query [here](https://github.com/coordinape/coordinape/blob/main/api/hasura/cron/_recoverVaults.test.ts) if needed. thanks!

levity: it may make sense to have a few different tests to cover the different cases (gift, fixed, combined)

api/hasura/actions/_handlers/allocationCsv.ts
```diff
@@ -29,88 +32,120 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
+    d =>
+      d.distribution_type === DISTRIBUTION_TYPE.FIXED ||
+      d.distribution_type === DISTRIBUTION_TYPE.COMBINED
+  );
+
+  giftTokenSymbol = circleDist ? circleDist.vault.symbol : giftTokenSymbol;
+  const { users } = circle;
+
+  const userValues: (string | number)[][] = [];
   users.map((u, idx) => {
```
topocount, line 125: this should probably be `.forEach` instead of `.map` since we're not using the returned result.

api/hasura/actions/_handlers/allocationCsv.ts
```diff
@@ -29,88 +32,120 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
     'sent',
-    'epoch_number',
-    'Date',
+    'givers',
+    'percentage_of_give',
+    'circle_rewards',
+    'circle_rewards_token',
   ];
   if (fixedPaymentsEnabled) {
     headers.push('fixed_payment_amount');
```
topocount, line 73:
```suggestion
    headers.push('fixed_payment_rewards');
```

api/hasura/actions/_handlers/allocationCsv.ts
```diff
@@ -29,88 +32,120 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
+  const fixedPaymentsEnabled =
+    isFeatureEnabled('fixed_payments') && !!circle.fixed_payment_token_type;
+
+  const userValues = generateCsvValues(
+    circle,
+    form_gift_amount,
+    gift_token_symbol,
+    totalTokensSent,
+    fixedPaymentsEnabled,
+    circle.fixed_payment_token_type,
```
topocount, line 57: This doesnt seem like the best way to provide this value (from the circle details) since it can be changed by the user at any time. Why not use the same token if its a combined distribution, or use the vault symbol from the fixed distribution if its a separate distribution?

topocount (changes requested):
I appreciate breaking this down into functions but it's still hard to follow everything that's going on. Maybe adding some comments here and there could help.

api/hasura/actions/_handlers/allocationCsv.ts
levity: i agree this is a real issue, and the root cause is that we don't take a snapshot of the circle's settings at the end of the epoch. 

> Why not use the same token if its a combined distribution, or use the vault symbol from the fixed distribution if its a separate distribution?

we can't assume there has been a distribution.

PR #1225, opened 2022-08-03
Allow option to read dev seed phrase from ENV

crabsinger:
Allow option to read dev seed phrase from ENV

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -8,6 +8,7 @@ IMAGES_AWS_BUCKET=coordinape
 IMAGE_DIR=assets/static/images/
 # separate allowed domains with commas
 SIWE_ALLOWED_DOMAINS=localhost:3000,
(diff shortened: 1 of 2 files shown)


.env.example
```diff
@@ -8,6 +8,7 @@ IMAGES_AWS_BUCKET=coordinape
 IMAGE_DIR=assets/static/images/
 # separate allowed domains with commas
 SIWE_ALLOWED_DOMAINS=localhost:3000,
+DEV_SEED_PHRASE="test test test test test test test test test test test junk"
```
crabsinger, line 11: Seems to not work... and the process.env.DEV_SEED_PHRASE just seems to interpret this string as "test" with only one word... unsure hwo to pass a string with spaces here?

levity: Spaces are supposed to work: https://github.com/motdotla/dotenv#what-rules-does-the-parsing-engine-follow

PR #1230, opened 2022-08-05
Claims Improvements

crabsinger:
Claims Page Improvements
- add circle logo to UI

- edgecase-fix: only update older claims:

If you load the FE, and get claims 1,2,3. And then while on that page a claim 4 is created, claiming 3 would cause claim 4 to be marked as claimed erroneously. This fixes that edge case.

### DIFF
--- a/src/pages/ClaimsPage/ClaimsPage.tsx
+++ b/src/pages/ClaimsPage/ClaimsPage.tsx
@@ -1,5 +1,5 @@
 import { LoadingModal, makeTable } from 'components';
-import { Box, Panel, Flex, Text, Button } from 'ui';
+import { Avatar, Box, Panel, Flex, Text, Button } from 'ui';
 import { SingleColumnLayout } from 'ui/layouts';
 import { makeExplorerUrl } from 'utils/provider';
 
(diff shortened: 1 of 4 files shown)


src/pages/ClaimsPage/ClaimsPage.tsx
```diff
@@ -23,10 +23,17 @@ const ClaimsRow: React.FC<ClaimsRowData> = ({ claim, group, children }) => {
   return (
     <tr>
       <td>
-        <Text>{claim.distribution.epoch.circle?.organization?.name}</Text>
+        <Flex>
+          {claim.distribution.epoch.circle?.logo ? (
+            <Avatar small path={claim.distribution.epoch.circle?.logo} />
```
crabsinger, line 28: this has a margin around it - and it's aligning wrong - so I need to remove all the margin except the right margin... not sure cleanest way to do this with stitches
Screen Shot 2022-08-05 at 4 22 25 PM

levity: Avatar is written to pass all props not otherwise specified to the AvatarRoot component, which takes the standard `css` prop, so adding something like `css={{ my: 0, ml: 0 }}` should do the trick

crabsinger: fixed this with help from @CryptoGraffe - but its kinda ugly.... and have to pass cssProps manually through. @levity

levity: oh that's interesting. you could inline your ternary if you want it to be a bit more compact:
```
{...(css ? { css } : {})}
```

teeolendo (approved):

PR #1232, opened 2022-08-05
Recover distributions

levity:
## Motivation and Context

https://www.notion.so/coordinape/Recover-distribution-transactions-f66554e0bbad40f3b3e85d2e69b6b1d6

## Description

extend the `recoverTransactions` cron job to handle Distribution transactions as well

### DIFF
--- a/api/hasura/cron/recoverTransactions.ts
+++ b/api/hasura/cron/recoverTransactions.ts
@@ -1,20 +1,19 @@
 import assert from 'assert';
 
 import type { TransactionReceipt } from '@ethersproject/abstract-provider';
+import { BigNumber } from '@ethersproject/bignumber';
 import { AddressZero } from '@ethersproject/constants';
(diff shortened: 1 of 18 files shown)


api/hasura/cron/recoverTransactions.ts
```diff
@@ -219,21 +225,113 @@ const handleClaim = async (
...
+      //     payload: {
+      //       tx_type: 'Distribution',
+      //       tx_hash,
+      //       distribution_id,
+      //       circle_id: data.epoch.circle_id,
+      //       vault_id: data.vault.id,
+      //     },
+      //   },
+      //   { id: true },
+      // ],
```
topocount, line 312: This throws an error because Hasura requires an authenticated profile to call this function, so the `created_by` field can be generated. The adminClient doesnt map to a profile so it cannot call this function.

topocount (approved):

PR #1243, opened 2022-08-07
delete api key table when seeding fresh

CryptoGraffe:
## Motivation and Context

this was just an oversight in the db_clean script when this new table was added


levity:
btw, i worked out a while back how to truncate all tables in postgres without having to list them out explicitly: https://github.com/coordinape/coordinape/blob/main/scripts/seed_hasura.sh#L24

didn't put this into action at the time because i wasn't sure it wouldn't be disruptive (adding a dependency on `psql` and certain enviroment variables for Postgres access instead of using Hasura), but just wanted to raise awareness that it's an option.

PR #1247, opened 2022-08-08
store distribution_json properly

levity:
## Motivation and Context

The `distributions.distribution_json` column is being stored as a giant plain-text string instead of JSON, even though the database table type is already `jsonb`.

This was a compromise when the feature was initially implemented due to lack of support from Zeus, but Zeus's support improved.

## Description

- store and read the `distribution_json` column as JSON
- add a migration to convert existing column values

----

This is based off of `recover-distributions` because that has lots of related changes; will rebase once #1232 is merged.


topocount (approved):
beautiful!

Issue #789, opened 2022-04-16
Add a nominee after expiry data


sshmm:
**Problem**
If nomination period ended, it is not possible to re-add a nominee.
![image](https://user-images.githubusercontent.com/34943689/163694223-e571d72c-a551-482e-981a-8ff046be1a70.png)

**Solution**
Update the expiry-date with a new value if the an expired nomination is re-nominated

levity:
> Update the expiry-date with a new value if an expired nomination is re-nominated

This solution makes sense to me. @reeserj do you think existing circle admins will be cool with this?

reeserj:
@levity yes. I don't think there'd be any pushback on this.

Should I move it to the ready for coding column?

levity:
Yes please!

levity:
good point! we never caught this in the original design.

PR #1249, opened 2022-08-08
feat: Update the expiry-date with a new value if the an expired nominee is re-nominated

karelianpie:
## Motivation and Context

If nomination period ended, it is not possible to re-add a nominee.

## Description

Update the expiry-date with a new value if the an expired nominee is re-nominated

## Related Issue

https://github.com/coordinape/coordinape/issues/789

### DIFF
--- a/api-lib/nominees.ts
+++ b/api-lib/nominees.ts
@@ -179,3 +179,32 @@ export const insertNominee = async (params: {
 
   return insert_nominees_one;
 };
+
+export const updateNominee = async (params: {
+  address: string;
+  circle_id: number;
+  nomination_days_limit: number;
+}) => {
+  const today = new Date();
+  const expiry = new Date();
+  expiry.setDate(today.getDate() + params.nomination_days_limit);
+  const input = { ...params, nomination_days_limit: undefined };
+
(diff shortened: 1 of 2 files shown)


crabsinger:
This looks like a fine solution to this:

I see the cron job running daily and seems to be running successfully

image

> So, the issue reported could potentially be happening because the trigger to mark the expired nominees ended to true is not running?

So maybe its that this isn't running FREQUENTLY ENOUGH - or there's a bug in that job...

@reeserj maybe we can test this in prod and see check if a nominee is getting set to expired true after this cron job runs as a follow up to this?

reeserj:
Waiting for a nom to time out will test tomorrow night

karelianpie:
> So maybe its that this isn't running FREQUENTLY ENOUGH - or there's a bug in that job...

I agree with this, the OP probably tried to re-add the nominee after the expiry date but before the cron had a chance to run. Although there's really no way to make the cron run frequently enough, unless we set the expiry date to EOD and make it run just after for example.

reeserj:
This is working!

zashton (approved):
looks good to me

Issue #1168, opened 2022-07-22
Minor controller update to DistributionForm


zashton:
Update DistributionForm to use `useController` instead of `FormControl`

sshmm:
@reeserj  @zashton  is anyone working on this? can I work on it?

reeserj:
You got it @sshmm

levity:
Beware though, this is a very active area of the code right now. You may
want to coordinate with @topocount in Discord so you don't conflict

PR #1251, opened 2022-08-09
Update DistributionForm to use useController instead of FormControl

sshmm:
## Motivation and Context
implements #1168

## Description
Update DistributionForm to use useController instead of FormControl

1- Removed FormControl from mui and Controller from react-hook-form
2- Used UseController from react-hook-form

## Test and Deployment Plan
1- Created a covault
2- test normal distribution and fixed payment distribution

image
image


levity:
@topocount maybe we can kill two birds with one stone and you can build on top of this when implementing disambiguating between Yearn-backed & simple token vaults with the same symbol?

levity:
merged to `almost-main` to run CI

PR #1259, opened 2022-08-11
Fix FormTokenField Decimal Handling

topocount:
FormTokenField was handling token decimals incorrectly. This fix makes
it impossible to enter a value with more decimals than a token has
available. The amount will remain at the value with the smallest
possible decimal. If a user pastes in a value with more decimals than is
can be handled, the amount is truncated to the valid decimal quantity.

Additionally, this refactor removes that annoying thing where the field
can't be empty and you need to "type around" a placeholder zero.
Validation treats empty fields correctly in my testing on deposits,
           withdraws and distributions.

This change also led to an upstream refactor to move all calculations
away from Numbers and to BigNumbers in all views that consume
FormTokenField.

Test Plan: unit and E2E tests should pass.

### DIFF
--- a/cypress/e2e/createVault.cy.ts
+++ b/cypress/e2e/createVault.cy.ts
@@ -36,7 +36,7 @@ context('Coordinape', () => {
 
     // Deposit USDC into the vault
     cy.contains('Deposit').click();
-    cy.get('input[type=number]', { timeout: 90000 })
+    cy.get('input[type=text]', { timeout: 90000 })
       .click()
       .wait(1000)
       .type('5000');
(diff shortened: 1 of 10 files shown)


levity:
i think i fixed the merge conflicts correctly--but i'd be more sure if there were unit tests 😉 it would be good especially to get a small test around the underflow error handling in FormTokenField

levity:
> > i think i fixed the merge conflicts correctly--but i'd be more sure if there were unit tests wink it would be good especially to get a small test around the underflow error handling in FormTokenField
> 
> 
> 
> Thanks @levity! Thats why this is in draft still!

Heh, I confess I overlooked the draft tag until after I fixed the merge 😅

levity (approved):

PR #1264, opened 2022-08-12
Add Members Page w/ just bulk add member by address

CryptoGraffe:
## Motivation and Context

Breaking off a chunk of the add members work to get deployed. This just includes the add one member or add many members by eth addr/name functionality.

This is based on our hybrid onboarding pod / 1.5 add members concepts.  Figma here: https://www.figma.com/file/qSap4xHYicSvZb3xXqmBzI/Onboarding-Wireframes?node-id=323%3A20725

This also includes feature flagging for the link sharing stuff, which is off for now and is half-baked because the UI isn't done. 

## Description

* New Add members page, which will have tabs for more styles of adding, but are hidden for now. 
* Getting rid of the add contributor dialog , which means adding users is now simple, and doesn't have advanced options

## Test and Deployment Plan

* I have tested this quite well, including validation varieties, but I could use thorough testing from a reviewer.  
* Also test editing members after simply adding them

### DIFF
--- a/cypress/e2e/createUser.cy.ts
+++ b/cypress/e2e/createUser.cy.ts
@@ -13,27 +13,35 @@ context('Coordinape', () => {
     });
   });
 
+  after(() => {
+    // might want something more surgical and lightweight
+    // to facilitate faster idempotent testing
+    // cy.exec('yarn db-seed-fresh');
+  });
   it('can create a new user with fixed payment amount', () => {
     cy.visit(`/circles/${circleId}/members`);
     cy.login();
 
-    cy.contains('Add Contributor', { timeout: 120000 }).should('be.visible');
-    cy.contains('Add Contributor').click();
-    cy.getInputByLabel('Contributor Name').type('A Test User').blur();
-    cy.getInputByLabel('Contributor ETH address')
-      .type('0xe00b84525b71ef52014e59f633c97530cb278e09')
-      .blur();
-    // enter the fixed payment amount
-    cy.getInputByLabel('Fixed Payment Amount').clear().type('1200').blur();
+    cy.contains('Add Members', { timeout: 120000 }).should('be.visible');
+    cy.contains('Add Members').click();
 
-    cy.contains('Save').click();
-    cy.reload(true);
+    cy.get('[data-testid=new-members]').within(() => {
+      cy.get('input').eq(0).click().type('A Test User').blur();
+      cy.get('input')
+        .eq(1)
+        .click()
+        .type('0xe00b84525b71ef52014e59f633c97530cb278e09')
+        .blur();
+    });
+
+    cy.get('button')
+      .contains('Add Members', { timeout: 120000 })
+      .should('be.enabled');
+    cy.get('button').contains('Add Members').click();
+    cy.contains('You have added 1 member!', { timeout: 120000 }).should(
+      'be.visible'
+    );
+    cy.get('button').contains('Back').click();
     cy.contains('A Test User', { timeout: 120000 }).should('be.visible');
-    // Verify new value in contributors table
-    cy.contains('A Test User', { timeout: 120000 })
-      .parents('tr')
-      .within(() => {
-        cy.get('td').eq(7).should('have.text', '1200');
-      });
   });
 });
--- a/src/config/features.ts
+++ b/src/config/features.ts
@@ -1,6 +1,11 @@
 // add your own feature names here
 
-export type FeatureName = 'vaults' | 'fixed_payments' | 'future_feature';
+export type FeatureName =
+  | 'vaults'
+  | 'fixed_payments'
+  | 'future_feature'
+  | 'csv_import'
+  | 'link_joining';
 
 // this is a very simple implementation of build-time feature flags that you can
 // hardcode or set with environment variables
@@ -8,6 +13,8 @@ export type FeatureName = 'vaults' | 'fixed_payments' | 'future_feature';
 const staticFeatureFlags: Partial<Record<FeatureName, boolean>> = {
   vaults: !!process.env.REACT_APP_FEATURE_FLAG_VAULTS,
   fixed_payments: !!process.env.REACT_APP_FEATURE_FLAG_FIXED_PAYMENTS,
+  csv_import: false,
+  link_joining: false,
 };
 
 // we make the export a function so that we can implement run-time feature flags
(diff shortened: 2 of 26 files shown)


CryptoGraffe:
This has a couple outstanding issues that I want to fix before review, should be done tomorrow morning:

* better error messages
* members table is not updating after adding members
* tests are broken

CryptoGraffe:
> This has a couple outstanding issues that I want to fix before review, should be done tomorrow morning:
> 
> * better error messages
> * members table is not updating after adding members
> * tests are broken

This is all resolved. ready for review.

src/pages/AddMembersPage/NewMemberList.tsx
```diff
@@ -0,0 +1,269 @@
...
+          })
+      )
+      .superRefine(data => {
+        if (data.filter(m => m.address != '' && m.name != '').length == 0) {
+          //TODO: I want to use this to prevent form submission of there are no valid entries
+          // but it just breaks the more useful errors from rendering
+          // ctx.addIssue({
+          //   code: z.ZodIssueCode.custom,
+          //   message: 'no valid members entered',
+          // });
```
topocount, line 76: I'm honestly not a huge fan of the the zodResolver myself. I think its a bit clumsy of an integration with react-hook-form. In cases like this where you want to prevent submission i create a clause in the submit button's `disabled` attribute.

src/config/features.ts
```diff
@@ -1,13 +1,20 @@
...
+  | 'link_joining';
 
 // this is a very simple implementation of build-time feature flags that you can
 // hardcode or set with environment variables
 
 const staticFeatureFlags: Partial<Record<FeatureName, boolean>> = {
   vaults: !!process.env.REACT_APP_FEATURE_FLAG_VAULTS,
   fixed_payments: !!process.env.REACT_APP_FEATURE_FLAG_FIXED_PAYMENTS,
+  csv_import: false,
+  link_joining: false,
```
topocount, line 17: why not use an env var here?

topocount (commented):
seems like there might be some unsaved changes in yarn.lock in this branch.

Super stoked about this and excited to play with it more!

CryptoGraffe:
re: the test , we don’t allow these parameters during creation anymore so I
moved it to the update user test

CryptoGraffe:
Re the env var cuz it’s mostly for me while hacking away and breaking this
work up

src/pages/AddMembersPage/NewMemberList.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  const submitNewMembers = async () => {
+    const newMembers = getValues('newMembers') as {
+      name: string;
+      address: string;
+    }[];
+    try {
+      setLoading(true);
+      setSuccessCount(0);
+      const filteredMembers = newMembers.filter(
```
zashton, line 114: Will you need to filter for duplicate name or addresses here ?

src/pages/AddMembersPage/NewMemberList.tsx
```diff
@@ -0,0 +1,269 @@
...
+        m => m.address != '' && m.name != ''
+      );
+      await client.mutate({
```
zashton, line 117: Should we show a meaningful error message if one of the address already exists in the circle and causes a constraint error for the whole batch or users ?

src/pages/AddMembersPage/useCircleTokens.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export const deleteWelcomeToken = (circleId: number) => {
+  return deleteToken(circleId, CircleTokenType.Welcome);
+};
+
+const useCircleTokens = (circleId: number, type: CircleTokenType) => {
+  return useQuery(
+    ['circle-token-', circleId, type],
+    async (): Promise<string> => {
+      const { circle_share_tokens } = await client.query({
```
zashton, line 28: will we need to check here if circle is already soft deleted ?

zashton (changes requested):

cypress/e2e/createUser.cy.ts
```diff
@@ -13,27 +13,35 @@ context('Coordinape', () => {
...
+      'be.visible'
+    );
+    cy.get('button').contains('Back').click();
     cy.contains('A Test User', { timeout: 120000 }).should('be.visible');
-    // Verify new value in contributors table
-    cy.contains('A Test User', { timeout: 120000 })
-      .parents('tr')
-      .within(() => {
-        cy.get('td').eq(7).should('have.text', '1200');
-      });
```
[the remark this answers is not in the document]
CryptoGraffe: it got moved to edit user since add is simplified

src/config/features.ts
CryptoGraffe: cuz its really just for me and my dependent branches

src/pages/AddMembersPage/NewMemberList.tsx
CryptoGraffe: hmm this is a good question, i was just using the existing bulkAdd code and didnt look at it too closely, i'll take a look now

src/pages/AddMembersPage/useCircleTokens.ts
CryptoGraffe: yeah i guess, that stuff got added after i built this (this code is getting old....)

src/pages/AddMembersPage/NewMemberList.tsx
CryptoGraffe: I think it does show a meaningful error on submit, i just tested this and it seems fine

src/pages/AddMembersPage/NewMemberList.tsx
CryptoGraffe: I dont think we do , this happens on the API side and it returns a decent error message

CryptoGraffe: I improved it a little bit

src/pages/AddMembersPage/useCircleTokens.ts
CryptoGraffe: ok i updated this

src/pages/AddMembersPage/NewMemberList.tsx
CryptoGraffe: ok i used a watch and moved to submit button, good call

zashton (approved):

PR #1266, opened 2022-08-12
Prepend 'Yearn - ' to Yearn Vaults in UI

topocount:
This allows us to distinguish between Yearn and simple token vaults in
the UI.

image


crabsinger:
These changes look good across the site - but it seems to break the Fixed Payment vault selection stuff, which I believe Wing is working on.

Screen Shot 2022-08-15 at 9 32 30 AM

crabsinger (approved):

PR #1267, opened 2022-08-12
bugfix: FormTokenField handles empty strings

topocount:
This used to throw an error when the string was parsed into a BigNumber.
Now it just short-circuits and doesn't attempt to parse.

### DIFF
--- a/src/components/FormTokenField/FormTokenField.tsx
+++ b/src/components/FormTokenField/FormTokenField.tsx
@@ -49,7 +49,7 @@ export const FormTokenField = React.forwardRef((props: Props, ref) => {
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     // only accept floating point inputs
     const matches = e.target.value.match(/\d*\.?\d*/);
-    if (matches == null) {
+    if (matches == null || e.target.value.length === 0) {
       onChange('');
       return;
     }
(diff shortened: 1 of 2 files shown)


src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -454,6 +454,7 @@ export function DistributionForm({
                   )
                 )} ${giftVaultSymbol}`}
                 onChange={value => {
+                  if (value == '') value = '0';
```
crabsinger, line 457: @topocount I needed to add this to fix the error I was encountering. Not sure if this is the best approach to this, but seems to work for now.

crabsinger (approved):

src/components/FormTokenField/FormTokenField.tsx
```diff
@@ -49,8 +49,8 @@ export const FormTokenField = React.forwardRef((props: Props, ref) => {
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     // only accept floating point inputs
     const matches = e.target.value.match(/\d*\.?\d*/);
-    if (matches == null) {
-      onChange('');
+    if (matches == null || e.target.value.length === 0) {
+      onChange('0');
```
crabsinger, line 53: The onChange to '' still causes the error, so we actually want a 0 value to be used in this case so our math works.

crabsinger (approved):

PR #1268, opened 2022-08-13
Admin Settings: 1) Fixed payments vault field,  2) Token name for CSV export field

wingfeatherwave:
Screen Shot 2022-08-12 at 3 38 48 PM
Screen Shot 2022-08-12 at 3 20 18 PM
 Motivation and Context

References: https://www.notion.so/coordinape/Cannot-use-Simple-Token-Vault-for-Fixed-payments-ff90ebe6bbc94beba09d61dd17c6cdf2

Problem: The Fixed Payment Vault Token autocomplete field was overloaded and confusing 
Solution: 

* Break up fixed payments token fields into two fields
* Use stitches select component to show available vaults to select and effectively enable fixed payments
* new column: circle.fixed_payment_vault_id
* repurpose fixed_payment_token_type for use in csv export only
* when vault is selected, csv token is disabled and uses the vault token...
* otherwise, when vault is not selected, csv token can be named whatever the heck you like

@zashton you can use this `<Select>`  component for vault selection on the distributions page


zashton (approved):
made some minor updates, new input field works for me.

zashton:
@wingfeatherwave btw for the select theres no disabled state and the ability to set labels like other input components

crabsinger:
Tested the migrations locally and they look good (both up and down)

crabsinger (changes requested):

crabsinger (approved):

PR #1275, opened 2022-08-15
Fixup/settings page errors simple

crabsinger:
Fix bugs introduced in #1268 

- non of this should load on mainnet if flags arent enabled
- fix race condition bug vault selection


src/hooks/gql/useVaults.ts
```diff
@@ -60,9 +65,9 @@ export function useVaults({
       return vaults;
     },
     {
-      enabled: !!orgId && !!chainId,
+      enabled: !!orgId && !!chainId && isFeatureEnabled('vaults'),
```
crabsinger, line 68: is this necessary?

crabsinger: tihink so.

wingfeatherwave (dismissed):
nice work!

wingfeatherwave:
🤝

wingfeatherwave (approved):
fo real this time

src/hooks/gql/useVaults.ts
```diff
@@ -16,6 +18,9 @@ export function useVaults({
   return useQuery(
     ['vaults-for-org-', orgId],
     async () => {
+      if (!isFeatureEnabled('vaults')) {
```
levity, line 21: this isn't redundant with the check in `enabled` below?

PR #1238, opened 2022-08-05
Distributions Page: handle deleted users edge case

crabsinger:
# Support Deleted Users on Distributions Page

When there are deleted users in a past epoch, render the page without errors, and show the number of deleted users and sum of GIVE allocated to all deleted users.

Screen Shot 2022-08-12 at 6 05 37 PM

@wingfeatherwave ideas on a way to style the "Note" a bit better?

Notes:
- [x]  remove deleted users from the table
- [x]  show num deleted users received X GIVE in text below.
- [x]  table should have % give sum to 100
- [x]  no fixed payments to deleted users
- [x] only submit claims that have a valid, no longer create 0.0 amount new claims for deleted users or users who are not receiving any new amount this distribution

Previously, we created 0 new_amount claims for all users in the circle every time, now its filtered to only claims with a new amount.

PR #1277, opened 2022-08-16
bugfix: enable FormTokenField to have an empty input

topocount:
One of the major considerations during the FormTokenField (and
downstream) refactor was allowing for empty string inputs. This
enables for much cleaner UX, since users aren't typing "around" a zero
and in my opinion, makes us look like we know what we're doing as devs.
There are other hacks that could make the UX better than the incumbent,
but why not just make it behave correctly and be done with it?

Test plan: all unit tests and E2E tests should pass


topocount:
attempting to address the failing tests here in #1278

crabsinger (approved):

crabsinger:
It seems like https://github.com/coordinape/coordinape/pull/1238/commits/949a0b587f1b809e4538ab8fe4e66868d8c7096a resolves this issue on the other branch, and will allow us to support empty strings.

I'm unsure why `fixup/allow-deleted-users-vaults` caused an error with the boundary case when `main` didn't raise an error, even though `fixup/allow-deleted-users-vaults` didn't seemingly change anything around the renderCombinedSum function.

crabsinger:
Updating this again off main - will see if the tests pass now or not...

zashton (approved):

PR #1280, opened 2022-08-16
Disabled state for radix select.  Label for radix select

wingfeatherwave:
Adding some props for the `<Select>`


wingfeatherwave:
@zashton I've added the variant `invertForm` which can be applied to `<Panel invertForm>`. This will give the white bg form inputs, and when they are disabled they will get a border color with transparent background, and will not receive opacity adjustment.

crabsinger (approved):

crabsinger:
Code looks good - I didn't test it locally tho.

zashton (approved):

PR #1282, opened 2022-08-17
update transaction descriptions to include circle and epoch info

teeolendo:
## Motivation and Context

Improve Recent Transaction info by adding circle and epoch ID to the modal

## Description

Updated the `useSubmitDistribution` and `useClaimAllocation` hooks to include more info. Also update the tests for the same file and they all work as expected.

image

### DIFF
--- a/src/pages/ClaimsPage/useClaimAllocation.ts
+++ b/src/pages/ClaimsPage/useClaimAllocation.ts
@@ -63,7 +63,7 @@ export function useClaimAllocation() {
         {
           showInfo,
           showError,
-          description: 'Claim Tokens',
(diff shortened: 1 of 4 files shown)


levity (changes requested):
Epoch ID isn't a user-facing value; you want to use `epoch.number` or a date range instead

crabsinger:
> Epoch ID isn't a user-facing value; you want to use `epoch.number` or a date range instead

Good catch - went ahead and added a fix for that real quick.

crabsinger (approved):

crabsinger (dismissed):

PR #1285, opened 2022-08-17
Fix numberWithCommas to handle precision properly

crabsinger:
Fix the numberWithCommas util to handle some edge cases. Add tests.

For example before this numberWithCommas(6, 2) returns 6, not 6.00.


levity (changes requested):
there's already a `utils/index.test.ts`; please relocate the tests there. 

note about test style: i personally think it's ok to have a single test or a few tests with a number of assertions in them, especially when the assertions are self-explanatory and/or test names would be redundant with test content.

topocount (dismissed):
thanks, this is great!

src/utils/index.test.ts
```diff
@@ -1,8 +1,39 @@
 import { numberWithCommas } from './';
 
-test('numberWithCommas', () => {
```
crabsinger, line 3: @levity @topocount how's this for test styling?

crabsinger: happy for any nitty feedback on making it even better :)

zashton (approved):

Issue #982, opened 2022-05-31
1.5.6- Fixed Payments Sections


reeserj:
Dependent on https://github.com/coordinape/coordinape/issues/962 and https://github.com/coordinape/coordinape/milestone/7 

Any data fetching to be rebuilt using Zeus

[Figma](https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=5303%3A21149)

Token Address should return the token name after entered
Members count of members on fixed payments
Fixed payments total = Sum of fixed payments per epoch
Available in vault should read 0 when no vault available.

EsteFilipe:
@reeserj interested in working on this. does #962 have to be finished first for this is workable on?

reeserj:
I think you could work on it, but would need to check in with @sshmm to ensure you're aligned. I'll assign it to you

crabsinger:
@FilipeAleixo #962 is merged. Do you still want to take this ticket and can begin working on it now? Or shall we assign to another person?

sshmm:
@FilipeAleixo  please sync with #1056 until it's reviewed and merged if you are going to work on it

FilipeAleixo:
@crabsinger I'm currently short on bandwith, so it's better assigning to someone else

crabsinger:
@spydermonkee is this a blocker for Vaults launching?

sshmm:
can I work on this?

reeserj:
You got it @sshmm

sshmm:
closed by #1287

PR #1287, opened 2022-08-18
Feature/Circle admin page fixed payment section

sshmm:
## Motivation and Context

implements #982 https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=5303%3A21149

## Description

1- Add Members, Fixed payments total and available in vault fields to Fixed Payment section
2- Added a new query to fix the number of fixed payment receivers and used that to calculate the total and the number of members
3- Reused some functions from DistributionForm to calculate the total in vaults

## Test and Deployment Plan

1- Reach Circle settings page and add a token name to the field Token name for CSV export
2- Add fixed Payment to multiple users in the circle
expected:
The total of fixed payments and the number of members should be shown in Circle Settings fixed payment section
actual:
image

3- Add vaults the circle and test with the token symbol of these vaults
image
image

### DIFF
--- a/src/pages/CircleAdminPage/CircleAdminPage.tsx
+++ b/src/pages/CircleAdminPage/CircleAdminPage.tsx
@@ -371,6 +371,20 @@ export const CircleAdminPage = () => {
     }
   };
 
+  const fixedPaymentToken = (vaultId: string | undefined) => {
+    const tokenType = circle?.fixed_payment_token_type;
+    return vaultId
+      ? removeYearnPrefix(
+          vaultOptions.find(o => o.value == getValues('fixed_payment_vault_id'))
+            ?.label ?? ''
+        )
+      : tokenType
+      ? tokenType.startsWith('Yearn')
+        ? removeYearnPrefix(tokenType)
+        : tokenType
+      : '';
+  };
+
   if (
     isLoading ||
     isIdle ||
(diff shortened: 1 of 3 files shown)


src/pages/CircleAdminPage/CircleAdminPage.tsx
```diff
@@ -323,13 +346,48 @@ export const CircleAdminPage = () => {
...
+  }: {
+    vaultId?: number | undefined;
+    symbol?: string | undefined;
+  }) => {
+    return vaultsQuery?.data?.find(v =>
+      vaultId ? v.id === vaultId : v.symbol === symbol
+    );
+  };
+
+  const updateBalanceState = async (symbol: string): Promise<void> => {
```
zashton, line 369: i've updated this in DistributionForm to take in the `vault_id` instead of `symbol` so it can handle multiple vaults of the same token & if there are tokens of the same symbol

src/pages/CircleAdminPage/CircleAdminPage.tsx
```diff
@@ -323,13 +346,48 @@ export const CircleAdminPage = () => {
     }
   };
 
+  const getDecimals = ({ symbol }: { symbol: string | undefined }) => {
```
zashton, line 349: this has been updated in DistributionForm to take in `vault_id` instead

zashton (changes requested):

src/pages/CircleAdminPage/CircleAdminPage.tsx
sshmm: done

src/pages/CircleAdminPage/CircleAdminPage.tsx
sshmm: done

zashton (approved):

PR #1288, opened 2022-08-18
Select Radix Import Fix

wingfeatherwave:
@levity is this the right correction for the import?  Variables seem to work properly


levity (approved):
Yep! Nice, didn't notice you were writing stories already

wingfeatherwave:
> Yep! Nice, didn't notice you were writing stories already

just modified an existing one

PR #1289, opened 2022-08-18
Update label for distribution form token input

zashton:
Reduce the text in the pre-label/label to free up more space to display token balance

Fixes one of issues in the UI nits


crabsinger (approved):
looks good

PR #1294, opened 2022-08-19
Refactor: expose all public vault info and conceal org id

topocount:
## Motivation and Context

Most of this data is public anyway. The org_id is the only thing worth hiding, since its entirely internal.

### DIFF
--- a/hasura/metadata/databases/default/tables/public_vaults.yaml
+++ b/hasura/metadata/databases/default/tables/public_vaults.yaml
@@ -32,23 +32,10 @@ select_permissions:
     - decimals
     - deployment_block
     - id
-    - org_id
     - simple_token_address
     - symbol
     - token_address
     - updated_at
     - vault_address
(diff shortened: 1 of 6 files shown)


hasura/metadata/databases/default/tables/public_vaults.yaml
```diff
@@ -32,23 +32,10 @@ select_permissions:
     - decimals
     - deployment_block
     - id
-    - org_id
```
crabsinger, line 35: should we expose org_id to users who are part of this, so we dont need to change the other queries in this pr?

zashton (approved):

hasura/metadata/databases/default/tables/public_vaults.yaml
topocount: I think this is an information leak. If a user has access to the org row, they can pull the id from the row.

PR #1300, opened 2022-08-22
remove extra query for claims button

levity:
## Motivation and Context

why run two separate queries on page load when one will do? instead of running a separate query in the claims button to determine whether it should be shown, combine that with the query that gets the list of orgs & circles.

cleanup: also pull some of the components that were in MainHeader.tsx into their own files

[Notion: Vaults QA item](https://www.notion.so/coordinape/226b9363ee0d45cead60e5365ae65d10?v=2137f63c15394b32b90fb80dd09a5244&p=e2825cd1967e419498d731b6a5eeff8c&pm=s)

### Notes for reviewers

Changes in OverviewMenu:
- moved the file 
- moved querying to MainHeader
- removed two unneeded top-level tags `<>` and `<Hidden>`
- The rest of the diff is just due to re-indentation


zashton (approved):
Just tried works for me , small nit is that i need to refresh the page after a distribution to make the claim button appear.

levity:
> Just tried works for me , small nit is that i need to refresh the page after a distribution to make the claim button appear.

ah, good catch

PR #1309, opened 2022-08-24
Allow access to distribution page form when there are no allocations

zashton:
## Motivation and Context

https://www.notion.so/coordinape/VAULTS-QA-dd5a668750314cbc8f9a431fac1132a0?p=d04cef68e3184b2187e504398d8bbc06&pm=s

## Description

This change allows distribution page to be accessed even if epoch has no allocations, so the admin still is able to make a fixed payments distribution if desired

## Test and Deployment Plan

Added a unit test for a test case with no allocations, alternatively you can test it manually by ending an epoch and going to the distribution page after creating a vault+ deposit

### DIFF
--- a/src/pages/DistributionsPage/DistributionForm.tsx
+++ b/src/pages/DistributionsPage/DistributionForm.tsx
@@ -357,11 +357,8 @@ export function DistributionForm({
     }
   };
 
-  const shouldDisableGiftInput = () => {
-    return (
-      giftSubmitting || !!circleDist || vaults.length === 0 || totalGive === 0
-    );
-  };
+  const shouldDisableGiftInput =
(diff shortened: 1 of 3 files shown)


src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -355,6 +357,12 @@ export function DistributionForm({
     }
   };
 
+  const shouldDisableGiftInput = () => {
```
levity, line 360: i don't think this needs to be a function

zashton: updated

levity (approved):

PR #1311, opened 2022-08-24
Fix number with commas utils function

zashton:
## Motivation and Context

Decimal numbers that start with zeroes after the period will end up with an incorrect rounded result.

## Description

Pad the zeroes back after it gets converted into a number for rounding.

## Test and Deployment Plan

added more test cases in the unit test with numbers that fit this scenario.

## Related Issue

https://www.notion.so/coordinape/VAULTS-QA-dd5a668750314cbc8f9a431fac1132a0?p=52af939d7d1a4cf8aeabb777293e5908&pm=s


CryptoGraffe (approved):
Looks good to me except i'm confused why its called numberWithCommas instead with decimals? idk

zashton:
> Looks good to me except i'm confused why its called numberWithCommas instead with decimals? idk

it also add commas to large numbers for easier reading on top of doing decimal precision

PR #1321, opened 2022-08-25
Update receive popover to use new Radix/Stitches component.  show it

wingfeatherwave:
### g'bye more material ui

Screen Shot 2022-08-25 at 11 59 16 AM


src/components/ReceiveInfo/ReceiveInfo.tsx
```diff
@@ -112,82 +32,113 @@ export const ReceiveInfo = () => {
...
+          onMouseEnter={() => setMouseEnterTrigger(true)}
+          onMouseLeave={() =>
+            setTimeout(() => setMouseEnterTrigger(false), 200)
+          }
+        >
+          {totalReceived} {selectedCircle?.tokenName}
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent
+        onMouseEnter={() => setMouseEnterPopover(true)}
```
CryptoGraffe, line 57: should this cancel the timeout that is going to dismiss?

CryptoGraffe: e.g. if(timeoutId) { clearTimeout(timeoutId) }

src/components/ReceiveInfo/ReceiveInfo.tsx
```diff
@@ -112,82 +32,113 @@ export const ReceiveInfo = () => {
...
+        onMouseLeave={() => setTimeout(() => setMouseEnterPopover(false), 200)}
+        align="end"
+        sideOffset={-37}
```
CryptoGraffe, line 60: this magic number is interesting, is there any other way to do it?

wingfeatherwave: Sadly, I don't know how to get around this magic number.  It really stinks.  It's for the purpose of overlaying the popover and the trigger

src/components/ReceiveInfo/ReceiveInfo.tsx
wingfeatherwave: I don't think it needs to.  It's all overwriting, and I think that if it dismisses it, it might introduce a flash.  not sure

src/components/ReceiveInfo/ReceiveInfo.tsx
CryptoGraffe: k

src/components/ReceiveInfo/ReceiveInfo.tsx
CryptoGraffe: we discussed this and it got addressed

CryptoGraffe (approved):

Issue #1314, opened 2022-08-24
Circle Members Header


wingfeatherwave:
- [x] Copy `Nominate New Member` button from /vouching to `Nominate Member` button on /members
- [x] Remove icon & tooltip from `Add Members` button
- [x] Scope check: only admins can see `Add Members` button.  Everyone can see `Nominate Member` button
- [x] Header button styles should match Figma `<Button size="small" ...>`
- [x] Show member count next to member header buttons.
- [x] Show nominee count next to member header buttons.  (Currently, when there is a nomination we show that information on /circles for the circle with a nominee.  Need it in both places)  
- [x] Add relevant design/copy from Figma

Figma: https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/App-1.5?node-id=11000%3A30373

PR to `members-page-updates` instead of main until all of 1.5.5 is complete

sshmm:
@wingfeatherwave does that mean regular users will have access to members page? 
can they see the full list of members?

spydermonkee:
> @wingfeatherwave does that mean regular users will have access to members page? can they see the full list of members?

Yep- All members will be able to see the members of the circle in this view, not just the admins. Note that this is only the members list for the circle, not the whole org.

sshmm:
@spydermonkee  can I work on it?

spydermonkee:
@sshmm - I think so! I'll let @wingfeatherwave chime in if he wants to do something different.

wingfeatherwave:
Nice work.  Merged into `coordinape:members-page-updates`

PR #1328, opened 2022-09-01
Circle members header

sshmm:
## Motivation and Context

implements #1314 

## Description

1- Add Nominate Member to Members page
2- make members page available to all users
3- make member settings and add member only available to admins
4- hide gives and fixed payment columns from normal members
5- show total number of nominees and members at the page header
6- Nominees will appear on vouching page until it is replaced by cards in members page
7- Update mobile layout

## Test and Deployment Plan
1- Enter Members page as an admin with vouching enabled
image
image

2- Enter Members page as an admin with vouching disabled
image
image

3- Enter as Member with vouching enabled
image
image

4- Enter as Member with vouching disabled
image

5- Nomination Modal
image


wingfeatherwave:
changing the base to `members-page-updates` instead of `main`

wingfeatherwave (approved):
nice work!

PR #1330, opened 2022-09-02
Fix Amount and vault select fields alignment in distribution form

sshmm:
## Motivation and Context

Fields were not aligned in horizontal axis in Distribution Form
image

## Description
1-  remove baseline alignment as it was causing the fields to be not aligned correctly

## Test and Deployment Plan
1- Go to Distribution Page and check the alignments of Covault and amount fields

image
image

### DIFF
--- a/src/pages/DistributionsPage/DistributionForm.tsx
+++ b/src/pages/DistributionsPage/DistributionForm.tsx
@@ -376,7 +376,7 @@ export function DistributionForm({
           <Text h2 css={headerStyle}>
             Gift Circle
           </Text>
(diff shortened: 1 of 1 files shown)


wingfeatherwave (approved):
I haven't been able to reproduce this bug in `main`.  when is it that they are out of alignment?  I added the align baseline for scenarios in which the available number caused the label to wrap to two lines.  But now that we are using "avail." instead of "available" the issue doesn't really come up.  so I'll approve and merge, and we can revisit if we need two-line labels later"

PR #1332, opened 2022-09-03
deleting single file directories in the components directory

CryptoGraffe:
## Motivation and Context

Trying to follow our desired pattern of "don't make a directory for one file unless its a page"

## Description

Moved up a bunch of files in the components dir and deleted their dirs


crabsinger (dismissed):
lets ship this!!

levity (approved):

PR #1339, opened 2022-09-05
Add Interaction_Events table and reporting

crabsinger:
## Motivation and Context

Introduce an `interaction_events` table to log key interaction events for a user. During key interactions such as create circle, add members we add a record to this table.

In a background worker log these interactions to Amplitude and Mixpanel for onboarding user conversion. We do not send any personally identifiable data to these services.

## Test and Deployment Plan

We need to set these env vars for staging/prod in Vercel

## Reviewers

@CryptoGraffe


api-lib/event_triggers/interactionEventToAmplitude.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    if (event.circle_id) {
+      await client.logEvent({
+        event_type: `circle_${event.event_type}`,
+        user_id: sha256(`${event.circle_id}`),
+        event_properties: {
+          event_subtype: event.event_subtype,
+          id: event.id,
+          data: event.data,
```
crabsinger, line 46: maybe also add user_id here?

api-lib/event_triggers/interactionEventToAmplitude.ts
```diff
@@ -0,0 +1,63 @@
...
+          circle_id: event.circle_id,
```
crabsinger, line 47:
```suggestion
          circle_id: event.circle_id,
          user_id: sha256(`${event.circle_id}`),
```

api-lib/types.d.ts
```diff
@@ -32,6 +32,7 @@ type RelationTypes =
   | 'profile'
   | 'nominations'
   | 'nominations_aggregate'
+  | 'user_events'
```
crabsinger, line 35: interaction_event

hasura/migrations/default/1662391842396_interaction_events/down.sql
```diff
@@ -0,0 +1,47 @@
+
+alter table "public"."interaction_events" rename to "user_events";
```
crabsinger, line 2: lets just clean this up to be one single migration rather than a create and a rename

api-lib/types.d.ts
crabsinger: fixed

hasura/migrations/default/1662391842396_interaction_events/up.sql
```diff
@@ -0,0 +1,37 @@
...
+  "event_type" text NOT NULL,
+  "event_subtype" text NULL,
+  "data" jsonb NULL,
+  "circle_id" integer NULL,
+  "org_id" integer NULL,
+  "id" serial NOT NULL,
+  PRIMARY KEY ("id")
+);
+
+CREATE OR REPLACE FUNCTION "public"."set_current_timestamp_updated_at"
```
levity, line 15: remove this

api-lib/event_triggers/interactionEventToMixpanel.ts
```diff
@@ -0,0 +1,89 @@
...
+  return new Promise<void>(resolve => {
+    mp.track(
+      eventName,
+      {
+        distinct_id: distinctId,
+        ...props,
+      },
+      () => {
+        resolve();
+      }
```
levity, line 86:
```suggestion
      resolve,
```

api-lib/gql/mutations.ts
```diff
@@ -479,3 +479,27 @@ export async function endNominees(circleId: number) {
...
+export async function insertInteractionEvent(
+  event: Omit<
+    GraphQLTypes['interaction_events_insert_input'],
+    'created_at' | 'id' | 'updated_at'
+  >
+) {
+  await adminClient.mutate(
+    {
+      insert_interaction_events_one: [
+        {
```
levity, line 492: don't split object literals across multiple lines unnecessarily pleae

api/hasura/actions/_handlers/createUsers.ts
```diff
@@ -153,6 +155,20 @@ async function handler(req: VercelRequest, res: VercelResponse) {
       `panic: insertedUsers is null; ${JSON.stringify(mutationResult, null, 2)}`
     );
 
+  let profileId: number;
+  if (sessionVariables.hasuraRole == 'user') {
+    profileId = sessionVariables.hasuraProfileId;
+  }
+
+  insertedUsers.map(async inviteduserId => {
```
levity, line 163: let's use `await Promise.all(insertedUsers.map...)` here so we don't have a delay time that grows linearly with the number of users... or better yet, is there a syntax for batch-inserting in a single hasura request?

api/hasura/event_triggers/eventManager.ts
```diff
@@ -25,6 +27,8 @@ const HANDLERS: HandlerDict = {
   createNomineeTelegram,
   createCircleCRM,
   createVouchedUser,
+  interactionEventToAmplitude,
+  interactionEventToMixpanel,
```
levity, line 31:
```suggestion
  sendEventToAmplitude,
  sendEventToMixpanel,
```

just a suggestion, since the convention here is to start with a verb

api/login.ts
```diff
@@ -88,16 +89,41 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
...
-          insert_profiles_one: [{ object: { address: address } }, { id: true }],
+          insert_profiles_one: [
+            { object: { address: address } },
+            {
+              id: true,
+              users: [
+                {},
+                {
+                  circle_id: true,
+                },
```
levity, line 100:
```suggestion
                { circle_id: true },
```

api/login.ts
```diff
@@ -88,16 +89,41 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
...
+            ? insert_profiles_one.users.length > 0
+            : false,
+          brandNewMember: insert_profiles_one.users
+            ? insert_profiles_one.users.length == 0
+            : false,
+        },
+      });
+      // if they have no users, this is a "clean signup"
     }
     assert(profile, 'panic: profile must exist');
```
levity, line 126: remove this assertion now that it's redundant

api/login.ts
```diff
@@ -88,16 +89,41 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
...
         }
       );
+      assert(insert_profiles_one, "panic: adding profile didn't succeed");
       profile = insert_profiles_one;
+      await insertInteractionEvent({
+        event_type: 'first_login',
+        profile_id: insert_profiles_one.id,
+        circle_id: insert_profiles_one.users?.[0]?.circle_id,
+        data: {
+          invitedMember: insert_profiles_one.users
```
levity, line 116: since these two props are exclusive, maybe what we really want is a boolean e.g. `brandNew`? 

actually i'm a bit confused by the logic here; is there a non-exceptional case when these should both be false? should it instead be checking `(insert_profiles_one.users?.length || 0)` instead of using a ternary, so one of these is always true?

api-lib/event_triggers/interactionEventToMixpanel.ts
```diff
@@ -0,0 +1,89 @@
...
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+import mixpanel, { Mixpanel } from 'mixpanel';
+
+import { MIXPANEL_PROJECT_TOKEN } from '../config';
+import { errorResponse } from '../HttpError';
+import { EventTriggerPayload } from '../types';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  try {
+    if (MIXPANEL_PROJECT_TOKEN == '') {
```
levity, line 12: maybe this should be an error that bubbles up to Sentry instead of failing silently?

levity (changes requested):

api/hasura/actions/_handlers/createUsers.ts
zashton: you should be able to batch insert like this.
```ts
insert_interaction_events: [
  { object: newEvents },
  { returning: {id: true  }}
]
```

zashton (changes requested):

hasura/migrations/default/1662391842396_interaction_events/down.sql
CryptoGraffe: this is cleaned up

hasura/migrations/default/1662391842396_interaction_events/up.sql
CryptoGraffe: done

api-lib/event_triggers/interactionEventToMixpanel.ts
CryptoGraffe: typing is unhappy with this

CryptoGraffe: image

api-lib/gql/mutations.ts
CryptoGraffe: ok

api/hasura/actions/_handlers/createUsers.ts
CryptoGraffe: i made this use the multiple insert model throughout with a varargs pattern

api/hasura/event_triggers/eventManager.ts
CryptoGraffe: renamed

api/login.ts
CryptoGraffe: done

api/login.ts
CryptoGraffe: done

api-lib/event_triggers/interactionEventToMixpanel.ts
CryptoGraffe: didnt want to break this on dev by throwing errors but i could see how that should work (since sentry isnt hooked up on dev)

CryptoGraffe: ok i changed this in both of the event triggers

api/login.ts
CryptoGraffe: ok i collapsed this to just `brandNew`

zashton (approved):

levity (approved):

hasura/migrations/default/1662391842396_interaction_events/up.sql
topocount: @levity why remove this trigger? do we have a trigger covering all tables already?

levity: yeah, this is the one that's caused deployment failures in the past because it's not owned by the current psql user

api/hasura/actions/_handlers/createCircle.ts
```diff
@@ -66,15 +67,24 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
   return async (fileName: string) => {
-    return await mutations.insertCircleWithAdmin(
+    const circle = await mutations.insertCircleWithAdmin(
       circleInput,
       userAddress,
       coordinapeAddress,
       fileName
     );
+
+    await mutations.insertInteractionEvents({
```
levity, line 81: @crabsinger i just noticed this while working on some related code... for performance, i think it would be ideal to bundle this with the request above rather than create another request-response loop. no need to fix this now, just flagging the general pattern.

crabsinger: @levity yes it would be more performant... but not sure how to batch these since we still want the sql logic separated in different mutations? ... 

do you see a clean pattern that would support this?

Idon't know if there is a Hasura "batch" syntax or something...

levity: > we still want the sql logic separated in different mutations?

not sure what you mean, but maybe we don't? there's a standard GraphQL syntax for doing multiple queries or mutations in a single request; you can see an example [here](https://github.com/coordinape/coordinape/blob/main/api/hasura/cron/recoverTransactions.ts#L150)

PR #1340, opened 2022-09-08
Fix Discord Webhook circle setting issue 

### LINKED ISSUE #1329: Discord Webhook circle setting issue
**Summary:**

Discord webhook field gets set to null when other circle settings gets updated. 

Discord webhook value should remain as it is.

3. **Steps to recreate:**
Change any input value in the circle settings page and click save. After save is completed go click on the Edit Discord Webhook button.
 
image

sshmm:
## Motivation and Context

Fixes #1329

## Description

~~1- add permission to fetch discord_webhook column
2- use the fetched value to provide default value for the input controller so that the value will be replaces with empty string upon not providing it with every change~~
1- Check if editwebhook button is clicked and only use the input field data if it is clicked, otherwise fetch the value from db 

## Test and Deployment Plan
1- Add a discord webhook then save
2- refresh the page to reset the controllers and fetched data
3- Change any input value in the circle settings page and click save. After save is completed go click on the Edit Discord Webhook button.

### Expected 
the discord webhook should remain unchanged

image


hasura/metadata/databases/default/tables/public_circles.yaml
```diff
@@ -120,6 +120,7 @@ select_permissions:
     - created_at
     - default_opt_in
     - deleted_at
+    - discord_webhook
```
zashton, line 123: The column should not be exposed to the public as the discord bot could be compromised 
Currently it is only available to the circle admins via the `circle_private` view

zashton (changes requested):

hasura/metadata/databases/default/tables/public_circles.yaml
sshmm: @zashton  thank you. I reverted this changes and came up with a much simpler solution please recheck

zashton (approved):

PR #1346, opened 2022-09-14
Create homebrewed dev stack server

topocount:
Run `yarn start` to try it out!

## Motivation

vercel was a great drop-in batteries-included dev server for our
serverless functions, but it required a bunch of external config that
was really painful to utilize without an internet connection, especially
when setting up a repo locally.

## Details

This new dev server handles all requests to serverless functions, except
for cron tasks, and proxies all other requests to the webpack dev server.
Nodemon is utilized to hot-reload backend code during handler development.

## Test Plan 
This is a drop-in replacement for the existing `vercel dev`
command and has been set up to work with E2E tests

## Future Work

Route(s) for the cron tasks need to be created at some point as work
evolves on those


crabsinger:
OMG - this is 1000 times faster!

scripts/serve_dev.ts
```diff
@@ -0,0 +1,61 @@
...
+});
+
+const proxy = createProxyMiddleware({
+  target: `http://localhost:${proxyPort}`,
+});
+
+app.use('/', proxy);
+
+app.listen(port, () => {
+  console.log(`Coordinape listening on port ${port}`);
```
crabsinger, line 60: I was initially confused by the console output and thought that the craco output 

```
Compiled successfully!

You can now view coordinape-app in the browser.

  Local:            http://localhost:3999
  On Your Network:  http://192.168.50.85:3999
```

Meant I should browse localhost:3999.

Visiting localhost:3999 renders the page but it errors on login.

I wonder if we could somehow make it more clear that devs should use port 3000 and not the proxy port.

Ideal would be to remove the craco output, but not sure how do to that.

Another option is just to add more info here, like this:

```
  console.log(
    `\n============================================================\n`
  );
  console.log(`Development server has started successfully!`);
  console.log(`Visit http://localhost:${port} to view the Coordinape app.`);
  console.log(`\n============================================================`);
```

crabsinger (approved):
@topocount amazing! this is SOO MUCH FASTER. Local page loads are now seemingly instant.

One note about the confusion if you visit the craco port directly, and thinking through how to make it more obvious to NOT go to the port directly and ALWAYS use the proxy...

levity (approved):

PR #1348, opened 2022-09-14
Update Contributions table to decouple epochs

topocount:
The contributions table will rely on a link to a specific circle instead
of a specific epoch and will use a possibly user-adjustable datetime_created
field to determine which epoch a contribution should be a part of.

Future Work

Build the Contributions interface on top of this table


hasura/migrations/default/1663180519908_contributions_decouple_epochs/down.sql
```diff
@@ -0,0 +1,9 @@
+
+alter table "public"."contributions" drop constraint "contributions_user_id_fkey";
+
+
+DROP TABLE "public"."contributions";
+
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
```
crabsinger, line 8: nit: remove this boilerplate

hasura/migrations/default/1663180519908_contributions_decouple_epochs/up.sql
```diff
@@ -0,0 +1,33 @@
...
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "updated_at" timestamptz NOT NULL DEFAULT now(),
+  "deleted_at" timestamptz,
+  PRIMARY KEY ("id"),
+  FOREIGN KEY ("circle_id") REFERENCES "public"."circles"("id") ON UPDATE restrict ON DELETE restrict,
+  FOREIGN KEY ("user_id") REFERENCES "public"."users"("id") ON UPDATE restrict ON DELETE restrict,
+  UNIQUE ("id")
+);
+CREATE OR REPLACE FUNCTION "public"."set_current_timestamp_updated_at"()
+RETURNS TRIGGER AS $$
```
crabsinger, line 19: I think @levity and @CryptoGraffe didnt like this trigger in the other interaction_events PR, but not sure why not...

hasura/migrations/default/1663180519908_contributions_decouple_epochs/up.sql
```diff
@@ -0,0 +1,33 @@
+DROP table "public"."contributions";
```
crabsinger, line 1:
```suggestion
DROP TABLE IF EXISTS "public"."contributions";
```

crabsinger: this errored on me while i tested going up and down the migrations.

crabsinger (changes requested):
Looks good. One suggestion that was erroring in the migrations.

One question for later: maybe we should allow contributions to be "added" to an epoch, rather than dynamically show variations of contributions depending on time functions? But I guess this will just depend on the UI and decisions made aroudn the exact logic later in implementation, so not needed to figure out now.

hasura/metadata/databases/default/tables/public_contributions.yaml
```diff
@@ -2,67 +2,52 @@ table:
...
-      columns:
-        - deleted_at
+        check: null
+        columns:
+        - datetime_created
         - description
-      filter:
-        _and:
-          - user:
+        filter:
```
zashton, line 49: might need to add filter to prevent updating contributions that epochs that have ended.

zashton (changes requested):
I got other 2 other open PRs thats related to contribution backend that seem to be in conflict, might be able to still use the delete contribution PR though.

topocount:
@zashton yeah the delete one still looks promising! Sorry about this miscommunication. The decision to detach epochs from contributions came out of a discussion at the retreat. We want to detach a lot of activity from epoch logic as a dependency. I like the idea of committing contributions to an epoch, and will find a way to prevent contributions from being modified after an epoch ends.

zashton:
> @zashton yeah the delete one still looks promising! Sorry about this miscommunication. The decision to detach epochs from contributions came out of a discussion at the retreat. We want to detach a lot of activity from epoch logic as a dependency. I like the idea of committing contributions to an epoch, and will find a way to prevent contributions from being modified after an epoch ends.

hey no prob man, are you making the permission update in another PR ? if yes i will just approve this. I guess easiest way to prevent it is to make the update an action handler

zashton (dismissed):

hasura/metadata/databases/default/tables/public_contributions.yaml
```diff
@@ -2,67 +2,52 @@ table:
...
-        - id
-        - epoch_id
-        - user_id
-        - description
-        - deleted_at
-        - created_at
-        - updated_at
+      - circle_id
+      - created_at
+      - datetime_created
```
crabsinger, line 32: @topocount `datetime_created` seems weird... how is this different than `created_at` and `updated_at`? `created_at` is when its created and any updates should bump the updated_at time...

topocount: @crabsinger created_at and updated_at are internal conventions in the db. datetime_created would be a user-modifiable column at some point.

hasura/migrations/default/1663180519908_contributions_decouple_epochs/up.sql
```diff
@@ -0,0 +1,17 @@
...
+  "description" text NOT NULL,
+  "datetime_created" timestamptz NOT NULL DEFAULT now(),
+  "created_at" timestamptz NOT NULL DEFAULT now(),
+  "updated_at" timestamptz NOT NULL DEFAULT now(),
+  "deleted_at" timestamptz,
+  PRIMARY KEY ("id"),
+  FOREIGN KEY ("circle_id") REFERENCES "public"."circles"("id") ON UPDATE restrict ON DELETE restrict,
+  FOREIGN KEY ("user_id") REFERENCES "public"."users"("id") ON UPDATE restrict ON DELETE restrict,
+  UNIQUE ("id")
+);
```
crabsinger, line 17: We still need a trigger here to update the updated_at right? 

Like this: https://github.com/coordinape/coordinape/blob/main/hasura/migrations/default/1662391842396_interaction_events/up.sql#L15

crabsinger (changes requested):
two questions, otherwise loosk good.

hasura/metadata/databases/default/tables/public_contributions.yaml
crabsinger: Oh, like the date of the contribution, but not the db date.

zashton: maybe change to `datetime_contributed` ? i think this column is used to identify which epoch the contribution belongs to

crabsinger: naming seems abit confusing to me, maybe just  `datetime` or `time` or something that doesnt have `created` on it ? thoughts? @topocount

levity: i vote for `contributed_at` to keep the convention that timestamps end in `_at`

hasura/migrations/default/1663180519908_contributions_decouple_epochs/up.sql
crabsinger:
```suggestion
);

CREATE TRIGGER "set_public_contributions_updated_at" BEFORE
UPDATE
  ON "public"."contributions"
  FOR EACH ROW
   EXECUTE PROCEDURE "public"."set_current_timestamp_updated_at"
   ();

COMMENT ON TRIGGER "set_public_contributions_updated_at" ON "public"."contributions"
IS 'trigger to set value of column "updated_at" to current timestamp on row update';
```

crabsinger (commented):
added a suggestion for the trigger

topocount:
thanks @crabsinger!

PR #1349, opened 2022-09-14
Chore: refactor icons

wingfeatherwave:
Big icon refactor. 

* no more deprecated icons
* use only icons that are in figma or from the [Feather library ](https://feathericons.com/)
* cleanup directory structure

Lots of files added / changed in this PR.  Please check the vercel link for visual testing.


src/components/FormModal.tsx
```diff
@@ -149,7 +146,6 @@ export const FormModal = ({
             }}
             disabled={submitDisabled}
           >
-            {icon || (!submitText && <DeprecatedSaveIcon />)}
```
CryptoGraffe, line 152: This intentional?

src/components/OrgLogoUpload.tsx
```diff
@@ -104,8 +105,8 @@ export const OrgLogoUpload = ({
                 background: '$white',
                 boxShadow: '$heavy',
                 borderRadius: '$3',
-                paddingLeft: '$2xl',
-                height: '$2xl',
+                paddingLeft: '64px',
```
CryptoGraffe, line 108: whats the story here

src/icons/__generated/CoinbaseColor.tsx
```diff
@@ -6,26 +6,26 @@
...
 import { SvgIcon, SvgIconProps } from '../../ui/SvgIcon/SvgIcon';
 const css = {
-  '& path': {
-    stroke: 'none',
-  },
+  // '& path': { stroke: 'none' }
+  // '& path': { stroke: 'none' }
 };
-export const SvgHamburger = (props: SvgIconProps) => (
+export const SvgCoinbaseColor = (props: SvgIconProps) => (
```
CryptoGraffe, line 12: These rename diffs are bizarre, the file is renamed to something totally diff??

src/ui/index.ts
```diff
@@ -23,6 +23,3 @@ export * from './TextArea/TextArea';
 export * from './TextField/TextField';
 export * from './Tooltip/Tooltip';
 export * from './Icon/Icon';
-
-// icons
-export * from './icons';
```
CryptoGraffe, line 28: whats the deal here? just curious

src/icons/__generated/CoinbaseColor.tsx
wingfeatherwave: that is really bizarre.  why would this happen?

src/ui/index.ts
wingfeatherwave: I was thinking that it was no longer needed.  Perhaps I'm wrong?

src/components/OrgLogoUpload.tsx
wingfeatherwave: This was for the org avatar upload.  I can do it with a calc, so I'll make that change.  

Screen Shot 2022-09-16 at 3 55 56 PM

src/components/FormModal.tsx
wingfeatherwave: Yes, we don't need a save icon in the primary submit button that says "save", IMO

wingfeatherwave: nvm... I can think of a case for keeping it.  I'll revert.

src/ui/index.ts
wingfeatherwave: .... ah, that was a relic of when there were some icons stored in /ui.  yea, no longer need to import those.  
Screen Shot 2022-09-16 at 4 20 50 PM

CryptoGraffe (approved):

PR #1351, opened 2022-09-15
update vault underlying to realtime share price

zashton:
## Motivation and Context

Underlying value does not sync with the pricePerShare real time hence its not a correct representative of the actual vault value.

## Description

Switched the calculation to use sharePrice of yTokens instead.

## Test and Deployment Plan

Clicking on the Max on the Gift input in distribution page should give a value of yTokens owned * pricePershare

image

## Reviewers

@levity


levity (dismissed):

topocount (changes requested):
I ran this locally and the createVault e2e test is broken after this change. It looks like it's probably just a question of changing the test to expect the correct value.

PR #1355, opened 2022-09-15
Add allocation updates to interaction_events

crabsinger:
Add allocation updates to interaction_events

### DIFF
--- a/api/hasura/actions/_handlers/updateAllocations.ts
+++ b/api/hasura/actions/_handlers/updateAllocations.ts
@@ -133,6 +133,22 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     const existingGift = pending_sent_gifts.find(
       g => g.recipient_id === gift.recipient_id
     );
+
+    ops['insertInteractionEvents'] = {
+      insert_interaction_events: [
+        {
+          objects: [
+            {
+              event_type: 'update_allocations',
+              profile_id: user?.id,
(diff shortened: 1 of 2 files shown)


api/hasura/actions/_handlers/updateAllocations.ts
```diff
@@ -133,6 +133,22 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     const existingGift = pending_sent_gifts.find(
       g => g.recipient_id === gift.recipient_id
     );
+
+    ops['insertInteractionEvents'] = {
```
crabsinger, line 137: more verbose to add it here, but its wrapped into one sql mutation vs if we used the insertInteractionEvents call...

topocount (dismissed):
This is great!

One thing to just verify: if any part of a tx fails, every single mutation would be rolled back. If there is no risk of erroneous rollback, this is good to go.

topocount:
> @topocount good to know, a mutation is always wrapped in a TRANSACTION from a sql standpoint?

@crabsinger when there is more than one mutation in a request, hasura wraps it in a database tx: https://hasura.io/docs/latest/mutations/postgres/multiple-mutations/#execution

levity:
why was this one closed?

crabsinger:
> why was this one closed?
@levity this set of changes got batched together into this PR: https://github.com/coordinape/coordinape/pull/1357

PR #1361, opened 2022-09-19
Smart rounding for allocation table and claims page

zashton:
## Motivation and Context

Update so asto  handle token numbers that need more decimals (like WBTC or YFI).

## Description

Use commify to truncate until the last significant number instead of hardcoded of 2 decimals point

## Test and Deployment Plan

Distribution page unit test should pass.

image

## Reviewers

@topocount @levity @crabsinger 

## Related Issue

https://www.notion.so/coordinape/78924b82ebe4445cbe20f586cdcb2795?v=bbf298c4fee34b5ea4eb18db638676bf&p=81b5681adae94c448482df112727e191&pm=s


crabsinger (approved):

levity (changes requested):
I'm not sure this is a good change as-is. With DAI and USDC, we don't need to show this many decimal places. What would be ideal is a "round to significant figures" function that is smart about showing extra decimal places only when the value would be rounded to 0 otherwise. 

But maybe a simpler rule would work, something like "round to 2 places if value is greater than 10, and to 4 places otherwise."

zashton:
> I'm not sure this is a good change as-is. With DAI and USDC, we don't need to show this many decimal places. What would be ideal is a "round to significant figures" function that is smart about showing extra decimal places only when the value would be rounded to 0 otherwise.
> 
> But maybe a simpler rule would work, something like "round to 2 places if value is greater than 10, and to 4 places otherwise."

updated

levity (approved):

PR #1364, opened 2022-09-19
Support local hasura console links

crabsinger:
Update the link from http://localhost:3000/developers

image

to properly go to a local hasura console for UI with current auth token.

Successful loading:
image

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -14,6 +14,8 @@ export const IN_PRODUCTION =
   process.env.NODE_ENV === 'production' &&
   process.env.REACT_APP_VERCEL_ENV !== 'preview';
 
+export const IN_DEVELOPMENT = process.env.NODE_ENV === 'development';
+
 // TODO: Have prod remove the trailing slash
 export const STORAGE_URL = getEnvValue(
   'REACT_APP_S3_BASE_URL',
@@ -32,6 +34,8 @@ export const REACT_APP_HASURA_URL = getEnvValue(
   'REACT_APP_HASURA_URL',
   'https://missing-hasura-url.edu'
 );
+
+export const LOCAL_HASURA_PORT = process.env.LOCAL_HASURA_PORT;
 
 export const HARDHAT_CHAIN_ID: number = +(process.env.HARDHAT_CHAIN_ID || 1337);
 export const HARDHAT_PORT: number = +(process.env.HARDHAT_PORT || 8545);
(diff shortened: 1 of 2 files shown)


src/config/env.ts
```diff
@@ -14,6 +14,8 @@ export const IN_PRODUCTION =
   process.env.NODE_ENV === 'production' &&
   process.env.REACT_APP_VERCEL_ENV !== 'preview';
 
+export const IN_DEVELOPMENT = process.env.NODE_ENV === 'development';
```
crabsinger, line 17: I could use !IN_PRODUCTION, but it feels safer to be explicit about "in development" than not prod.

src/utils/apiKeyHelper.ts
```diff
@@ -1,10 +1,14 @@
-import { REACT_APP_HASURA_URL } from '../config/env';
+import { IN_DEVELOPMENT, REACT_APP_HASURA_URL } from '../config/env';
 
 const EXAMPLE_QUERY =
   'query+MyCircles+%7B%0A++circles%28limit%3A+3%29+%7B%0A++++name%0A++++organization+%7B%0A++++++name%0A++++%7D%0A++++epochs%28limit%3A+3%2C+where%3A+%7Bended%3A+%7B_eq%3A+true%7D%7D%29+%7B%0A++++++id%0A++++++end_date%0A++++++start_date%0A++++%7D%0A++++users%28limit%3A+5%2C+order_by%3A+%7Bcreated_at%3A+asc%7D%29+%7B%0A++++++id%0A++++++name%0A++++++give_token_received%0A++++++give_token_remaining%0A++++++profile+%7B%0A++++++++twitter_username%0A++++++%7D%0A++++%7D%0A++%7D%0A%7D%0A';
 
 export const getConsoleUrl = (authToken: string, withExample?: boolean) => {
-  return `https://cloud.hasura.io/public/graphiql?endpoint=${REACT_APP_HASURA_URL}&header=Authorization:${encodeURIComponent(
+  const hasuraGui = IN_DEVELOPMENT
+    ? `http://localhost:8080/console`
```
crabsinger, line 8: This port is always running, so easier to use this than the http://localhost:9695/ console started by `yarn hasura console`. However, it seems if you use localhost:8080 to make database edits, it doesn't write out the changes to the local filesystem, where as if you use http://localhost:9695/ it does...

src/config/env.ts
```diff
@@ -33,6 +35,8 @@ export const REACT_APP_HASURA_URL = getEnvValue(
   'https://missing-hasura-url.edu'
 );
 
+export const LOCAL_HASURA_PORT = process.env.LOCAL_HASURA_PORT;
```
levity, line 38: i realize now that using LOCAL_ and CI_ prefixes is an antipattern... will have to go back and fix this at some point. just FYI, not a blocker

levity (approved):

Issue #545, opened 2022-02-15
Refactor ApeAvatar to only take user image props


teeolendo:
## Updated task

Remove ApeAvatar and NewApeAvatar; refactor to use src/ui/Avatar instead.

previously:

----

**Problem**
The ApeAvatar component uses the IUser object to as a prop. It's overkill to use the entire object.

**Solution**
ApeAvatar only requires either the user profile image or the user image path.

**Specification**
Ensure that the ApeAvatar only uses the single string prop that points to the location of the users avatar and refactor all instances of ApeAvatar to use the new prop.

**Stakeholders**
@levity

levity:
As can be seen in #533 above, there's currently a NewApeAvatar component that goes in this direction but is duplicative of ApeAvatar. 

As part of this task, once ApeAvatar is changed as described, remove NewApeAvatar and replace its uses with the new ApeAvatar.

sshmm:
@levity does that mean this is implemented?  I see mixed usage between ApeAvatar and NewApeAvatar.

levity:
Right, it's still only partly done.

amrro:
I just implemented a new `ui/Avatar` using stitches and updated it to only accept user's avatar path and user's name in #739.

@levity, I can utilize the new avatar component, but waiting for merging.

CryptoGraffe:
@amrro do you want to make this change now that you are merged, or should someone else pick it up?

levity:
Assigned to you @amrro. Please note that I've updated the task description.

yahiamarzouk:
I can change every component  using Avatar.tsx instead of ApeAvatar.tsx and NewApeAvatar.tsx, is this is the task required for this issue?
or there is more to it?

levity:
Yes, and then remove those other two from the codebase

PR #1365, opened 2022-09-20
switching to Avatar.tsx all over the CodeBase

yahiamarzouk:
## Motivation and Context

 this changes remove all ApeAvatar & NewApeAvatar from all the code base and replace it with Avatar

## Related Issue

 #545


src/components/ReceiveInfo.tsx
```diff
@@ -140,7 +140,7 @@ export const ReceiveInfo = () => {
                     maxWidth: '20rem',
                   }}
                 >
-                  <ApeAvatar user={tokenGift.sender} />
+                  <Avatar path={tokenGift.sender} />
```
levity, line 143: `sender` here isn't an avatar path

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -70,7 +60,7 @@ export const MyAvatarMenu = () => {
             }}
           >
             <Link href="#">
-              <ApeAvatar profile={myProfile} className={classes.avatarButton} />
+              <Avatar path={myProfile.avatar} size="medium" />
```
levity, line 63: When showing the avatar for a user profile, you should also set the `name` property, so that if the profile doesn't have an avatar image, we can show the initials of the name instead

src/ui/layouts.tsx
```diff
@@ -70,16 +69,10 @@ export const OrgLayout = ({
...
-          src="/imgs/avatar/placeholder.jpg"
-          style={{
-            width: 36,
-            height: 36,
-            borderRadius: '50%',
-            border: '1px solid rgba(94, 111, 116, 0.7)',
-            marginRight: '16px',
-          }}
+        <Avatar
+          name="organization"
```
levity, line 73: here, `name` should be the organization's name

levity (changes requested):
Thanks for the PR! See below. By the way, you can also remove ApeAvatar and NewApeAvatar from the code in this PR.

src/ui/layouts.tsx
yahiamarzouk: I can't reach the Organization's Name

src/components/ReceiveInfo.tsx
yahiamarzouk: I don't know the correct path for the token gift sender
can u tell me?

src/pages/DistributionsPage/AllocationsTable.tsx
```diff
@@ -5,9 +5,9 @@ import uniqBy from 'lodash/uniqBy';
 
 import { Paginator } from '../../components/Paginator';
 import { DISTRIBUTION_TYPE } from '../../config/constants';
-import { NewApeAvatar, makeTable } from 'components';
-import { Flex, Text, Panel, Button, Link } from 'ui';
-import { smartRounding, numberWithCommas, shortenAddress } from 'utils';
```
levity, line 10: `smartRounding` shouldn't have been removed here

levity (changes requested):

src/ui/layouts.tsx
levity: hmm, this layout should actually be removed entirely. will see if i can fix this in a different PR

src/components/ReceiveInfo.tsx
levity: it should be `tokenGift.sender.avatar` since `sender` is a User

src/pages/DistributionsPage/AllocationsTable.tsx
yahiamarzouk: Screen Shot 2022-09-23 at 6 18 16 AM

are we sure it should stay?

levity: your IDE might just be out of date -- try opening the `utils` file and see if the error goes away

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -70,7 +60,11 @@ export const MyAvatarMenu = () => {
             }}
           >
             <Link href="#">
-              <ApeAvatar profile={myProfile} className={classes.avatarButton} />
+              <Avatar
+                path={myProfile.avatar}
+                name={myProfile.discord_username}
```
levity, line 65: This one too :)

levity (changes requested):

yahiamarzouk:
@levity I think that's it

levity (changes requested):
Sorry for taking so long to get back to you on this. I've been traveling.

So, I finally checked these changes in a browser, and I noticed that many of the image sizes changed too much. You want to minimize the visual difference here, since we're just cleaning up code, not trying to change the user experience. I added a commit to fix some of these differences; please review that commit. (Note that when `size` is omitted, it is "large", because that's the default variant set in Avatar.tsx.)

The image on the profile page is the wrong size and in the wrong position; that still needs to be fixed.

e.g.: https://capture.dropbox.com/VME8mS22G5g7x5ar

levity (approved):

Issue #1363, opened 2022-09-19
CSV inconsistent with distribution page 


ashmbanks:
**Summary:**

1. **What is actually happening:**
When I have an epoch where a member has received GIVE, but also was deleted, the CSV export is showing different data than the distribution page. 
**Distribution Page**
<img width="1348" alt="Screen Shot 2022-09-19 at 9 34 24 AM" src="https://user-images.githubusercontent.com/113928419/191067766-28089576-723a-4746-b983-3249e0aa981f.png">
**CSV**
<img width="982" alt="Screen Shot 2022-09-19 at 9 32 47 AM" src="https://user-images.githubusercontent.com/113928419/191067497-c7508e13-e451-4c49-9fbb-e41fcd805ef7.png">

**(NOTE that these values are ephemeral for the time being,; i.e. there is no way to save the amount that I put into the gift circle if I am not doing a vault distribution)**

2. **What should be happening:**
The CSV should match the distribution page.

3. **Steps to recreate:**
1. Have circle with a completed epoch with multiple members who have received GIVE.
2. Delete one of the members.
3. Go to Distribution page and enter some amount into the Gift Circle section. (Do not use a vault)
4. Export the CSV.
5. See that the amounts in the CSV and Distribution page differ and that the deleted member is still in the CSV.

PR #1367, opened 2022-09-20
fix: Remove deleted users from CSV

karelianpie:
## Motivation and Context

When there is an epoch where a member has received GIVE, but also was deleted, the CSV export is showing different data than the distribution page because the deleted users are still being taken into account.

## Description

Remove deleted users from CSV so their received GIVE will not be taken into consideration when calculating the percentages of other users. This is what happens in the distribution page's table.

## Test and Deployment Plan

Ran locally and observed the CSV without the deleted user as well as the correct adjusted percentages to account for the absence of the deleted users, refer to the screenshots below

### CSV (before)

csv-before

### CSV (after)

csv-after

### Table (before)

table-before

### Table (after)

table-after

## Reviewers

@ashmbanks

## Related Issue

Fixes https://github.com/coordinape/coordinape/issues/1363

### DIFF
--- a/api-lib/gql/queries.ts
+++ b/api-lib/gql/queries.ts
@@ -544,6 +544,10 @@ export async function getEpoch(
           token_gifts: [
             {},
             {
+              recipient: {
+                id: true,
+                deleted_at: true,
+              },
               tokens: true,
             },
           ],
--- a/api-lib/gql/queries.ts
+++ b/api-lib/gql/queries.ts
@@ -542,12 +542,8 @@ export async function getEpoch(
             },
           },
           token_gifts: [
-            {},
+            { where: { recipient: { deleted_at: { _is_null: true } } } },
             {
-              recipient: {
-                id: true,
-                deleted_at: true,
-              },
               tokens: true,
             },
           ],
--- a/api/hasura/actions/_handlers/allocationCsv.ts
+++ b/api/hasura/actions/_handlers/allocationCsv.ts
@@ -35,19 +35,9 @@ async function handler(req: VercelRequest, res: VercelResponse) {
   const grant = payload.grant ?? epochObj.grant;
 
   const totalTokensSent = epochObj.token_gifts.length
-    ? (
-        epochObj.token_gifts?.filter(
-          ({ recipient }: Gift) => !recipient.deleted_at
-        ) || []
-      )
-        .map(g => g.tokens)
-        .reduce((total, tokens) => tokens + total)
+    ? epochObj.token_gifts.reduce((total, { tokens }) => total + tokens, 0)
     : 0;
-  const circle = await getCircleDetails(
-    circle_id,
-    epochObj.id,
-    epochObj.end_date
-  );
+  const circle = await getCircleDetails(circle_id, epochObj.id);
   assert(circle, 'No Circle Found');
   const fixedPaymentsEnabled =
     isFeatureEnabled('fixed_payments') && !!circle.fixed_payment_token_type;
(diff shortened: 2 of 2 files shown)


api-lib/gql/queries.ts
```diff
@@ -544,6 +544,10 @@ export async function getEpoch(
           token_gifts: [
             {},
             {
+              recipient: {
+                id: true,
+                deleted_at: true,
+              },
```
crabsinger, line 550: is this needed, or can we just filter the data using the `where` on the backend graphql side and not do any FE side checking of the `deleted_at` field?

karelianpie: Good question, to be fair I'm not sure. I wonder if the requirements have changed for the CSV, as we currently have the following
```typescript
{
  where: {
    _or: [
      { deleted_at: { _is_null: true } },
      { deleted_at: { _gt: epochEndDate } },
    ],
  },
}
```
So I'm assuming at some point we wanted the deleted users only if they were deleted after the epoch ended? In either case, I do agree with you that this filtering would be better done in the backend. Perhaps @zashton could shed some light here

zashton: if the user is deleted after the epoch ends , the GIVES he received are still valid and not refunded so it should be represented somehow in the frontend e.g in the map etc so the accounting is correct.

zashton: After reading through the issue i'm not sure this is a bug though.
Deleting a user only after the epoch ends can also mean the admin wants the user to be included in the distribution.
My understanding is that deleting a user should not affect any of the epochs that he has participated in the past, otherwise the allocation math would not add up.

karelianpie: @zashton could the bug actually be in the distribution table rather than the csv? In the FE the GIVE that the deleted user got are not taken into account when calculating the percentages of the other users

crabsinger: > Deleting a user only after the epoch ends can also mean the admin wants the user to be included in the distribution.
> My understanding is that deleting a user should not affect any of the epochs that he has participated in the past, otherwise the allocation math would not add up.

Hey @karelianpie - sorry for the confusion on this. I think we want the CSV export to match and have the exact same functionality as the Distribution page. Can you jsut wait a day or two for us to confirm and clarify the behavior this should have.

crabsinger: But I think what you've done IS what we want, i.e. that the CSV excludes the deleted users GIVE and recalculates just like the distribution page.

crabsinger: @karelianpie @zashton ok, so your original implementation has the correct and desired functionality for the CSV, which is to exclude deleted users and NOT show them as a line item in the CSV and recalculate the total give excluding their amount received. This follows the functionality of the Distributions page. So the only changes to this PR should be cleaning up the code, but the screenshots of the functionality are correct as done.

karelianpie: Thank you @crabsinger for the clarification! I've refactored it so that it filters the deleted users directly in the `where` clause rather than doing it in the FE, please have a look https://github.com/coordinape/coordinape/pull/1367/commits/b3139fe985cd742069f421ebdadbeb41eaf654b4

api/hasura/actions/_handlers/allocationCsv.ts
```diff
@@ -35,15 +35,9 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
   const totalTokensSent = epochObj.token_gifts.length
-    ? epochObj.token_gifts
-        .map(g => g.tokens)
-        .reduce((total, tokens) => tokens + total)
+    ? epochObj.token_gifts.reduce((total, { tokens }) => total + tokens, 0)
     : 0;
-  const circle = await getCircleDetails(
-    circle_id,
-    epochObj.id,
-    epochObj.end_date
```
karelianpie, line 45: End date no longer needed as it's irrelevant when the user was deleted, we don't want to account their GIVE if they are deleted, even if after the epoch has ended

crabsinger:
@karelianpie nice work simplifying this. Can you rebase this off main, or merge main into it, and then I will finalzie review and merge! 🙏

karelianpie:
> @karelianpie nice work simplifying this. Can you rebase this off main, or merge main into it, and then I will finalzie review and merge! 🙏

Thanks @crabsinger, rebased

crabsinger (approved):

PR #1372, opened 2022-09-21
Standardize Linter runs across CI and husky

topocount:
It's "pretty" apparent some of us aren't running lint-staged on commit,
which is fine. However, lint-stage prettier fixing was divergent from
our CI-enforced `yarn prettier:check`. This has caused some spam for
git-blames which isn't the biggest deal but has really irked me for
a while.

This change makes prettier executions consistent across lint-staged and
CI.

I've also configured our prettier CLI commands to run on all
our config files, which is best practice.

I'm also open to getting rid of husky, since I think devs should be responsible for configuring their IDEs to hook into a project's config correctly.

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,4 +1,4 @@
 {
-    "singleQuote": true,
-    "arrowParens": "avoid"
+  "singleQuote": true,
+  "arrowParens": "avoid"
 }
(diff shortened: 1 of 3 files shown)


crabsinger (approved):
nice looks good!

levity (approved):

zashton (approved):

PR #1373, opened 2022-09-21
Create Contribution Seed script logic

topocount:
Contributions are a useful primitive we are building around and this
saves us some lift to begin hacking on them.


scripts/db_clean.ts
```diff
@@ -8,6 +8,7 @@ type MutationName = keyof ValueTypes['mutation_root'];
     'delete_vault_transactions',
     'delete_pending_vault_transactions',
     'delete_claims',
+    'delete_contributions',
```
zashton, line 11: this is already declared once above

topocount: @zashton I ran into a fk error on deletion, so I guess we need to move it up

crabsinger (dismissed):
maybe the default count should be more than 3 tho?

PR #1374, opened 2022-09-23
Radix modal, now with drawer power

wingfeatherwave:
Now you can turn a `<Modal>` look like a right-sliding drawer with `<Modal drawer>`

(FYI, the create vault modal is still a regular ol modal, I was just using that as an example.)

Screen Shot 2022-09-22 at 11 43 07 AM


src/pages/VaultsPage/VaultsPage.tsx
```diff
@@ -186,16 +186,18 @@ const VaultsPage = () => {
           )}
         </>
       )}
-      {modal === 'create' && currentOrg && (
+      {currentOrg && (
```
levity, line 189: do the animations still work if you re-add `!!modal` to this? i think there's a slight performance advantage to not rendering the modal component at all until it's opened

levity: ah, they don't, the close animation gets lost...

levity (approved):

levity:
looks great! good call adding it as a prop to Modal

src/pages/VaultsPage/VaultsPage.tsx
wingfeatherwave: Radix doesn't add the modal to the dom until it's triggered, so there's no render hit before mount... but the radix root primitive I guess is still in the react tree somewhere.  But to answer your question the open animation will work, but not the close.

PR #1377, opened 2022-09-23
Enable duplicate vault types

zashton:
## Motivation and Context

It allows circles to have multiple vaults of the same token for distribution.

## Description

Remove the checks that restrict creation of duplicate token vaults, add more identifiers e.g vault addresses on top of their token names to differentiate between same type of vaults.

## Test and Deployment Plan

Current unit tests should pass, you should be able to create vaults of the same token type

image
image
image
image


src/pages/VaultsPage/VaultRow.tsx
```diff
@@ -66,11 +68,16 @@ export function VaultRow({ vault, css = {} }: { vault: Vault; css?: CSS }) {
         onUpdateBalance={updateBalance}
       />
       <Box
-        css={{ display: 'flex', alignItems: 'center', gap: '$md', mb: '$md' }}
+        css={{ display: 'flex', alignItems: 'center', gap: '$md', mb: '$xs' }}
       >
         <Text h3 css={{ flexGrow: 1 }}>
           {vault.symbol || '...'} CoVault
+          <VaultExternalLink
```
levity, line 75: in order for this to have the correct spacing next to the rest of the text, it should be set to `display: inline` (or inline-block) rather than using a fixed left margin

src/lib/vaults/tokens.ts
```diff
@@ -95,3 +96,13 @@ export const getDisplayTokenString = (vault: {
   simple_token_address: string;
 }): string =>
   hasSimpleToken(vault) ? vault.symbol : addYearnPrefix(vault.symbol);
+
+export const getVaultSymbolAddressString = (vault: {
+  symbol: string;
+  vault_address: string;
+}): string => `${vault.symbol} ${shortenAddress(vault.vault_address, false)}`;
+
+export const removeAddressSubfix = (symbol: string, vaultAddress: string) => {
```
levity, line 105: this should be "Suffix"

src/pages/DistributionsPage/DistributionsPage.test.tsx
```diff
@@ -147,7 +147,9 @@ test('render with no allocations', async () => {
 
   expect(screen.getByText('Gift Circle')).toBeInTheDocument();
   expect(
-    screen.getAllByText('Yearn USDC')[0].closest('button[role="combobox"]')
+    screen
+      .getAllByText('Yearn USDC 0xA0b8')[0]
```
levity, line 151: just in case this changes, you can also match a regex instead of a string: https://testing-library.com/docs/queries/about#textmatch

levity (changes requested):

src/pages/VaultsPage/VaultRow.tsx
zashton: tried and seem to only work when i used `&nbsp;`

src/lib/vaults/tokens.ts
```diff
@@ -95,3 +96,13 @@ export const getDisplayTokenString = (vault: {
...
 }): string =>
   hasSimpleToken(vault) ? vault.symbol : addYearnPrefix(vault.symbol);
+
+export const getVaultSymbolAddressString = (vault: {
+  symbol: string;
+  vault_address: string;
+}): string => `${vault.symbol} ${shortenAddress(vault.vault_address, false)}`;
+
+export const removeAddressSuffix = (symbol: string, vaultAddress: string) => {
+  const subfix = shortenAddress(vaultAddress, false);
```
levity, line 106: Change the spelling here too, please

levity (changes requested):
Just one more thing :)

PR #1388, opened 2022-09-24
adding contributions fk object relationship to users

CryptoGraffe:
## Motivation and Context

for new allocation UI I need to be able to aggregate the count of contributions for each user

## Description

modifies hasura to take advantage of the existing fk relationship


CryptoGraffe:
* main
  * **PR #1388** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1388" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1388?utm_source=stack-comment).

hasura/metadata/cron_triggers.yaml
```diff
@@ -1,44 +1 @@
-- name: check-nominee
```
zashton, line 1: don't think this should be removed

zashton (changes requested):

hasura/metadata/cron_triggers.yaml
CryptoGraffe: woo good catch. i had my local state in a funk from seeding issues when i did a metadata export

CryptoGraffe: fixed this

zashton (approved):
lgtm!

PR #1394, opened 2022-09-26
Update deposit of secondary wallet button

zashton:
To make the link more prominent

### Before:
image

### After:
image


levity:
I don’t think we actually need to make this that big. there are things on
that post-mvp list that we should talk about before you build; not
everything there is something we actually all agree we should do. But
thanks for taking the initiative

PR #1405, opened 2022-09-28
Use cloud hasura graphiql for all dev portals

### LINKED ISSUE #1391: Developer console link isn't setting authorization header
1. **What is actually happening:** In dev, the link to API explorer from /developers doesn't set the Authorization header

2. **What should be happening:** It should set the Authorization header so that you can immediately use the console as that user

3. **Steps to recreate:** visit /developers in a dev server and click the Open API Explorer button

Maybe this broke due to the recent upgrade to hasura-cli?

crabsinger:
Resolves #1391 

- bump hasura version to 2.11.2 (matching prod's current version)
- fix DevPortal link

@levity 

Seems this fix is working for me in localhost. Just use prod console pointed at localhost.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
   # Hasura
   graphql-engine: &graphql_engine
     profiles: ['dev']
-    image: hasura/graphql-engine:v2.1.1.cli-migrations-v3
+    image: hasura/graphql-engine:v2.11.2.cli-migrations-v3
     volumes:
       - ./hasura/migrations:/hasura-migrations
       - ./hasura/metadata:/hasura-metadata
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
   # Hasura
   graphql-engine: &graphql_engine
     profiles: ['dev']
-    image: hasura/graphql-engine:v2.1.1.cli-migrations-v3
+    image: hasura/graphql-engine:v2.11.2.cli-migrations-v3
```
crabsinger, line 18: bumping to match prods version

levity (approved):

PR #1413, opened 2022-09-28
Setup Gitpod for instant ready to use dev environments

crabsinger:
## Motivation and Context

For new devs, having an easy way to load our dev environment and contribute is extremely helpful.
For existing devs, having an easy way to test a PR or push heavy computation to a remote server is very useful.
For everyone, having a fully automated "Dev environment as code" is great.

## Description

Setup a .gitpod.yml and assocaited changes to make Gitpod launch our dev environment.

## Test and Deployment Plan

Go to Gitpod, create an account, launch this branch in Gitpod: or just go to 
https://gitpod.io/#https://github.com/coordinape/coordinape/tree/feature/setup-gitpod

If you want to use an existing .env, just open the Gitpod, copy that into .env, and then run
`gp env DOTENV="$(base64 .env | tr -d '\n')"` which will save it to your Gitpod account and restore from that file on all new workspace starts.

## Remaining Issues
If you have $ in strings in your .env, docker compose will fail.

If you get an error like:
```
$ docker compose
Invalid template: "coordinape.com$$,localhost:3000,.*.gitpod.io$"
```
Change your .env value of `SIWE_ALLOWED_DOMAINS="coordinape\.com\$,localhost:3000,.*\.gitpod\.io\$"` to something like this where all $ are escaped with a $.

Due to CORS and different domains, this doesn't properly work if you are visiting the local site using gitpods hosted domains such as https://3000-coordinape-coordinape-9zxapbh7lqn.ws-us67.gitpod.io. This could likely be set up in the future.

For now, if you open the Gitpod workspace in VS Code for Desktop, it should forward local ports properly and then just visit localhost:3000.

image

image

Screen Shot 2022-09-28 at 12 32 27 PM


topocount (commented):
My primary anxieties around this are not usability, but maintainability. We have a lot of tooling that requires attention right now and im hesitatnt to merge another tool that some devs will expect to be maintained. It might be more worthwhile to prioritize the maintenance of existing tooling before introducing another thing that requires support.

crabsinger:
> My primary anxieties around this are not usability, but maintainability. We have a lot of tooling that requires attention right now and im hesitatnt to merge another tool that some devs will expect to be maintained. It might be more worthwhile to prioritize the maintenance of existing tooling before introducing another thing that requires support.

I think that's a valid concern. What existing tooling would you propose should be prioritized for more maintainance over introducing Gitpod?


That said, right now getting a Coordinape dev environment set up takes people a few hours usually I think, and often they run into weird issues and errors. Additionally, there is only the text in the README, and no code that tries to actually set up the environment and automate it. I think having more code that sets up a dev env will move us towards easier ability to make sure changes work on new dev envs and keep our stack up to date. Additionally, there's lots of things that only some devs have setup or know how to setup like connecting and testing vaults on a local EVM network, or even getting CI to run locally (mine currently errors and I haven't put the time into debugging it). Ideally, these things would be able to run with one command on every dev env, and doing that with something like Gitpod is more more feasible and easy than anything we have right now

Right now, it can easily happen that a new dev env will encounter an issue that an existing dev env would not encounter.

I propose we try it out, see if it is useful and helps new folks and we can remove support for this if we find it is requiring attention and work to maintain.

levity:
Sounds like you’re volunteering to champion this :)

crabsinger:
> Sounds like you’re volunteering to champion this :)

Yes, I am.

levity (approved):

Issue #1401, opened 2022-09-27
manual entry into the GiveAllocator can't erase the 0 because gifts is non optional


CryptoGraffe:
The new GiveAllocator allows for manual entry of a number, but you can't actually erase the 0 to start from scratch. I think we need to make the tokens field optional to fix this bug I haven't investigated.

The field should probably also use some kind of zod validation.

karelianpie:
I'll look into this once https://github.com/coordinape/coordinape/pull/1384 is merged

crabsinger:
@karelianpie #1384 is merged.

karelianpie:
> @karelianpie #1384 is merged.

Will work start working on this then, thanks for the heads up 👍

PR #1423, opened 2022-09-29
fix: Allow for empty string in the `GiveAllocator` input

karelianpie:
## Motivation and Context

The new GiveAllocator allows for manual entry of a number, but you can't actually erase the 0 to start from scratch. 

## Description

* Made the `tokens` field optional;
* Added a logic to handle the empty string and zero on the input field.

## Test and Deployment Plan

Ran locally and tried it out, refer to screenshot below

Screen Shot 2022-09-29 at 10 27 46 am

## Reviewers

@CryptoGraffe

## Related Issue

Fixes https://github.com/coordinape/coordinape/issues/1401

### DIFF
--- a/src/pages/GivePage/GiveAllocator.tsx
+++ b/src/pages/GivePage/GiveAllocator.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import { ApeInfoTooltip } from '../../components';
 import { MinusCircle, PlusCircle } from '../../icons/__generated';
@@ -26,6 +26,8 @@ export const GiveAllocator = ({
   maxedOut,
   optedOut,
 }: GiveAllocatorProps) => {
+  const [isInputEmpty, setIsInputEmpty] = useState(false);
+
   //incGift increments the gift by 1
   const incGift = (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
     e.stopPropagation();
@@ -36,7 +38,7 @@ export const GiveAllocator = ({
   //decGift decrements the gift by 1
   const decGift = (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
     e.stopPropagation();
-    if (gift.tokens > 0) {
+    if ((gift.tokens ?? 0) > 0) {
       adjustGift(gift.recipient_id, -1);
     }
   };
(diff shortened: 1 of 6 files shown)


src/pages/GivePage/GiveAllocator.tsx
```diff
@@ -104,14 +103,15 @@ export const GiveAllocator = ({
...
+          {((gift.tokens ?? 0) < 1 || disabled) && <ClickTrapperIcon />}
         </Flex>
         <TextField
           data-testid="tokenCount"
-          value={gift.tokens}
+          value={gift.tokens ?? ''}
           onChange={evt => setGiftTokens(evt.target.value)}
           maxLength={5}
           disabled={disabled}
+          onBlur={() => setGiftTokens(gift.tokens ? String(gift.tokens) : '0')}
```
karelianpie, line 114: Added this so that when the user loses focus on an empty input field it will automatically populate with a `0` -- otherwise, the input field would stay empty. It's just for a better UX, the maths for GIVEs would work correctly.

src/pages/GivePage/index.tsx
```diff
@@ -184,19 +185,19 @@ const GivePage = () => {
   };
 
   // adjustGift adjusts a gift by an amount if it is allowed wrt the max give, then schedules saving
-  const adjustGift = (recipientId: number, amount: number) => {
+  const adjustGift = (recipientId: number, amount: number | null) => {
```
crabsinger, line 188: Could we continue to require this to be non-null and just make sure to not fire the adjustGift event if the number is empty? Would be nice to not have to pass forward the optional logic all across this...

karelianpie: We update the `gift.tokens` in the `adjustGift()` and I could not find a clean way to tell that function that the input is empty. We want to allow `gift.tokens` to be `null` or `undefined` so that we can set the `TextField` value to empty string in
```typescript
<TextField
  data-testid="tokenCount"
  value={gift.tokens ?? ''}
  // ...
/>
```
I've initially added a new param `isEmpty` to that function signature, but that felt a bit overkill, hence why I've ended up with `amount` being `number | null` precisely  -- as you've correctly mentioned -- to tell the `adjustGift` that the number is empty.

Feel free to play give it a go, a few things to test:
1. Delete all the numbers;
2. Type `0` into an empty field;
3. Type any number as long as they are less than the total gives left in the empty field.

crabsinger (approved):
Nice work on this - def a UX improvement.

PR #1429, opened 2022-09-29
Update hasura image

crabsinger:
Updating the hasura image again. This broke types... so don't merge this until we figure this out...

Why does this break types...?


crabsinger:
@topocount any idea why this broke our types?

PR #1436, opened 2022-09-30
Add support for WETH vaults

zashton:
## Motivation and Context

Adding support for WETH token vaults due to heavy request from users

## Test and Deployment Plan

Unit tests should pass, you should be able to create an WETH vault and make a distribution. 
To mint WETH tokens do `yarn hardhat mint --token WETH --amount 10 --network ci`

image

## Reviewers

@levity @topocount


src/pages/VaultsPage/CreateForm.tsx
```diff
@@ -136,14 +136,15 @@ export const CreateForm = ({
       onChange({ target: { value: undefined } });
     } else if (vaultType == 'simple') {
       setDisplayCustomToken(false);
-      if (symbol == 'USDC') {
+      if (
```
levity, line 139: Hmm, can we find a way to avoid repeating the list of tokens here?

zashton: hows about this ?

Issue #1409, opened 2022-09-28
Visual regression on the epoch dropdown on the map


karelianpie:
**Summary:**

1. **What is actually happening:**

![image](https://user-images.githubusercontent.com/78794805/192841656-2f8c5ede-7d71-4d86-b8dd-a6806e4deb62.png)

3. **What should be happening:**

It should have a border & be full-width.

5. **Steps to recreate:**

Go to the maps page

karelianpie:
Root cause was this change https://github.com/coordinape/coordinape/pull/1268/files#diff-04f735dc4a31f83d2f95252292d564b622a5b6b5039dbd2cef7ca5b2c297b6efR25

wingfeatherwave:
Resolved with https://github.com/coordinape/coordinape/pull/1438

PR #1438, opened 2022-09-30
Map `<Select /> ` background styles

wingfeatherwave:
Screen Shot 2022-09-30 at 9 56 19 AM

`<Panel invertForm>` as a parent gives correct form stylings on $surface bg

References https://github.com/coordinape/coordinape/issues/1409


crabsinger (approved):
👍

levity:
In this screenshot, the select is wider than the search bar

PR #1439, opened 2022-09-30
Create Organizations page

### LINKED ISSUE #1463: Create Organizations Page

crabsinger:
Create Organizations page

resolves #1463 
 
* Create new page for Organizaitons showing all circles under org
* Add link to Organizations in main header
* Add link to Organizations from Org Name in Overview page

Screen Shot 2022-10-04 at 5 03 40 PM

Screen Shot 2022-09-30 at 10 28 06 AM


crabsinger:
resolves [#1463](https://github.com/coordinape/coordinape/issues/1463)

src/pages/OrganizationPage/getOrgData.ts
```diff
@@ -0,0 +1,46 @@
+import { order_by } from 'lib/gql/__generated__/zeus';
+import { client } from 'lib/gql/client';
+
+export const getOrgData = (orgId: number, address: string) =>
+  client.query(
+    {
+      organizations_by_pk: [
```
wingfeatherwave, line 7: How do we unify / simplify this with the `src/pages/CirclesPage/getOrgData.ts` version which is basically the same @CryptoGraffe ??

CryptoGraffe: just unify em and move into  `src/lib/gql/queries.ts`

CryptoGraffe: maybe more appropriate to make a useOrg hook that wraps the react query part too?

levity:
Please don’t do that 😅 the original purpose of `lib` is for standalone
code that could eventually be moved into their own node modules. Queries
are very coupled to front-end needs and should stay co-located with the
components that use them. You could just leave the original one where it is
and import it from there for use here.

CryptoGraffe:
Huh ok - how about a useOrg hook that does the react query wrapping as
well? Just seems weird to import from a diff page rather than a shared root
of some sort

CryptoGraffe:
I guess the downside is bloat in the common layer (eg too many columns etc)
vs the specific needs of each page. If we want to go that route I’d argue
we shouldn’t import cross page either

crabsinger:
@levity @CryptoGraffe is there a pattern for having shared FE components that are not in lib as standalone things? That seems like a nice pattern for both.

But still, the issue with this getOrgData query is that the queries are **slightly** different with one getting all orgs and the other getting ONE org by _pk... so I could do some weird graphql interpolation to support both in one function, but not sure what pattern is cleanest for this, and that's what I'm also wondering.

Circles:

```
  client.query(
    {
      organizations: [
        {},
        {
          id: true,
          name: true,
...
```


Orgs:
```
  client.query(
    {
      organizations_by_pk: [
        { id: orgId },
        {
          id: true,
          name: true,
```

src/pages/OrganizationPage/getOrgData.ts
crabsinger: seems we dont...

wingfeatherwave (approved):

levity:
I understand why it seems weird, but I’ve also been burned by premature
frontend code reuse attempts so many times, I kind of like how it signals
that the reuse is ad-hoc & coincidental.

Anyway, if you still want, we could make a shared root like
src/organizations. Or e.g. src/features/organizations if you want some
top-level grouping like src/pages & src/components.

CryptoGraffe:
I like the idea of feature based grouping but upon review it’s not
necessary here - the used are sufficiently different . I think what would
be nice is to try to reduce our useQuery boilerplate somewhat

levity:
This is a good idea if you’re sure the queries are going to stay exactly
the same, because then you can use the Sam e query key.

src/pages/OrganizationPage/OrganizationPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+  const goToCircle = (id: number, path: string) => {
+    scrollToTop();
+    navigate(path);
+  };
+
+  if (query.isLoading || query.isIdle || query.isRefetching)
+    return <LoadingModal visible note="OrganizationPage" />;
+
+  if (!org) {
+    navigate(paths.circles);
```
levity, line 43: just noticed this -- you can't call this during rendering, it needs to be in e.g. `useEffect`

Issue #1441, opened 2022-10-03
Story for the `Panel` component


karelianpie:
**Problem**

The `Panel` component is lacking documentation when it comes to its usage, for example https://discord.com/channels/845400066532704256/898299792575299654/1025452274824253591

**Solution**

Add a story so developers know how to use the `Panel` component with other components such as the `Select`

**Specification**

A story for the `Panel` component

**Stakeholders**

@wingfeatherwave

Issue #1255, opened 2022-08-09
SPIKE: Zodiac plugin for Gnosis Safe - Admin Vaults


reeserj:
**Problem**
What problem is this feature addressing?
Currently a circle admin can drain a vault without triggering another signature from the multisig

**Solution**
What's your proposed solution to the problem?
We need to have distributions be behind a multi-sig approval

**Specification**
What precisely needs to get built? (include any requirements or design files)
TBD

**Stakeholders**
Who needs to provide input? Who might be impacted? Who needs to approve?
@spydermonkee @levity @topocount

karelianpie:
@reeserj feel free to assign this to me, I'll start discovering it

levity:
closing this as a duplicate; we'll use #1253 instead to coordinate around Gnosis Safe matters

PR #1442, opened 2022-10-03
story: Panel with Select

karelianpie:
## Motivation and Context

The `Panel` component is lacking documentation when it comes to its usage, for example https://discord.com/channels/845400066532704256/898299792575299654/1025452274824253591

## Description

Add a story for the `Panel` component

## Test and Deployment Plan

Ran storybook and saw the `Panel` story working correctly with a `Select` and `TextArea` components

Screen Shot 2022-10-03 at 10 46 40 am

## Reviewers

@wingfeatherwave 

## Related Issue

Closes https://github.com/coordinape/coordinape/issues/1441

### DIFF
--- a/src/ui/Panel/Panel.stories.tsx
+++ b/src/ui/Panel/Panel.stories.tsx
@@ -1,0 +1,56 @@
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import { withDesign } from 'storybook-addon-designs';
+
+import { Select } from '../Select/Select';
+import { TextArea } from '../TextArea/TextArea';
+
+import { Panel as PanelComponent } from './Panel';
+
+export default {
+  title: 'Design System/Components/Panel',
+  component: PanelComponent,
+  decorators: [withDesign, Story => <Story />],
+} as ComponentMeta<typeof PanelComponent>;
+
+const options = [
+  { value: 'apple', label: 'Apple' },
+  { value: 'banana', label: 'Banana' },
+  { value: 'orange', label: 'Orange' },
+];
+
+const Template: ComponentStory<typeof PanelComponent> = args => (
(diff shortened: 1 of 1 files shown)


crabsinger:
Thanks for adding a story for this. We've thought about (but not yet setup I don't think) a deployment of our Storybook so people can play around with our components on a hosted version as well. And add even more components to it. So thanks for adding this one!

karelianpie:
> Thanks for adding a story for this. We've thought about (but not yet setup I don't think) a deployment of our Storybook so people can play around with our components on a hosted version as well. And add even more components to it. So thanks for adding this one!

I would be happy to add stories for other components, would that be considered high(ish) importance? I just added this one since Wing mentioned that this `Panel` should be better documented.

Let me know if I should start writing more stories or if there's something more important, I am looking at the Zodiac module for the Gnosis Safe integration at the moment (https://github.com/coordinape/coordinape/issues/1255) but I'm lacking a lot of context, so going through documentation and so on at the minute

wingfeatherwave (approved):

crabsinger:
@karelianpie I think it would be great to find Storybook deployment solution as well, on gh-pages or someplace else so we can easily view Storybook not only in local, then yes, adding more examples to it would be pretty valuable, if you want to work on this more.

karelianpie:
> @karelianpie I think it would be great to find Storybook deployment solution as well, on gh-pages or someplace else so we can easily view Storybook not only in local, then yes, adding more examples to it would be pretty valuable, if you want to work on this more.

Sounds good, created a PR https://github.com/coordinape/coordinape/pull/1454 -- we only need now to create an account on [Chromatic](https://www.chromatic.com/) and set the env vars for the CI to publish it automatically

Once that's merged I can start adding a few more stories

Issue #1421, opened 2022-09-29
Map should not be cut off at top of page


levity:
![Coordinape](https://user-images.githubusercontent.com/193667/192940640-b4b7a9d1-2338-462d-841b-8e57dbc52938.png)

levity:
@karelianpie maybe while you're working on the map page, you could look into this one as well?

karelianpie:
> @karelianpie maybe while you're working on the map page, you could look into this one as well?

Yup, happy to, just assigned this to myself

PR #1443, opened 2022-10-03
fix: Map cut off

karelianpie:
## Motivation and Context

Map should not be cut off at top of page, refer to issue https://github.com/coordinape/coordinape/issues/1421

## Description

Makes the position of the text absolute pinned to the top of the page

## Test and Deployment Plan

Ran locally and saw the map not being cut off, refer to screenshots below

### Drawer opened

Screen Shot 2022-10-03 at 3 12 40 pm

### Drawer closed

Screen Shot 2022-10-03 at 3 12 52 pm

## Reviewers

@levity 

## Related Issue

Fixes https://github.com/coordinape/coordinape/issues/1421

### DIFF
--- a/src/components/Drawer.tsx
+++ b/src/components/Drawer.tsx
@@ -19,6 +19,11 @@ export const Drawer = ({ open, setOpen, children }: IProps) => {
           fontSize: '$h1',
           fontWeights: '$semibold',
           ml: '$lg',
+          position: 'absolute',
+          top: 0,
+          bottom: 0,
+          zIndex: 2,
+          flexDirection: 'column',
         }}
       >
         Circle Map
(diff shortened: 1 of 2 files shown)


levity:
Thanks, it looks good!

I noticed in testing that there's a visual artifact that popped up at the bottom of the profile card list: a horizontal scrollbar. It also scrolls horizontally when it shouldn't--this wasn't caused by your changes, since it happens in production as well, but maybe you can fix it since you're already working in this area? See image below.

Coordinape

karelianpie:
> Thanks, it looks good!
> 
> I noticed in testing that there's a visual artifact that popped up at the bottom of the profile card list: a horizontal scrollbar. It also scrolls horizontally when it shouldn't--this wasn't caused by your changes, since it happens in production as well, but maybe you can fix it since you're already working in this area? See image below.
> 
> Coordinape

So the issue was that one of the users had a really long name, in order to address that and prevent horizontal scrolling we've added ellipsis to the end when the name overflows https://github.com/coordinape/coordinape/pull/1443/commits/c12f440adf1f53c63cf8696bee8c28bf59fe864c

PR #1447, opened 2022-10-03
Wrapping of ETH to WETH during deposit and unwrapping to ETH when claming

zashton:
## Motivation and Context

To allow users to deposit to our vaults using ETH into the WETH vault and claiming back into ETH.

## Description

Addition of a checkbox in the deposit modal to allow user to toggle between ETH/WETH and if ETH is picked additional wrapping transaction will be included.

For claims if distribution is from a WETH vault, modal will popup for users to choose whether to claim in WETH/ETH tokens.

## Test and Deployment Plan

Current unit tests should pass, the deposit WETH/ETH into vaults and claiming of it with or without wrapping should work.

image
image


levity (approved):
@zashton please note my changes

levity (approved):

levity (dismissed):

levity:
@zashton the Cypress test is failing due to this:

Screen Shot 2022-10-06 at 9 30 41 PM

can you apply smart rounding to the vault balance display so it rounds up to 5000?

zashton:
> @zashton the Cypress test is failing due to this:
> 
> Screen Shot 2022-10-06 at 9 30 41 PM
> 
> can you apply smart rounding to the vault balance display so it rounds up to 5000?

i've fixed this one, the final part of the cypress that is failing should be because the table that holds the transaction records doesn't exist

levity (approved):

Issue #1005, opened 2022-06-06
Show/Hide Circle Option via API


crabsinger:
**Problem**
Users can be added to any circle, and when in many circles or testing, some users belong to many circles with no way to remove themselves from a circle.

**Solution**
Enable API support (UI changes in separate issue) for the following use cases:
- `circle contributor` - Allow a user to remove themselves (update `deleted_at = current_date` in `users` table) from a circle. Note: column `deleted_at` already exists.
- `circle admin` - Fully delete a circle, hiding it for themselves and all members (update `deleted_at = current_date` in `circles` table.  Note: requires a new `circles.deleted_at` column to be created via migration

**Specification**
- Add a `deleted_at` column to circles to enable deletion of circles.
- Expose `deleted_at` via API for users and circles.
- Set permissions so users can delete themselves from circles, admins can delete circles they are admins of, and admins can (still) delete users from circles they admin.

Once this task is complete, the UI changes to expose these settings will follow

levity:
Note that deleting a user from a circle is already possible and uses a deleted_at column on the user table. Maybe there are actually 2 different user needs here?

1. Circle admin: delete a circle so it vanishes for all users (this would require a new circles.deleted_at column)

2. Circle user: remove yourself from a circle (this would just be new UI on top of the existing data schema)

crabsinger:
@levity thanks for the notes - I've broken this into a backend only task (UI to follow) and clarified the descriptions.

spydermonkee:
Note that when this issue is completed, we need to have UI design and F/E work defined as the next step.

karelianpie:
Hi @levity, I was beginning to work on this when I've noticed you've already done some work in https://github.com/coordinape/coordinape/pull/1213/files related to this, could you please confirm? Thank you!

levity:
> Hi @levity, I was beginning to work on this when I've noticed you've already done some work in https://github.com/coordinape/coordinape/pull/1213/files related to this, could you please confirm? Thank you!

Yes, that's right, @sshmm started it in #1196 and I was just helping run tests in #1213. Looks like just neglected to mark this task as in-progress when that work started.

crabsinger:
@sshmm are you still actively working on this?

sshmm:
@crabsinger  I was not working on it. I finished my work on #1196. I'd like to continue work on it if it's not completed

crabsinger:
@sshmm great - I'll assign this to you and let you continue working on it.

PR #1451, opened 2022-10-03
Add userDeleter middleware to allow users to delete themselves from circles

sshmm:
## Motivation and Context

finalizes #1005 

## Description

1- Check if the current user is circle admin or self deleting to delete user from circle

## Test and Deployment Plan
1- Used API explorer to self delete from a circle where the user is not admin to this circle
image
2- Try to delete other users in the same circle
image
3- Try to delete other users from another circle as an dmin
image


topocount (commented):
@sshmm this looks solid. My only request is that you add some jest unit tests for this file to demonstrate it behaves correctly. An example for this type of file can be found here: https://github.com/coordinape/coordinape/blob/910bacecdbf3ac68bbb1e5c2baa9ae6993db4eba/api/hasura/actions/_handlers/deleteContribution.test.ts

topocount (approved):
Thanks! it'd be good to test unauthorized cases in the unit tests too, for future reference.

PR #1460, opened 2022-10-04
Implement useContribution hook that accepts dateranges

topocount:
# Description

We are migrating away from epoch-specific dateranges to facilitate more
generalized stories in the app.

This is also a pre-req for using contribution integrations in the new
Contributions page

# Test Plan 

I tested the behavior of the dework integration and it is nearly in parity, save for a marginal improvement that gets us a bit further away from tight epoch-coupled logic.
# code maintenance plan

~~I preserved old hooks in their entirety in order to preserve the incumbent allocation
page functionality. These can be deleted when we remove the original allocation page.~~

I deleted the old hooks and updated the ProfileCard invocation accordingly. I tests the behavior has well. and it appears to have succeeded in decoupling contributions we pull in from epoch start dates, so that's an improvement.


src/hooks/useContributions.ts
```diff
@@ -38,7 +38,98 @@ const mockData: UserContributions = {
...
+          .filter(
+            integration =>
+              integration.type === 'dework' || integration.type === 'wonder'
+          )
+          .map(integration => ({
+            queryKey: `circle-integration-contributions-${integration.id}-${timeInput.start_date}-${timeInput.end_date}`,
+            queryFn: () => {
+              if (integration.type === 'dework') {
+                return fetch(
+                  `https://api.deworkxyz.com/integrations/coordinape/${
```
CryptoGraffe, line 58: this should probably be a const higher up in file

src/hooks/useContributions.ts
```diff
@@ -38,7 +38,98 @@ const mockData: UserContributions = {
...
+    integrations.data
+      ? integrations.data
+          .filter(
+            integration =>
+              integration.type === 'dework' || integration.type === 'wonder'
+          )
+          .map(integration => ({
+            queryKey: `circle-integration-contributions-${integration.id}-${timeInput.start_date}-${timeInput.end_date}`,
+            queryFn: () => {
+              if (integration.type === 'dework') {
```
CryptoGraffe, line 56: would be nice if these strings like 'dework' were consts

src/hooks/useContributions.ts
```diff
@@ -38,7 +38,98 @@ const mockData: UserContributions = {
...
+                  }?epoch_start=${timeInput.start_date}&epoch_end=${
+                    timeInput.end_date
+                  }&workspace_ids=${encodeURIComponent(
+                    integration.data.workspaceIds?.join(',') || ''
+                  )}`
+                )
+                  .then(res => res.json())
+                  .then(res => res as Response);
+              }
+              if (integration.type === 'wonder') {
```
CryptoGraffe, line 69: this should be a func

src/hooks/useContributions.ts
CryptoGraffe: and this whole block should be a func

src/hooks/useContributions.ts
```diff
@@ -38,7 +38,98 @@ const mockData: UserContributions = {
...
+export function useContributionUsers(timeInput: {
+  start_date: string;
+  end_date: string;
+}): UserContributions {
+  const integrations = useCurrentCircleIntegrations();
+  const responses = useQueries(
+    integrations.data
+      ? integrations.data
+          .filter(
+            integration =>
```
CryptoGraffe, line 50: this integration filtering should be a func perhaps? or a lookup map of integrations to funcs to handle them , somehow there is a cleaner pattern here without these inline string compares

src/hooks/useContributions.ts
```diff
@@ -38,7 +38,98 @@ const mockData: UserContributions = {
...
+          }))
+      : []
+  );
+  /**
+   * responses are individual responses from each integration
+   * looping over the responses from various integration stiching together to make a userContributions object that's easier to work eith
+   */
+
+  return useMemo(() => {
+    const combinedContribution: UserContributions = {};
```
CryptoGraffe, line 90: is there any way to sort these?

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -12,7 +12,7 @@ import {
 } from 'components';
 import { USER_ROLE_ADMIN, USER_ROLE_COORDINAPE } from 'config/constants';
 import { useNavigation } from 'hooks';
-import { useContributions } from 'hooks/useContributions';
+import { useDeprecatedContributions as useContributions } from 'hooks/useContributions';
```
CryptoGraffe, line 15: can we just ditch the deprecated one?

CryptoGraffe (commented):
looks good i just want a little refactoring to tighten up concerns and remove nested logic a bit

topocount:
> looks good i just want a little refactoring to tighten up concerns and remove nested logic a bit

Thanks @CryptoGraffe I just copy-pasta'd the original impls and modified the params as needed. Wasn't sure if a more thorough refactor was worth the time in others' eyes but will gladly do it.

src/components/ProfileCard/ProfileCard.tsx
karelianpie: or if not, use the actual deprecated name so that we know it should be used with caution when reading the code

topocount: done

src/hooks/useContributions.ts
topocount: why sort them? That sounds unnecessarily expensive. Indexing them makes more sense, no?

src/hooks/useContributions.ts
topocount: yeah i've create consts that live in the integration creation file.

src/hooks/useContributions.ts
topocount: done

src/hooks/useContributions.ts
topocount: your wish is my command

topocount:
@sshmm I made source mandatory in the refactor. Hope this helps!

src/components/ProfileCard/ProfileCard.tsx
```diff
@@ -124,7 +125,12 @@ const ProfileCardInner = ({
   const hideUserBio =
     (userBioTextLength > 93 && skillsLength > 2) || userBioTextLength > 270;
 
-  const contributions = useContributions(user.address);
+  const epochs = useSelectedCircle().circleEpochsStatus;
```
levity, line 128: i would flag this use of `useSelectedCircle`, but i guess ProfileCard is going away once the new allocation flow is done anyway

src/hooks/useContributions.ts
```diff
@@ -30,53 +41,93 @@ export interface UserContributions {
...
+    { title: 'Gib me tokens', link: 'http://gib.com', source: 'gib' },
+    { title: 'I crushed it', link: 'http://crush.com', source: 'crush' },
   ],
 };
 
-export function useContributionUsers(): UserContributions {
+const ensureSource =
+  (source: string) =>
+  (res: Response): Response => {
+    return {
```
levity, line 63: you can return an object literal without using `return` if you wrap it in parentheses: `() => ({ foo: 'bar' })`

src/hooks/useContributions.ts
```diff
@@ -104,11 +155,17 @@ export function useContributionUsers(): UserContributions {
...
 }
 
-export function useContributions(
-  address: string,
-  mock?: boolean
-): Array<Contribution> | undefined {
-  const userToContribution = useContributionUsers();
+export function useContributions(input: {
+  address: string;
+  start_date?: string;
```
levity, line 160: camelCase?

src/hooks/useContributions.ts
```diff
@@ -1,15 +1,26 @@
...
+} from 'pages/IntegrationCallbackPage/IntegrationCallbackPage';
 
-import { useCurrentCircleIntegrations } from './gql/useCurrentCircleIntegrations';
+import {
+  Integration,
+  useCurrentCircleIntegrations,
+} from './gql/useCurrentCircleIntegrations';
 
+interface TimeInput {
+  start_date: string;
```
levity, line 17: camelCase?

levity (changes requested):

src/components/ProfileCard/ProfileCard.tsx
topocount: I was gonna figure out how to parse the URL and query the circle Id, but I figured this was 1) going away and 2) most import to just get this call out of the `useContributions` hook.

src/hooks/useContributions.ts
topocount: with all the nesting this felt cleaner, but okay.

PR #1464, opened 2022-10-05
Overview page: toggle display of my/all circles

### LINKED ISSUE #1414: Give Users Option to hide circles they are not in
Hide circles you're not a member of by default.  If there are circles you are not part of, show a `show all circles` toggle switch.  No need to save any setting, just hide it by default.

### What's happening:
Currently all circles in an org are displayed regardless of a users membership status in those circles
### What should be happening:
Users should only see the circles they are a part of and have the ability to toggle display of all circles in the org. 
See Figma Design @Ivy can you please drop the design link here?

This toggle selection does not need to persist between page views 

Toggle text should read "Show all Circles" and "Show my Circles" 

#### Backend Considerations: None

sshmm:
## Motivation and Context

resolves #1414 

## Description

1- Hide non-member circles by default
2- Create toggle link with text Show all circles and Show only my circles
3-  add 300ms ease in transition for show/hide. used 'react-transition-group' for that

## Test and Deployment Plan
1- Go to OverviewPage click show all circles/show my circles only and check the displayed circles
expected: circles that you are not a member in should only appear when you click show all circles

image
image


src/pages/CirclesPage/CirclesPage.tsx
```diff
@@ -231,6 +265,8 @@ const CircleRow = ({ circle, onButtonClick }: CircleRowProps) => {
           '.circle-row-menu-indicator': { display: 'none' },
         },
         ...(nonMember ? nonMemberPanelCss : { cursor: 'pointer' }),
+        transition: 'opacity 300ms ease-in-out',
```
wingfeatherwave, line 268: Nice.

The borders are missing from the non-member circle. please add a transparent border to this panel css block, after line 261:
`border: '1px solid transparent',`

That way all the panels have a border and the non member panel will get its correct style as defined here:
```
const nonMemberPanelCss: CSS = {
  backgroundColor: '$background',
  borderColor: '$borderMedium',
};
```

wingfeatherwave: I went ahead and made this change and also addressed another issue with circleRow coming from `main`, as circle row is also being used on the new Org page, and the state prop isn't used there, so I made it optional

wingfeatherwave (approved):

package.json
```diff
@@ -74,6 +74,7 @@
     "react-query": "3.34.8",
     "react-router-dom": "6.2.1",
     "react-string-replace": "0.4.4",
+    "react-transition-group": "^4.4.5",
```
karelianpie, line 77: We should pin dependencies for security reasons

PR #1466, opened 2022-10-05
setup morgan

crabsinger:
Enable local dev logging by setting DEV_LOGGING=true in your env

Here's what logs look like

image


wingfeatherwave (approved):
Cool.  Will try this

PR #1473, opened 2022-10-05
Speed up nodemon build process

topocount:
Let's use SWC, a rust-based transpiler that is super quick.

I've also added a typecheck hook that does more thorough typechecking
concurrently on every start, and added the watch list to nodemon.json

reviewer: @crabsinger


nodemon.json
```diff
@@ -0,0 +1,9 @@
+{
+  "watch": ["api/", "api-lib/", "scripts/"],
+  "delay": 3000,
```
crabsinger, line 3: Why 3000? Seems a bit long. I think of the delay being useful if you are saving multiple files, but it seems long if I'm going to "edit code", "save", "reload" and need to wait for the reload to be processed?

nodemon.json
```diff
@@ -0,0 +1,9 @@
+{
+  "watch": ["api/", "api-lib/", "scripts/"],
```
crabsinger, line 2: Nice!

nodemon.json
```diff
@@ -0,0 +1,9 @@
...
+  "delay": 3000,
+  "execMap": {
+    "ts": "ts-node --swc"
+  },
+  "events": { "start": "yarn typecheck:backend" },
```
crabsinger, line 7:
```suggestion
```

Running 

❯ yarn typecheck:backend
yarn run v1.22.19
$ tsc -p tsconfig-backend.json --noEmit
✨  Done in 12.36s.

takes 12 seconds on my machine... so could we just omit this start event and let errors bubble up if necessary to have faster start times? 

Or if not that , run it async from the yarn start command somehow?

crabsinger: Actually, the slowness of this start only happened on the first run, and it seems to cache and start faster after that. 

I'm seeing starts of 12s locally, and reloads in 1-2 seconds.

crabsinger (approved):
@topocount one question on the delay!

Will using SWC be different than how we're compiling in prod, and could this cause errors somehow?

nodemon.json
topocount: prettier takes more than a second and this debounces any file changes due to orettier

nodemon.json
topocount: The initial lod is due to the frontend"'s initial compilation.

topocount:
> @topocount one question on the delay!
> 
> Will using SWC be different than how we're compiling in prod, and could this cause errors somehow?

Swc is used by vercel under the hood, so this actually bring us closer to how prod behaves.

I don't think it makes any semantic difference in behavior or risk of error since typescript is an open specification and our tsconfigs are observed by all environments

nodemon.json
levity: `src` needs to be in this list too:

```
# this is a lower bound because it excludes multi-line imports
❯ ack -l "import.*src" api api-lib
api/hasura/actions/_handlers/createUserWithToken.ts
api/hasura/actions/_handlers/allocationCsv.ts
api/hasura/actions/_handlers/createVault.ts
api/hasura/actions/_handlers/updateContribution.test.ts
api/hasura/actions/_handlers/logoutUser.ts
api/hasura/actions/_handlers/deleteContribution.test.ts
api/hasura/cron/recoverTransactions.ts
api-lib/signature.ts
api-lib/HttpError.ts
api-lib/images.ts
```

Issue #1459, opened 2022-10-04
explore Jest + Hasura testing infra


crabsinger:
Explore setting up a Jest + Hasura testing setup that would allow Jest testing of Hasura backends.

Ideas:
- use different accounts per test for greater idempotence
- rather than setup & teardown of Postgresql, instead create all the data you expect in the test itself.
- look for examples of this pattern on the web:
 https://dev.to/trigo/how-to-test-a-hasura-api-with-jest-36jh

PR #1486, opened 2022-10-10
feat: Jest + Hasura testing infra

karelianpie:
## Motivation and Context

 
We need a fast, easy, light weight way of testing our Hasura backend. Right now, we have E2E tests with Cypress, but they are not very fast.

## Description

> "use different accounts per test for greater idempotence"

If you look into `api-lib/__tests__/deleteContribution.test.ts` you'll see that we generate a unique address and then create a user and profile with that address. We will try to generate a unique address 100x, we can work on improving this logic for better idempotency.

> "rather than setup & teardown of Postgresql, instead create all the data you expect in the test itself"

That's what I've done in `api-lib/__tests__/deleteContribution.test.ts`, arguably we can try and create even more helpers that make it even easier to create entities;  Right now you'd need to create all the entities you need which means you'd need to have knowledge regarding the way these entities are connected between one another. For the contribution test above, would we want a `createContribution()` function that would know how to create a user, a circle, a profile and an org? Where do we draw the line, or should the person writing the test know that in order to create a contribution, there needs to be a user and a circle and so on?

On top of the above, we also have a way to create `admin` and `user` clients, which we can use to make the request as an admin or impersonating a particular user (with it's profile id and address), that's helpful to test permissions and that type of thing

## Test and Deployment Plan

Ran the test `deleteContribution.test.ts` and saw it pass

```cmd
$ /coordinape/node_modules/.bin/jest deleteContribution.test.ts
PASS tests/deleteContribution.test.ts (5.769 s)
  deleteContribution()
    ✓ should soft delete a contribution (145 ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        5.822 s
Ran all test suites matching /deleteContribution.test.ts/i.
```

## Reviewers

@crabsinger

## Related Issue

Closes https://github.com/coordinape/coordinape/issues/1459


api-lib/helpers/circles.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export async function createCircle(
+  client: any,
+  object?: Partial<CircleInput>
+): Promise<{ id: number }> {
+  let organizationId = object?.organization_id;
+
+  if (!organizationId) {
+    const organization = await createOrganization(client, { name: 'Org' });
```
crabsinger, line 30: Can we use `faker` for the org name as well.

api-lib/helpers/contributions.ts
```diff
@@ -0,0 +1,64 @@
...
+>;
+
+type DeleteContributionInput = typeof deleteContributionInput['_type'];
+
+type ContributionResult = Pick<
+  GraphQLTypes['contributions'],
+  'id' | 'description' | 'deleted_at' | 'user_id'
+>;
+
+export async function createContribution(
```
crabsinger, line 20: I almost wonder if we could re-use this functions in our app, beyond jsut testing. And also, if they're just for testing, then if `api-lib/helpers` is the right place for them as I can imagine it being unclear this is a helper ONLY for test.

api-lib/helpers/getUniqueAddress.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  return profiles.length === 0;
+}
+
+export const getUniqueAddress = async () => {
+  let i = 100;
+
+  while (i > 0) {
+    const address = HDNode.fromMnemonic(SEED_PHRASE)
+      .derivePath(getAccountPath(Number(Math.random().toString().slice(2, 11))))
```
crabsinger, line 33: I'm very surprised we're getting overlaps...

What if we just had a little helper function like:

function getRandomInt(max) {
  return Math.floor(Math.random() * max);
}

And then use that to choose a random number in the max getAccountPath range - which should REALLY not be already used, but in case it is then I guess we can verify it and try again.

api/hasura/auth.ts
```diff
@@ -8,12 +8,24 @@ import { adminClient } from '../../api-lib/gql/adminClient';
...
+      if (req.headers?.authorization === 'generate') {
+        // For generating libraries from inspection
+        res.status(200).json({
+          'X-Hasura-Role': req.headers?.['x-hasura-role'],
+        });
+        return;
+      }
+
+      // Skip auth checks when testing
+      if (req.headers?.authorization === 'test-skip-auth') {
```
crabsinger, line 21: ideally I'd like this to be in a IS_TEST_ENV so this is really clearly only for when tests are running

src/lib/zod/index.ts
```diff
@@ -47,6 +47,20 @@ export const createCircleSchemaInput = z
     'Either Protocol name should be filled in or a Protocol should be selected.'
   );
 
+export const createContributionSchemaInput = z
```
crabsinger, line 50: If we're adding these, it would be nice if we used them in the app...

crabsinger:
@karelianpie nice work getting this running. Left a few comments. 

So what is remaining on this PR before you think its complete and ready to merge (beyond PR review)?

It seems to me like we need to get a CI test runner in place (or do this after).

crabsinger (commented):
left a few comments - looks really good.

api-lib/gql/userClient.ts
```diff
@@ -0,0 +1,15 @@
+import { NODE_HASURA_URL } from '../config';
```
topocount, line 1: This file can probably be deleted, right

topocount (changes requested):
I hate to be a stick in the mud, but let's get rid of the `any` types, especially at function interfaces. The `typeof` TS operator might be a useful shortcut here.

api-lib/helpers/circles.ts
karelianpie: Definitely, done with https://github.com/coordinape/coordinape/pull/1486/commits/729b3ff990b65da0247d9ee19429bb579379b51e

api-lib/helpers/contributions.ts
karelianpie: You're right, moved them under `/api-test` with https://github.com/coordinape/coordinape/pull/1486/commits/cb825d4de93ba620e94d71fb76d22d6cc3efcd50

api-lib/helpers/getUniqueAddress.ts
karelianpie: I got one conflict one time running the test, so I'm assuming if we have a bunch of tests running in parallel that that could happen. Thus, added the uniqueness check, we could potentially try and discover a safer way to generate unique addresses, one that comes to mind is for example hardcode a huge number of addresses and then just use them rather than generate them at run time. TBD

For now, applied your suggestion with https://github.com/coordinape/coordinape/pull/1486/commits/3d6cd6df32580e071a981f8b9f8d8a6c85889d1c

api/hasura/auth.ts
karelianpie: I've started implementing this suggestion with https://github.com/coordinape/coordinape/pull/1486/commits/86b2a76d0f7ab094d9979ae228ec898a58090da3, although looking at how nodemon is setup in `nodemon.json`:
```json
{
  "watch": [
    "api/",
    "api-lib/",
    "scripts/",
    "src/lib",
    "src/common-lib",
    "./*.json"
  ],
  "delay": 300,
  "execMap": {
    "ts": "ts-node --swc"
  },
  "events": { "start": "yarn typecheck:backend" },
  "env": { "NODE_ENV": "development" },
  "ignore": ["api-test/__tests__"]
}
```
it appears that it will always start in the `"development"` environment, and that's what gets executed when

`yarn exec nodemon \
  -- scripts/serve_dev.ts $PORT $PROXY_PORT`

Wondering if we need a new script for starting nodemon with a test environment just for tests, unsure how that would affect the app if at all. For now just added a placeholder with `export const IS_TEST_ENV = process.env.NODE_ENV === 'development' || 'test';`

Thoughts?

src/lib/zod/index.ts
karelianpie: @levity mentioned that these were no longer used and they should even be co-located in the file that uses them, I've tried to just continue the existing pattern and added the few that were missing for my use case, but happy to do it differently if that's preferable. At the time I was not 100% sure what the best way to create the types for the helper functions was, so just went the zod way. Thoughts?

api-lib/gql/userClient.ts
karelianpie: I believe not, this file is to run a query impersonating as a user, looking at the test `api-test/__tests__/deleteContribution.test.ts`

```typescript
const client = userClient({
  profileId: profile.id,
  address: ADDRESS,
});

const contribution = await createContribution(client, {
  circle_id: circle.id,
  user_id: user.id,
});
```
I'm creating a GQL client impersonating as a user in order to create a contribution as them, then in the `api-lib/gql/makeThunder.ts` we map that to the headers that go to hasura as if we are the user. Without this file, we can only making requests as an admin, which will not allow us to test for example permissions as a normal user. Or create multiple clients impersonating different users interacting with one another.
```typescript
const mapOptionsToHeaders = (options: UserOptions) => ({
  'x-hasura-role': 'user',
  'x-hasura-address': options.address,
  'x-hasura-user-id': String(options.profileId),
});

export const makeThunder = ({ url, options }: ThunderOptions) => {
  return Thunder(async (...params) => {
    return apiFetch([
      url,
      {
        method: 'POST',
        headers:
          'adminSecret' in options
            ? {
                'x-hasura-admin-secret': options.adminSecret,
              }
            : {
                ...mapOptionsToHeaders(options),
                authorization: 'test-skip-auth',
              },
      },
    ])(...params);
  });
};

```

karelianpie:
> I hate to be a stick in the mud, but let's get rid of the `any` types, especially at function interfaces. The `typeof` TS operator might be a useful shortcut here.

No, you're totally right it was my bad. Refactored with https://github.com/coordinape/coordinape/pull/1486/commits/8ff99949e2ace95839f016c24cc8f4fdf9208a10

karelianpie:
> I think this has to wait for #1540 to get merged first. @karelianpie if you want to move that forward while I’m out, that’s fine with me, but waiting for me to be back is fine too. There might still be some non-trivial loose ends in that PR.

It might be worth to wait for you @levity, as I believe this is not extremely high priority and I'd like to start focusing on the Discord Bot project.

api/hasura/auth.ts
crabsinger: I'm hoping once we figure out how the runner for this works, we can just pass in some ENV_VAR in the invocation of that runner that lets us know we're in a TEST env...

crabsinger:
@karelianpie this looks really good to me - seems like maybe a good place to let it sit while you dive into the Discord work and then @levity can pick it up and merge it after #1540 is resolved. Nice work putting this together!

api-lib/gql/userClient.ts
topocount: @karelianpie thanks for that explanation! Since it isn't supposed to be used in production, can we rename it `mockUserClient`?

api-lib/gql/makeThunder.ts
```diff
@@ -1,15 +1,42 @@
...
+
+type ThunderOptions = {
+  url: string;
+  options: Options;
+};
+
+const mapOptionsToHeaders = (options: UserOptions) => ({
+  'x-hasura-role': 'user',
+  'x-hasura-address': options.address,
+  'x-hasura-user-id': String(options.profileId),
```
topocount, line 22: nit: use the .toString() method

scripts/ci.sh
```diff
@@ -57,17 +57,19 @@ until curl -s -o/dev/null http://localhost:"$CI_WEB_PORT"; do
...
-else
-  yarn cypress open > /dev/null
-fi
+# if [ -z "$LOCAL" ]; then
+#   yarn cypress run ${OTHERARGS[@]}
+# else
+#   yarn cypress open > /dev/null
+# fi
+
+yarn jest api-test/__tests__
```
topocount, line 75: I think these changes should be undone

src/lib/zod/index.ts
topocount: I don't think leaving these defs in this file is wrong, since it's at least consistent with the current state.

yarn.lock
```diff
@@ -23969,4 +23969,4 @@ zod@3.17.3:
 zwitch@^1.0.0:
   version "1.0.5"
   resolved "https://registry.yarnpkg.com/zwitch/-/zwitch-1.0.5.tgz#d11d7381ffed16b742f6af7b3f223d5cd9fe9920"
-  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
+  integrity sha512-V50KMwwzqJV0NpZIZFwfOD5/lyny3WlSzRiXgA0G7VUnRlqttta1L6UQIHzd6EuBY/cHGfwTIck7w1yH6Q5zUw==
```
topocount, line 23972: very nitty nit: why was the trailing newline removed here?

jest.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  roots: ['<rootDir>/api-test'],
+  transform: {
+    '^.+\\.tsx?$': 'ts-jest',
+  },
+  testRegex: '(/__tests__/.*|(\\.|/)([i]?test|[i]?spec))\\.[jt]sx?$',
+  moduleFileExtensions: ['ts', 'tsx', 'js', 'jsx', 'json', 'node'],
+};
```
topocount, line 8: We store all our jest config here: https://github.com/coordinape/coordinape/blob/7231ec28b8d53090fd58d386698ad0aa8b213be0/craco.config.js#L25-L58

and run test via `yarn craco test`, so this can be deleted. Sorry this wasn't communicated before.

scripts/ci.sh
levity: @karelianpie I can take care of this once I'm back. My plan is to merge this into the other branch anyway

api-lib/gql/userClient.ts
karelianpie: Good idea @topocount, renamed with https://github.com/coordinape/coordinape/pull/1486/commits/69dcf95aee4424365953c6a71eca723f762e1409

api-lib/gql/makeThunder.ts
karelianpie: Wondering why is that preferable, refactored with https://github.com/coordinape/coordinape/pull/1486/commits/269330308f6017784c34ea4d8a369e31169cf3c5

jest.config.js
karelianpie: Ah, should have searched for it better, thanks for pointing this out. Removed with https://github.com/coordinape/coordinape/pull/1486/commits/00dd7426ef469c69cbc2ed4514a9cf9bb065ffe4

karelianpie:
@crabsinger @topocount if my reply/commit resolved your suggestion/comment please resolve the conversation, so it's easier for me to skim through this PR and see if there are any outstanding comments to look into. Cheers 🙇

jest.config.js
karelianpie: @levity removing this file caused the SyntaxError:

> SyntaxError: Cannot use import statement outside a module

Unsure if we can use this craco config for our Jest setup in ci @topocount, perhaps need to bring the file back

api-lib/gql/makeThunder.ts
topocount: @karelianpie it's a matter of reliability in patterns in JS. Almost everything in JS has a `.toString` method, including Strings, so it's a reliable primitive for conversion. Using Class constructors for the purpose of conversion can have some weird side effects on the prototype chain, and can short circuit prototype inheritance, causing some weird bugs if you want to disambiguate something or do a comparison using the JS `typeof` or `instanceof` operators. 

I think it's insignificant in this case (and thus why it's a nit) but I like to reinforce good patterns when I have a chance.

jest.config.js
topocount: @karelianpie just FYI this can be fixed by uncommenting this line: 
https://github.com/karelianpie/coordinape/blob/438e6178130f7d4db490d7cb542df4d90b0e0901/scripts/ci.sh#L62-L67
and deleting the explicit `jest` invocation at the bottom of the file.

topocount (commented):
I've added some comments here to get tests working in CI again.

@karelianpie if you can show that this config works in CI, we can probably handle getting this merged along with @levity's work on #1540 either before or after he gets back.

Thanks for all your hard work on this!

api-lib/gql/makeThunder.ts
karelianpie: Thanks for the explanation! For this particular case as it is a header value, I thought we would probably want some safety in case the value is `null` or `undefined` which would just throw with `.toString()` but with a class would just be converted to `"null"` or `"undefined"` -- in general your argument makes sense, and I agree with you that for this case it's probably not relevant.

karelianpie:
> I've added some comments here to get tests working in CI again.
> 
> @karelianpie if you can show that this config works in CI, we can probably handle getting this merged along with @levity's work on #1540 either before or after he gets back.

It works @topocount! Only had to refactor the cypress gqlQuery helper https://github.com/coordinape/coordinape/pull/1486/commits/d0fde330efda26956e686ce633782f9d352db291

The test from this PR (i.e. `api-lib/__tests__/deleteContribution.test.ts`) ran here: https://github.com/coordinape/coordinape/actions/runs/3331806519/jobs/5511972379#step:11:337

topocount (approved):

api-lib/gql/makeThunder.ts
topocount: @karelianpie yes, there are other considerations if the variable is nullish. Fortunately, those are logical exceptions to to the rule the the `toString` method and they throw a TypeError and more importantly, fail typechecking

topocount (dismissed):

levity:
* main
  * **PR #1539** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1539" target="_blank">Graphite</a> 
    * **PR #1540** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1540" target="_blank">Graphite</a> 
      * **PR #1486** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1486" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1486?utm_source=stack-comment).

levity:
Github auto-closed this one when I merged #1540. Continued in #1646

PR #1491, opened 2022-10-10
Logging Fixed Payment distribution claim amounts

zashton:
## Motivation and Context

So we could track historical fixed payment amounts at the time of distribution.

## Description

Saving the fixed payment amounts for each contributor upon distribution into the DB

## Test and Deployment Plan

Updated unit tests should pass


levity (changes requested):
I think we should avoid pushing more fixed payment amount info into the merkle distributor code. This lowest level of the code should stay focused on just what is necessary to generate a merkle tree & proof for the distributor contract. The fixed payment amounts can be saved to the db higher up.

This is important, because fixed payments is basically our first strategy, and we're going to have more in the future, and we don't want to change the whole stack of code here for each new thing.

Suggestion: In `useSubmitDistribution.ts`, it seems like we could take the `fixedGifts` data and insert it into `distribution_json` directly around line 142 after doing some light transformations, without needing to change the logic for `createDistribution` at all.

Then there's no need to change the claims data model either. When you're rendering the allocations table, just pull the fixedGifts off of `distribution_json` again and get the value for each address.

PR #1501, opened 2022-10-11
Update sentry packages

crabsinger:
Upgrading to sentry versions 7.15


api-lib/HttpError.ts
```diff
@@ -3,6 +3,7 @@ import * as Sentry from '@sentry/node';
 // Importing @sentry/tracing patches the global hub for tracing to work.
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import * as Tracing from '@sentry/tracing';
+import { SeverityLevel } from '@sentry/types';
```
topocount, line 6:
```suggestion
import { SeverityLevels } from '@sentry/types';
```

api-lib/HttpError.ts
```diff
@@ -97,7 +99,7 @@ export function errorResponseWithStatusCode(
         scope.addBreadcrumb({
           category: 'query-path',
           message: error.details.path,
-          level: Sentry.Severity.Debug,
+          level: SEVERITY_DEBUG,
```
topocount, line 102: Try this instead, so we can avoid the unneeded typecast on line 16:
```suggestion
          level: SeverityLevels.Debug,
```

topocount (commented):
Thanks for doing this, @crabsinger! I'd like to avoid the unnecessary typecast when Sentry provides an enum for use.

topocount (approved):

PR #1503, opened 2022-10-12
Logging Fixed Payment distribution claim amounts

zashton:
## Motivation and Context

So we could track historical fixed payment amounts at the time of distribution.

## Description

Saving the fixed payment amounts for each contributor upon distribution into the DB

## Test and Deployment Plan

Updated unit tests should pass


src/pages/DistributionsPage/AllocationsTable.tsx
```diff
@@ -88,7 +88,7 @@ export const AllocationsTable = ({
     { title: 'Fixed Payments', css: styles.alignRight },
   ];
   if (combinedDist) {
-    headers.push({ title: 'Funds Allocated' });
+    headers.push({ title: 'Funds Allocated', css: styles.alignRight });
```
topocount, line 91: nice, I like this

src/pages/DistributionsPage/DistributionsPage.tsx
```diff
@@ -100,21 +101,35 @@ export function DistributionsPage() {
...
     dist?: typeof epoch.distributions[0]
   ) => {
-    if (!id || !dist) return 0;
-    return (
-      (dist.claims.find(c => c.profile?.id === id)?.new_amount || 0) *
-      dist.pricePerShare.toUnsafeFloat()
-    );
+    if (!id || !dist) return { claimed: 0, fixedPayment: 0, circleClaimed: 0 };
+    const claim = dist.claims.find(c => c.profile?.id === id);
+    const pricePerShare = dist.pricePerShare.toUnsafeFloat();
```
topocount, line 125: This is the case across our app I supposed, but why not convert this straight from a BigNumber to a String? It doesn't seem like it'd be a much bigger lift to make this more robust by avoiding the lossy conversion to a JS `Number` type.

src/pages/DistributionsPage/DistributionsPage.tsx
```diff
@@ -100,21 +101,35 @@ export function DistributionsPage() {
...
+    const fixedPaymentAmt = getUserClaimedFixedPaymentAmt(dist, claim?.address);
+    return {
+      claimed: (claim?.new_amount || 0) * pricePerShare,
+      fixedPayment: fixedPaymentAmt * pricePerShare,
+      circleClaimed:
+        ((claim?.new_amount || 0) - fixedPaymentAmt) * pricePerShare,
```
topocount, line 131: Same goes for here: Why not do BigNumber math here?

src/pages/DistributionsPage/AllocationsTable.tsx
```diff
@@ -158,34 +158,28 @@ export const AllocationsTable = ({
...
-              {user.circle_claimed
-                ? `${smartRounding(
-                    circleDist &&
-                      circleDist.distribution_type ===
-                        DISTRIBUTION_TYPE.COMBINED
-                      ? user.circle_claimed - user.fixed_payment_amount
-                      : user.circle_claimed
-                  )} ${tokenName || 'GIVE'}`
+              {circleDist
+                ? `${smartRounding(user.circleClaimed)} ${tokenName || 'GIVE'}`
```
topocount, line 162: In order to be truly precise, this would mean refactoring `smartRounding` to use ethers [formatUnits](https://docs.ethers.io/ethers.js/v3.0/html/api-utils.html#ether-strings-and-wei), but I think it'll minimize tech debt later.

topocount (commented):
This looks good, but I think a relatively minor refactor to avoid JS Numbers could save a bunch of headaches and bugs later.

Open to disagreement though

levity:
I'll try to review this later today. Sorry for the delay

src/pages/DistributionsPage/DistributionsPage.tsx
zashton: tried using `BigNumber` but `pricePerShare` is always has decimals and would cause an underflow 
or should i just change it to using a `FixedNumber` instead

levity: the amount & new_amount columns are unfortunately already lossy due to bad support for the NUMERIC postgres column type in Zeus. this is one of those pieces of vault tech debt i was talking about. so there's nothing to be gained in being more precise with calculations here until we fix that issue. (when we have to be exact, i.e. interactions with the smart contracts, we use the values stored in `distributions.distribution_json` instead)

levity (changes requested):

src/common-lib/distributions.ts
```diff
@@ -0,0 +1,91 @@
+import { FixedNumber } from 'ethers';
+import { formatUnits } from 'ethers/lib/utils';
+
+interface Distribution {
```
zashton, line 4: @levity any idea how to share the types here between BE/FE without hardcoding ?

src/common-lib/distributions.ts
```diff
@@ -0,0 +1,88 @@
+import { FixedNumber } from 'ethers';
```
levity, line 1: can you add some tests for this file to verify that the calculations are correct? ideally a new dev would be able to look at the tests and quickly understand how to use these methods by seeing example pairs of input-output.

src/pages/DistributionsPage/DistributionsPage.test.tsx
```diff
@@ -104,7 +104,13 @@ test('render with a distribution', async () => {
           pricePerShare: FixedNumber.from('1.08'),
           distribution_type: 1,
           vault: mockEpochData.circle.organization.vaults[0],
-          claims: [{ new_amount: 10, profile: { id: recipient.profile.id } }],
+          claims: [
+            {
+              new_amount: 10,
+              profile_id: recipient.profile.id,
+              profile: { id: recipient.profile.id },
```
levity, line 111: this mock doesn't match the query

src/common-lib/distributions.ts
zashton: updated with tests

src/pages/DistributionsPage/DistributionsPage.test.tsx
zashton: fixed

crabsinger (commented):
Where are these being stored in the DB? Is this only for logging, or will this also enable the historical display of fixed payments for previous epoch payments?

zashton:
> Where are these being stored in the DB? Is this only for logging, or will this also enable the historical display of fixed payments for previous epoch payments?

yes these are to display in the allocation table as historical data, currently its not logged so if you update the user fixed payment amount the past epoch fixed payment allocations goes wrong.

api/hasura/actions/_handlers/allocationCsv.ts
```diff
@@ -115,63 +117,61 @@ export function generateCsvValues(
...
-    const rowValues: (string | number)[] = [
-      idx + 1,
-      u.name,
-      u.address,
-      received,
-      u.sent_gifts.length
-        ? u.sent_gifts
+  return (
+    users?.map((u, idx) => {
+      const claimAmt = circleDist?.claims.find(
```
topocount, line 122: This has an `any` type, which isn't great. Since there are so many nullish fallbacks, it might be good to type this as a number and set a fallback for it.

topocount (approved):
I've added some nits but am otherwise feeling quite good about this.

api/hasura/actions/_handlers/allocationCsv.test.ts
```diff
@@ -60,33 +61,67 @@ function getMockCircleDistribution(type?: DistributionType | undefined) {
     case DISTRIBUTION_TYPE.FIXED:
       distributions.push({
         distribution_type: DISTRIBUTION_TYPE.FIXED,
+        distribution_json: {
```
topocount, line 64: Sorry to be a stick in the mud about this, but an easy way to track that this mock matches the query is to type this function to return a piece of the actual query. So if the query is ever updated, this will fail to typecheck until it's updated as well and will be more maintainable.

src/common-lib/distributions.ts
```diff
@@ -0,0 +1,88 @@
...
+    fixedGifts?: Record<string, string>;
+    pricePerShare?: FixedNumber;
+  }) => {
+    return {
+      fixedPaymentAmt: getUserClaimedFixedPaymentAmt({
+        decimals,
+        fixedGifts,
+        address,
+      }),
+      pricePerShare: pricePerShare?.toUnsafeFloat() || 1,
```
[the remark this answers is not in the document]
zashton: > Based on the use of calc in the following lines, can we make `pricePerShare` a required parameter?
> 
> Is there a more general case where pricePerShare isn't available?

yeah its possible, the `pricePerShare` will not be logged yet in the DB if the specific fixed or gift distribution hasn't been initiated yet

src/common-lib/distributions.ts
```diff
@@ -0,0 +1,88 @@
...
+  circleDistDecimals?: number;
+  circleDistClaimAmount?: number;
+  circleDistPricePerShare?: FixedNumber;
+  circleFixedGifts?: Record<string, string>;
+}) => {
+  let fixedPayment = 0,
+    circleClaimed = 0;
+  if (!address) return { fixedPayment, circleClaimed };
+
+  const calc = ({
```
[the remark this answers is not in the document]
zashton: updated

api/hasura/actions/_handlers/allocationCsv.ts
zashton: updated

api/hasura/actions/_handlers/allocationCsv.test.ts
zashton: great suggestion have updated it

topocount (approved):
LGTM thanks!

PR #1504, opened 2022-10-12
Create new column on users to track source of invitation

### LINKED ISSUE #1490: Create new column on `users` to track source of invitation
Goal:

We want to be able to know, track, and funnelize how a user enters a circle.

Solution: Create a new column on `users` named `entrance` to track how the user was added, either via:
- admin self-add: admin created circle
- magic link
- manual address entry
- CSV
- vouched in

Then wherever we add users to circles, we add the information.

sshmm:
## Motivation and Context
resolves #1490

## Description
1- Add column entrance to users table
2- Update create, insert and update user mutations to affect this column according to #1490

## Test and Deployment Plan
I test this by adding users using csv, magic link, and eth address by add member form and checking the entrance column in users table
Tested Creating a circle and check the created user in this circle
Vouching for a user if required vouchers are more than one and also when only one voucher is required (auto creating)

Screenshot from 2022-10-12 18-27-05

image

image


hasura/migrations/default/1665513818085_alter_table_public_users_add_column_entrance/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."users" add column "entrance" text
```
crabsinger, line 1: https://www.postgresql.org/docs/14/datatype-character.html

We don't need this to be a `text` type, just a `varchar` should be good, these will all be short strings.

api-lib/event_triggers/createVouchedUser.ts
```diff
@@ -50,7 +50,7 @@ async function handler(req: VercelRequest, res: VercelResponse) {
       const { insert_users_one } = await adminClient.mutate(
         {
           insert_users_one: [
-            { object: { name, address, circle_id } },
+            { object: { name, address, circle_id, entrance: 'vouched in' } },
```
crabsinger, line 53: Let's make all the `entrance` values have no spaces, so `vouched-in`, `admin-self-add` etc, so its a bit easier in logs and dbs to search and parse these.

hasura/migrations/default/1665513818085_alter_table_public_users_add_column_entrance/up.sql
sshmm: done

levity (dismissed):
Please make a single enum with all the names, like we do with [`Asset`](https://github.com/coordinape/coordinape/blob/ccc3ef83d075e3f0a499a4b3fe9fdd48d0a25234/src/lib/vaults/index.ts#L8); then use that type instead of String where possible

sshmm:
> Please make a single enum with all the names, like we do with [`Asset`](https://github.com/coordinape/coordinape/blob/ccc3ef83d075e3f0a499a4b3fe9fdd48d0a25234/src/lib/vaults/index.ts#L8); then use that type instead of String where possible

is it ok to add it in two different locations like I did?

levity:
Can you try putting the constant somewhere in src/common-lib instead of duplicating it?

api-lib/constants.ts
```diff
@@ -18,3 +18,11 @@ export const DISTRIBUTION_TYPE = {
   FIXED: 2,
   COMBINED: 3,
 };
+
+export enum INVITATION_SOURCE {
```
levity, line 22: maybe this should be called Entrance since the user's db column is called that

hasura/migrations/default/1665513818085_alter_table_public_users_add_column_entrance/up.sql
crabsinger: @sshmm we can remove this entire migraiton file now 1665513818085, since you added a new migraiton with the new type.

sshmm: I removed the other one and updated the types in the first one

src/pages/AddMembersPage/NewMemberList.tsx
```diff
@@ -55,6 +57,7 @@ const NewMemberList = ({
         .object({
           address: zEthAddress.or(z.literal('')),
           name: zUsername.or(z.literal('')),
+          entrance: z.string().optional(),
```
topocount, line 60: Out of curiosity, why make this optional? Don't we want to record this everywhere? Zod also know about typescript enums if you want to get extra fancy.

hasura/migrations/default/1665513818085_alter_table_public_users_add_column_entrance/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."users" add column "entrance" varchar
+null;
```
topocount, line 2: At the risk of increasing scope here significantly, we could use a [Hasura Enum](https://hasura.io/data-hub/data-models-and-authorization/schema-share-enum-reference-table/) instead of a string and then zeus would generate the enum for us, rather than us just enforcing it at the typescript layer manually.

This is just a suggestion if we have the appetite for changing the lift on this

crabsinger:
> Can you try putting the constant somewhere in src/common-lib instead of duplicating it?

Does this work for you two? @levity @sshmm 

Trying to run this I'm seeing type errors and `yarn typecheck:backend` fails ...

So maybe that's not a good shared location for it, or something else isn't working with it being there?

src/common-lib/constants.ts
```diff
@@ -0,0 +1,7 @@
+export enum ENTRANCE {
+  ADMIN = 'admin-self-add',
```
crabsinger, line 2: let's use the value: `circle-create-initial-admin`
```suggestion
  ADMIN = 'admin-self-add',
  ADMIN = 'circle-create-initial-admin',
```

src/common-lib/constants.ts
```diff
@@ -0,0 +1,7 @@
...
+  LINK = 'magic-link',
+  ETH = 'manual-address-entry',
```
crabsinger, line 4: let's rename `ETH` to something a bit more readable, how about
```suggestion
  MANUAL = 'manual-address-entry',
```

src/lib/zod/index.ts
```diff
@@ -125,6 +125,7 @@ export const createUserSchemaInput = z
     non_receiver: z.boolean().optional(),
     role: z.number().min(0).max(1).optional(),
     fixed_payment_amount: z.number().min(0).max(100000000000).optional(),
+    entrance: z.string().optional(),
```
crabsinger, line 128: Seems to me that `    entrance: z.string(),` is fine, and no need for the optional? Is that raising any errors for you @sshmm ?

src/pages/AddMembersPage/NewMemberList.tsx
crabsinger: How about using this commit: https://github.com/coordinape/coordinape/commit/7e1281cc739b9e144dd25b509beffb143c929fa6

- We make the field mandatory.
- We create a hidden form field to hold the data

I tried it and validaiton still passes and the values are correct as well.
@topocount @sshmm thoughts?

crabsinger (commented):
Looking good, a few more cleanup items and finalizing things

src/pages/AddMembersPage/NewMemberList.tsx
sshmm: What is the benefit of doing that? this will only be used in two cases CSV or manual address entry and the user will not have access to modify that so it doesn't need validation or to be mandatory

sshmm: @crabsinger  I cherry picked your commit and tested it. it had an issue if you tried to manually add after csv because the field doesn't have a default value to reset to. I fixed it in [5273807](https://github.com/coordinape/coordinape/pull/1504/commits/52738076b0a85c6e91a5b3d45448b76e8bb089db)

crabsinger: @sshmm I think the benefit is that its more apparent that this value is part of the form, and allows us to enforce the non-null zod validation, and potentially easier to reason about in the future, but could agree with it being fine to do it another way. Curious @topocount thoughts on if he thinks this is better or not.

crabsinger (dismissed):

crabsinger:
@mergifyio queue

crabsinger:
@mergifyio queue

crabsinger:
@sshmm so after this we now want to track this field in Mixpanel.

PR #1513, opened 2022-10-17
Update Receive Info on opt out change

### LINKED ISSUE #1475: Opting-Out doesn't update the X GIVE in <ReceiveInfo> in nav bar
When you have give allocated to you and opt-out, the ReceiveInfo component doesn't update properly to reflect 0 GIVE. This doesn't matter if you are using the give beta or the old flow. I tried quickly to change it to do:

```
const totalReceived = myUser.give_token_received;
```

And that updates, but it was showing a negative amount in some cases until refresh, so more work needs to be done here.

sshmm:
## Motivation and Context
resolves #1475 

## Description
1- Fetch the receive data by react-query
2- refetch the data and rerender on opt-out change
3- used user.non_receiver to handle corner cases for the current epoch because gift data won't be ready from hasura immediately after opt out

## Test and Deployment Plan
1- opt out by old and new give after receiving gives from another user
expected: received gives should change to 0 immediately
2- Check last epoch received gives if there is no current epoch and still opted in
3- Check last epoch received gives if there is no current epoch and opted out after

Screenshot from 2022-10-17 12-28-04
tent.com/34943689/196160370-db76f988-4bf6-4116-a6a6-e5a078184ce0.png)

### DIFF
--- a/src/components/ReceiveInfo.tsx
+++ b/src/components/ReceiveInfo.tsx
@@ -43,6 +43,9 @@ export const ReceiveInfo = () => {
   );
   const noEpoch =
     !data?.myReceived?.currentEpoch[0] && !data?.myReceived?.pastEpochs[0];
+  //handle if member was a receiver and no current epoch
+  const currentNonReceiver =
+    data?.user?.non_receiver && data?.myReceived?.currentEpoch[0];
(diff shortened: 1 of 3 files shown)


crabsinger (approved):
looks good - merging

PR #1520, opened 2022-10-18
pricePerShare in Hasura

levity:
## Motivation and Context

With Yearn vaults, we often have to calculate `pricePerShare` to convert between wrapped and unwrapped amounts. When all other data comes through Hasura, having to make additional async requests to fetch data from an Ethereum provider is unwieldy ([e.g.](https://github.com/coordinape/coordinape/blob/main/src/pages/DistributionsPage/queries.ts#L122)). Ideally it would all happen through Hasura.

## Description

This PR:
- adds a new serverless function that stands up a tiny GraphQL server whose only query operation is a way of calculating pricePerShare
- adds the remote schema to Hasura config
- adds a remote relationship in Hasura between `vaults` and `price_per_share`

## Test and Deployment Plan

- [x] deploy to staging to verify that Hasura still works 
- [x] run a test GraphQL query, e.g.:
```
{
  organizations_by_pk(id: [id of some org with vaults]) {
    vaults {
      id
      symbol
      token_address
      price_per_share
    }
  }
}
```

## Notes

There's lots of room for performance improvement via caching. And maybe we'll decide that this particular use case isn't so important, but at least now we have example code for setting up a remote schema.
- I found myself wishing we already had a monorepo to keep dependencies separate. I had to work with Apollo Server v3, because v4 (just released) uses newer Typescript features, which confuses our linter.

### DIFF
--- a/api/hasura/remote/vaults.ts
+++ b/api/hasura/remote/vaults.ts
@@ -1,0 +1,83 @@
+// adapted from:
+// https://www.apollographql.com/docs/apollo-server/v3/api/apollo-server
+// https://www.apollographql.com/docs/apollo-server/v3/integrations/middleware/
+// https://github.com/vercel/next.js/pull/30082
+
+import deploymentInfo from '@coordinape/hardhat/dist/deploymentInfo.json';
+import {
+  RegistryAPI__factory,
+  VaultAPI__factory,
+} from '@coordinape/hardhat/dist/typechain';
+import { AddressZero } from '@ethersproject/constants';
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+import { ApolloServerPluginLandingPageDisabled } from 'apollo-server-core';
+import { ApolloServer, gql } from 'apollo-server-express';
+import debug from 'debug';
+import { BigNumber, FixedNumber } from 'ethers';
+
+import { getProvider } from '../../../api-lib/provider';
+const log = debug('vaultsql:handler');
+
+// The GraphQL schema
+const typeDefs = gql`
+  type Query {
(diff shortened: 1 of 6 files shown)


zashton (dismissed):
image

remote schema works for me,
but had to click on reload to make the hasura metadata consistent
hasura also has built in response caching that we can use here
https://hasura.io/docs/latest/queries/response-caching/

levity:
Welp, deployed to staging, and I'm seeing an issue that I didn't see on dev, possibly because the Hasura version numbers are different between the two environments.

Screen Shot 2022-10-22 at 3 36 49 PM

levity:
* main
  * **PR #1520** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1520" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1520?utm_source=stack-comment).

crabsinger:
> Welp, deployed to staging, and I'm seeing an issue that I didn't see on dev, possibly because the Hasura version numbers are different between the two environments.

Yeah, so maybe we need to figure out and resolve https://github.com/coordinape/coordinape/pull/1429/files before this goes ahead.

CryptoGraffe:
The hasura upgrade is done so maybe this can be retested now?

levity:
> The hasura upgrade is done so maybe this can be retested now?

Still running into issues; it appears unrelated to the Hasura version after all. I split out just the API handler into #1624 so I can deploy it first for further diagnosis. This PR now contains only the Hasura metadata changes to add the remote schema and relationship.

levity:
Merged this one back down into #1624 so it deploys all at once

PR #1521, opened 2022-10-18
upgrade to Node 16

levity:
- [ ] toggle Vercel to use 16.x before merging this to main


levity:
manually toggled Vercel node version to 16.x before pushing, then toggled it back

topocount (approved):
approved, pending passing tests

PR #1522, opened 2022-10-18
Re-enable webpack stdout to console

topocount:
I've run into some frustrating situations where I can't tell what
webpack is doing to my machine. to compensate for re-enabling the
logging of misleading URLs that we proxy, I've changed our
call to action link to bold.

image

### DIFF
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,6 +1,14 @@
 {
-  "watch": ["api/", "api-lib/", "scripts/", "src/lib", "src/common-lib"],
+  "watch": [
+    "api/",
+    "api-lib/",
+    "scripts/",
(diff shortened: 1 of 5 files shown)


levity (approved):

nodemon.json
```diff
@@ -1,6 +1,14 @@
...
+  "watch": [
+    "api/",
+    "api-lib/",
+    "scripts/",
+    "src/lib",
+    "src/common-lib",
+    "*.json"
+  ],
   "delay": 300,
+  "verbose": true,
```
levity, line 11: can we turn this off? it's had this effect:

Screen Shot 2022-10-19 at 3 24 00 PM

PR #1528, opened 2022-10-19
Fix allocate debounce

CryptoGraffe:
## Motivation and Context

Allocate setTimeout clearing wasn't really working and its a mystery, tried to debug with levity but I decided to get rid of it anyway. This is actually a better and more react-friendly impl that has some extensibility to it. I use a state machine and changed all the setTimeouts to lodash debounce and was able to eliminate some. 

## Test and Deployment Plan

Try to allocate and edit notes and edit epoch statement. 

## Reviewers

@levity


CryptoGraffe:
Yes

PR #1532, opened 2022-10-19
Disable self.id profile data

levity:
This code makes a bunch of requests when loading a list of users, but Self.ID is not widely used enough to be worth it. We can revisit it in the future.

See #215, #225 for more context


package.json
```diff
@@ -28,7 +28,6 @@
     "@radix-ui/react-popover": "^0.1.6",
     "@radix-ui/react-radio-group": "^0.1.5",
     "@radix-ui/react-select": "0.1.1",
-    "@self.id/core": "0.1.0",
```
crabsinger, line 31: @levity any reason you didn't remove the package as well? I went ahead and removed this... in https://github.com/coordinape/coordinape/pull/1532/commits/215bc5e088cb49676cfe34d72ab3fc4ab7a5ad3d

crabsinger (approved):
Looks good and thanks for doing this! - this has annoyed me for a while, save the bandwidth if not being used.

package.json
levity: Good catch!

PR #1539, opened 2022-10-20
use default internal ports for docker services

levity:
There may be an opportunity to simplify our service config so that:

- it's easier to explain to new devs
- it's easier to reason about
- it's easy to spin up multiple sets of the containers that don't collide, for CI or other purposes

This is a small change that I hope leads in that direction, a prelude to #1540

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -8,7 +8,7 @@ services:
     volumes:
       - db_data:/var/lib/postgresql/data
     ports:
-      - ${LOCAL_POSTGRES_PORT}:${LOCAL_POSTGRES_PORT}
+      - ${LOCAL_POSTGRES_PORT}:5432
     environment:
       POSTGRES_PASSWORD: ${LOCAL_POSTGRES_PASSWORD}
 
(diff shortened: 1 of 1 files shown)


levity:
* main
  * **PR #1539** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1539" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1539?utm_source=stack-comment).

docker-compose.yml
```diff
@@ -43,19 +43,18 @@ services:
...
     ports:
-      - ${CI_POSTGRES_PORT}:${LOCAL_POSTGRES_PORT}
+      - ${CI_POSTGRES_PORT}:5432
 
   graphql-engine-ci:
     <<: *graphql_engine
     profiles: ['ci']
     ports:
-      - ${CI_HASURA_PORT}:${CI_HASURA_PORT}
+      - ${CI_HASURA_PORT}:8080
```
karelianpie, line 52: Should this be 8081?
```suggestion
      - ${CI_HASURA_PORT}:8081
```

levity: The second value is the internal port; the first value is the one that's exposed to outside the container. I'm intentionally reverting these to the default because it reduces the amount of configuration needed for the service.

levity: Anyway, I found a way to get this all to work that will allow removing those two `-ci` containers entirely.

crabsinger:
@levity not sure I understand how this is an improvement towards  "it's easy to spin up multiple sets of the containers that don't collide, for CI or other purposes"

I agree hard-coded port values would be slightly easier to reason about, but I think there are use cases and reasons why individuals would want/need to override these on their local machines (maybe you run dev env on the same machine and wnat to override one of them to use different ports.

So I generally feel opposed to this change because it limits a key functionality that I can image folks would need.

I would support clearly organizing and defining a consistent nomenclature, defaults, and system for all our ports, and make it easier to reason, see, and change those port definitions, but to completely hard-code this and remove any ability to change them feels a bit forceful to me...

levity:
That’s not what this does
It only sets the default port *inside* the container
You can still use the env vars to change the port that is exposed on your
machine
It may make more sense in the context of the “upstack” PR (that’s Graphite
terminology), where it’s become possible to start arbitrarily many sets of
environments, not just 1 dev and 1 testing

levity:
^ @crabsinger

docker-compose.yml
```diff
@@ -43,19 +43,18 @@ services:
...
     depends_on:
       - postgres-ci
     environment:
       <<: *graphql_engine_environment
-      HASURA_GRAPHQL_DATABASE_URL: postgres://${LOCAL_POSTGRES_USER}:${LOCAL_POSTGRES_PASSWORD}@postgres-ci:${LOCAL_POSTGRES_PORT}/${LOCAL_POSTGRES_DATABASE}
-      HASURA_GRAPHQL_SERVER_PORT: ${CI_HASURA_PORT}
```
topocount, line 58: @karelianpie this is why we're able to reset the poron line 52 to the default.

topocount (approved):

topocount:
I'm very pleased with this simplification. Thanks!

crabsinger:
@levity ah, ok - so this is just the internal ports. 👍

PR #1548, opened 2022-10-20
Fixup/contributions allocations UI fixes

wingfeatherwave:
UI fixes for allocations and contributions

* line height and font size unification / dialing for drawer fields
~~* persistent `✔ Saved` indicator for when text has been saved in the past (ie: not blank)~~
* css animations for hover states 
* remove double panel bubbles for contributions in give drawer.  use separator lines instead

TBD: adding an `Autosave` hint text beneath the initial form field to be edited.

### DIFF
--- a/src/components/FormInputField.tsx
+++ b/src/components/FormInputField.tsx
@@ -20,6 +20,7 @@ type TFormInputField<TFieldValues extends FieldValues> = {
   textArea?: boolean;
   infoTooltip?: string;
   description?: string;
+  placeholder?: string;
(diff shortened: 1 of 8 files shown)


crabsinger:
I think one problem with this design is that it says "Saved" when the content is not saved.

I think the "Saved" text needs to dissapear whenever there have been changes, to indicate that auto-save is enabled and waiting for a pause to trigger. Since the data is not instantly saved when I add a new char on to a saved text field.

crabsinger (approved):
looks good.

PR #1555, opened 2022-10-24
delete amplitude since we are favoring mixpanel

CryptoGraffe:
## Motivation and Context

We were testing out amplitude and mixpanel for doing some event analysis, mixpanel seems fine so we are removing amplitude. 

## Description

deletes sendInteractionEventToAmplitude event handler and all related dependencies


CryptoGraffe:
Dang it I did a search but somehow missed this

crabsinger:
and you have a requested review on #1536  as well @CryptoGraffe ...

PR #1556, opened 2022-10-24
adding sandbox to org model and interaction events

CryptoGraffe:
## Motivation and Context

New sandbox field will be used for sample/example/test orgs and circles. Getting started with the basic data model and flag on interaction events and plan to stack more PRs on top of this work.


CryptoGraffe:
* main
  * **PR #1556** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1556" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1556?utm_source=stack-comment).

hasura/migrations/default/1666579812519_alter_table_public_organizations_add_column_sandbox/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
```
crabsinger: two migration files

crabsinger: can we delete the second migration file and take the down migration and put it here.

crabsinger (approved):
@CryptoGraffe looks good - some cleanup requested on the migration files, so I went ahead and cleaned them up.

Only remaining annoyance is lack of consistent nomenclature for boolean fields in this table.

image

Feels like maybe this should be a `is_sandbox` field, but I think that's unncessary, so I guess I jsut don't like "is_verified" much...

Issue #1558, opened 2022-10-24
Unable to enter organization when creating a new Circle if the user doesn't have an account.


reeserj:
**Summary:**

1. **What is actually happening:**
New users are unable to create a circle because the ORG field is missing
2. **What should be happening:**
Org field should always appear!
3. **Steps to recreate:**
Create a new wallet and connect to Coordinape and try to create a new circle
4. **Screenshots/GIF/Video recording**
![image](https://user-images.githubusercontent.com/82838264/197620094-70b54601-7534-4cc8-bf0e-28d9d27623a9.png)

PR #1560, opened 2022-10-24
Fix circle create form for new users

crabsinger:
Fix new user create circle form
Resolves https://github.com/coordinape/coordinape/issues/1558

The else clause when organizations are not present was missing the input field

## Testing
Create a circle as a brand new user with brand new address - should work
Create circle as existing user - should work


CryptoGraffe (approved):
tested and works

PR #1574, opened 2022-10-26
remove avatar old paths

### LINKED ISSUE #1531: Remove use of ui-avatars.com and AVATAR_PLACEHOLDER
See [domain.ts](https://github.com/coordinape/coordinape/blob/main/src/utils/domain.ts#L52:L54)

1. We have a placeholder image in the code that we should no longer use.
2. We have code that makes requests to ui-avatars.com to generate fallback images for avatars when there is no image set. This was a quick & handy solution at the time, but now we have [a better way to do it (in src/ui/Avatar)](https://github.com/coordinape/coordinape/blob/main/src/ui/Avatar/Avatar.tsx#L145) that doesn't depend upon an external server.

- Remove the placeholder.jpg image
- Remove `AVATAR_PLACEHOLDER`
- Remove `getInitialsUrl` 
- Remove `getAvatarPath` and rename `getAvatarPathWithoutPlaceholder` to `getAvatarPath`
- Remove `getCircleAvatar`
- Change all the code downstream so it still works
  - i.e.: Show an object's avatar image if it exists, or its initials according to [`getInitialFromName`](https://github.com/coordinape/coordinape/blob/main/src/utils/domain.ts#L101), using src/ui/Avatar.

yahiamarzouk:
## Motivation and Context

in "domain.ts"

Removed the placeholder.jpg image
Removed AVATAR_PLACEHOLDER
Removed getInitialsUrl
Removed getAvatarPath and rename getAvatarPathWithoutPlaceholder to getAvatarPath
Removed getCircleAvatar

removing AVATAR_PLACEHOLDER will make an Error in AMForceGraph.tsx 
## Related Issue

Fixes #1531


src/utils/domain.ts
```diff
@@ -47,27 +45,8 @@ export const AUTO_OPEN_WALLET_DIALOG_PARAMS = '?open-wallet';
 export const APP_URL_OPEN_WALLET = `${APP_URL}${AUTO_OPEN_WALLET_DIALOG_PARAMS}`;
 export const APP_PATH_CREATE_CIRCLE = `/new-circle`;
 export const APP_URL_CREATE_CIRCLE = `${APP_URL}${APP_PATH_CREATE_CIRCLE}`;
-export const AVATAR_PLACEHOLDER = '/imgs/avatar/placeholder.jpg';
```
crabsinger, line 50: yes, this is correct.

src/ui/Avatar/Avatar.tsx
```diff
@@ -3,10 +3,7 @@ import React from 'react';
 import * as AvatarPrimitive from '@radix-ui/react-avatar';
 
 import { styled, CSS } from '../../stitches.config';
-import {
```
crabsinger, line 6: Can you also create a Storybook story for Avatar that shows how this component looks both with a logo supplied, and without one.

crabsinger (commented):
@yahiamarzouk this looks off to a great start!

src/ui/Avatar/Avatar.tsx
yahiamarzouk: I still don't understand the storybook idea but I will get into it soon enough, maybe tomorrow

src/utils/domain.ts
```diff
@@ -102,13 +75,8 @@ export const getInitialFromName = (name: string) => {
   // Returns PR if name is "Peter Edwards Roman"
   // Returns PR if name is "Preben"
 };
+export const getAvatarPathWithFallback = (avatar?: string, name?: string) => {
```
karelianpie, line 78: This function needs to return a `string` but it's returning `string | undefined`. [`useImageUploader()`](https://github.com/coordinape/coordinape/pull/1574/files#diff-8fe7cd1d94726f02888d91e0873c8e24ba50cd7daf274077e798cfed738fa4a2R28) expects a `string`, but it can potentially get an `undefined` and that's why you're getting

> Type 'undefined' is not assignable to type 'string'.ts(2345)

crabsinger:
@yahiamarzouk I'm converting this to a draft until its fully working, tests are passing, and you are ready for a final review.

src/pages/CreateCirclePage/CreateCirclePage.tsx
```diff
@@ -41,6 +40,7 @@ export const SummonCirclePage = () => {
     avatar: string;
     avatarRaw: File | null;
   }>({
+    // i don't know who to remove this getCircleAvatar in line 44
```
levity, line 43: i think this can just be `null`. if you look at where `logoData.avatar` is used, it's used in `<Avatar>` on line 194, so if you just set the `name` property on that, it should work fine

src/components/CircleWithLogo.tsx
```diff
@@ -36,10 +31,7 @@ const CircleWithLogo = ({
             size="xs"
             margin="none"
             css={{ mr: '$sm' }}
-            path={getCircleAvatar({
-              avatar: orgLogo,
-              circleName: orgName,
-            })}
+            path={logo}
```
levity, line 34:
```suggestion
            path={orgLogo}
```

src/components/OrgLogoUpload.tsx
```diff
@@ -180,6 +180,7 @@ export const OrgLogoUpload = ({
             )}
           </Box>
         )}
+        {/* input error  */}
```
levity, line 183:
```suggestion
```

levity (approved):

Issue #1395, opened 2022-09-26
Move names to profiles


crabsinger:
This feature moves a user's `name` field from being a db field attached to `users` (and being different per each circle) to having one single name per address.

The following changes are required to implement this:

Phase 1 - DB / App Changes
* add column `name` on `profiles` table.
* Use `profiles.name` instead of `users.name` across the site -- fall back to `users.name` only when `profiles.name` is not set
* If an address has only one name (either in only one circle, or has the same name across all circles) automatically copy this user's name into their new `profile.name`
* If an address has multiple names: do not set any profile name for them, and prompt them to complete their profile info on next login.

Phase 2 - Updated Username Functionality
* Remove ability for circle admin to change a users name (remove from settings page, leave for add member as initial value)
* Add "change name" option to profile edit page

Phase 3 - Rollout and comms
* [Optional] think about notifying users with different names about this new options

Phase 4 - clean up:
* drop column `name` on `users` table once no longer in use and rollout was successful

----

### Detailed to-do sequence (work in progress)

- [x] Resolve conflicts (users with same name and different addresses)
  - [x] Sample users
  - [x] Actual users
    - there are over 800 name conflicts: ```select name, count(distinct address) ac from users group by name having count(distinct address) > 1 order by ac desc```
    - [x] create a migration for this: just append a number to each name
- [ ] Set `profiles.name` site-wide
  - due to the above, once this query is changed, the only profiles without names should now be ones with the same address and different names
  - we don't want to automatically pick one of these names for the address because it could cause information leakage between circles ("I don't want Circle A to know I'm named Foo in Circle B")
  -   [ ] Migration to enforce no empty profile manes and no nominees without profiles (run 1 month after profile name is implemented.)
      - Remove `name` from nominees and Users tables
  - [x] Remove Functionality to edit profile name from members table
  - [ ] Only fetch profile name and not user.name
 
```
UPDATE profiles
SET name = users.name
FROM users
WHERE users.address IN (
  SELECT users.address 
  FROM users 
  GROUP BY users.address 
  HAVING COUNT(distinct LOWER(users.name)) = 1
) 
AND users.address = profiles.address
```

----

Note to Coordinape staff: see [this Notion page](https://www.notion.so/coordinape/2022-12-06-Dev-f11c01039cb14915ba61a1feed9ad0ef) for more context

reeserj:
Do we want to enforce no duplicates here too , so I can't change my name to Crabsinger and get some of your give?

[...]

PR #1584, opened 2022-10-27
Add name column to profiles table

sshmm:
## Motivation and Context

Add name column to profiles tables as part of #1395 

## Description

name column is added to profiles tables
name is unique and nullable which I think should be discussed

## Test and Deployment Plan

 we need to run this sql statement in order to copy name values from users table to profiles
it will check if the address corresponding to one name before copying
users with multiple names in different circles will have null as name in profiles table

`UPDATE profiles`
`SET name = users.name`
`FROM users`
`WHERE users.address IN (
    SELECT users.address
    FROM users
    GROUP BY users.address
) AND users.address = profiles.address 
`

it can be run in data tab SQL section in hasura
image

I tested it by changing some user names in some of the circles and checked that these will remain null and the other will have the correct value in profiles.name


zashton:
@sshmm maybe we could make the name comparisons case-insensitive
`HAVING COUNT(distinct LOWER(users.name)) = 1 )`

sshmm:
> LOWER(users.name)
thanks I tested it. I will update the description

crabsinger (approved):

PR #1585, opened 2022-10-27
Claim Tokens Button on MobileHeader

### LINKED ISSUE #1488: Add Claim Distributions to Mobile Layout
**Summary:**

Users can't claim distributions on mobile because there's no button for it
Users should be able to claim distributions via a claim button in the hamburger menu.
3. **Steps to recreate:**
Log in on mobile with funds available to claim.

yahiamarzouk:
## Motivation and Context

Added the Claim tokens Button on mobile UI

Screen Shot 2022-10-27 at 8 12 19 PM

## Related Issue

Fixes #1488


src/components/MainLayout/MobileHeader.tsx
```diff
@@ -155,6 +156,20 @@ export const MobileHeader = ({
                   </Link>
                 )}
               </Box>
+              <Suspense fallback={null}>
+                {isFeatureEnabled('vaults') && (
```
crabsinger, line 160: @yahiamarzouk I think we should follow the logic from the [main header](https://github.com/coordinape/coordinape/blob/main/src/components/MainLayout/MainHeader.tsx#L39) and only show this button if there are claims, not always:

```suggestion
          {isFeatureEnabled('vaults') && showClaimsButton && (
```

crabsinger (changes requested):

levity (changes requested):
can you please move the button so that it's at the top of the menu? and make sure the left edge of the button is aligned with the other items in the list

wingfeatherwave:
@levity @yahiamarzouk I"m updating the mobile menu to not look like trash, which will fix the alignment issues described.  I will merge this code into my branch to avoid the conflicts.

PR #1587, opened 2022-10-27
Keyboard friendly navigation in Coordinape

### LINKED ISSUE #1432: Keyboard friendly allocations notes in the give beta build
### What's happening:
Currently navigating on the allocation screen is not keyboard friendly.
### What should be happening:
When you open the allocation screen your cursor should start on the allocation amount, and tab should take you to the notes, and tab again should take you to the next user

The tab flow should be:

1. Allocation amount
2. Note
3. Next Member button
4. Repeat

wingfeatherwave:
resolves #1432

* Lots of navigation updates so you can use `tab` to get around the site.  
* in some components, like our popover menus, you can use `return` to open them up, `escape` to close
* in allocations, when you closing a drawer you will keep your tab focus


wingfeatherwave:
why did my commit history get duplicated when I rebased main locally?

reeserj:
@wingfeatherwave When I create a new contribution from the Keyboard and tab to the close button and hit enter tabbing again i have to go back through the header to get to a new contribution. 

Is is possible to have it pop back out on the add contribution button from there?

src/components/MainLayout/OverviewMenu.tsx
```diff
@@ -182,7 +196,14 @@ export const OverviewMenu = ({
                     <Link
                       key={circle.id}
                       type="menu"
-                      onClick={() => closeAndGo(paths.history(circle.id))}
+                      // onClick={() => closeAndGo(paths.history(circle.id))}
```
crabsinger, line 199: comment remove

src/pages/GivePage/EpochStatementDrawer.tsx
```diff
@@ -200,6 +200,8 @@ export const EpochStatementDrawer = ({
           value={statement}
           onChange={e => statementChanged(e.target.value)}
           placeholder="Summarize your Contributions"
+          // eslint-disable-next-line jsx-a11y/no-autofocus
```
crabsinger, line 203: ? what's this for?

crabsinger: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-autofocus.md

crabsinger: hmm ok

src/pages/GivePage/MyGiveRow.tsx
```diff
@@ -28,11 +31,34 @@ export const MyGiveRow = ({
...
   optOutOpen: boolean;
   statementCompelete: boolean;
+  selected: boolean;
 }) => {
+  const newRef = useRef<HTMLButtonElement>(null);
+  const [lastSelected, setLastSelected] = useState<boolean>(false);
+
+  useEffect(() => {
+    if (selected) {
+      // eslint-disable-next-line no-console
```
crabsinger, line 41: remove consoles...

src/pages/GivePage/index.tsx
```diff
@@ -713,6 +697,10 @@ const AllocateContents = ({
         contributionCount={
           myMember?.contributions_aggregate?.aggregate?.count ?? 0
         }
+        // selected={true}
```
crabsinger, line 700: remove

src/pages/GivePage/index.tsx
```diff
@@ -818,7 +744,8 @@ const AllocateContents = ({
           paddingTop: 0,
           overflowY: 'scroll',
         }}
-        onClose={() => {
+        onOpenChange={() => {
+          // giveRowFocus()
```
crabsinger, line 748: remove

src/ui/Modal.tsx
```diff
@@ -61,7 +67,6 @@ const Title = styled(Dialog.Title, {
 type ModalProps = {
   children: ReactNode;
   title?: string;
-  onClose: () => void;
```
crabsinger, line 64: should we make `onOpenChange` required? Or is it only needed to overwrite normal functionality?

src/pages/GivePage/EpochStatementDrawer.tsx
wingfeatherwave: This is to ignore the accessibility warning about hijacking focus.  in this case we are returning the user's focus to the drawer trigger they clicked, which is the ideal a11y behavior

src/ui/Modal.tsx
wingfeatherwave: hrm.   maybe we should.  but technically if we're just using a radix trigger to manage the open state, then there is no need for open or onOpenChange

src/components/MyAvatarMenu/MyAvatarMenu.tsx
```diff
@@ -44,9 +44,18 @@ export const MyAvatarMenu = () => {
       <Hidden smDown>
         <Popover open={mouseEnterPopover}>
           <PopoverTrigger
+            tabIndex={-1}
             css={{ outline: 'none' }}
-            asChild
             ref={triggerRef}
+            onKeyDown={e => {
+              if (e.key === 'Enter') {
```
CryptoGraffe, line 51: This condition is different than the one in OverviewMenu. maybe a lil helper function that is like:
```
const onEnter = (f: func(e: uhhSomething): void) => {
 if (e.key === 'Enter' || e.key === ' ') {
   f(e);
 }
}
```

src/pages/GivePage/index.tsx
```diff
@@ -748,69 +735,7 @@ const AllocateContents = ({
...
-                  left: 0,
-                  zIndex: 2,
-                }}
-              ></Box>
-            </Box>
-          </Flex>
-        </Box>
-      )}
-
+      {}
```
CryptoGraffe, line 738: Whatsup with this change? This deletes the example of how to add a contributor, which was in the design

src/ui/Modal.tsx
```diff
@@ -88,22 +92,12 @@ export const Modal = ({
       onOpenChange={onOpenChange}
     >
       <Dialog.Portal>
-        <Overlay onClick={onClose} />
```
CryptoGraffe, line 91: I like overly onclick close, why go bye bye?

wingfeatherwave: it still does.  just letting radix methods handle it via onOpenChange

src/components/MyAvatarMenu/MyAvatarMenu.tsx
wingfeatherwave: good idea.  will add.

src/pages/GivePage/index.tsx
wingfeatherwave: will replace.

CryptoGraffe (approved):
Approved with the caveat that we need a new ticket that fixes the clearing out the give allocator, wing and I discussed the details here

PR #1593, opened 2022-10-29
bugfix: Contributions doesn't re-render when data changes

crabsinger:
There was an edge case when if you switch between one selected Contribution to another selected Contribution BEFORE the mutation finished saving, the useMemo would not be invoked, and the page would show old state even though the save succeeded. We change from that pattern to use a dataUpdatedAt memo pattern that properly updates when new data is fetched.

### DIFF
--- a/src/pages/ContributionsPage/ContributionsPage.tsx
+++ b/src/pages/ContributionsPage/ContributionsPage.tsx
@@ -278,7 +278,7 @@ const ContributionsPage = () => {
       returnData.epochs = createLinkedArray([currentEpoch, ...data.epochs]);
 
     return returnData;
-  }, [data?.epochs.length, data?.contributions.length, dataUpdatedAt]);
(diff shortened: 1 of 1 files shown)


src/pages/ContributionsPage/ContributionsPage.tsx
```diff
@@ -280,11 +278,7 @@ const ContributionsPage = () => {
       returnData.epochs = createLinkedArray([currentEpoch, ...data.epochs]);
 
     return returnData;
-  }, [
-    data?.epochs.length,
-    data?.contributions.length,
-    updateStatus === 'success' && isFetching === false,
-  ]);
+  }, [data?.epochs.length, data?.contributions.length, dataUpdatedAt]);
```
crabsinger, line 281: can we just use `dataUpdatedAt` and not the other two deps?

topocount (approved):

Issue #1570, opened 2022-10-25
Setup Storybook stories for all components in src/ui


crabsinger:
We're moving towards broader implementation of [Storybook](https://storybook.js.org/), and as part of this work we need to add all our current components into Storybook via stories.

Request access to our Chromatic Storybook build: https://www.chromatic.com/builds?appId=633dcabf2dd22d5956bffa53

End Result is that every component in [src/ui ](https://github.com/coordinape/coordinape/tree/main/src/ui)should have a story for it, which includes all optional states and errors.

This task can be worked on all at once, or piece by piece. I'm creating only one issue to track this, but if someone wants to tackle just one of these, just comment on the task saying which tasks they are taking on.

The currently created stories are:

```
./src/ui/SvgIcon.stories.tsx
./src/ui/ToggleButton.stories.tsx
./src/ui/Tooltip/Tooltip.stories.tsx
./src/ui/Panel/Panel.stories.tsx
./src/ui/TextField/TextField.stories.tsx
./src/ui/TextArea/TextArea.stories.tsx
./src/ui/Button/Button.stories.tsx
./src/ui/IconButton/IconButton.stories.tsx
./src/ui/Text/Text.stories.tsx
./src/ui/Select/Select.stories.tsx
./src/ui/Modal.stories.tsx
./src/pages/DefaultPage/DefaultPage.stories.tsx
```

Stories to be created are:
- [ ] RadioGroup/RadioGroup.tsx
- [ ] BackButton.tsx
- [ ] Box/Box.tsx
- [ ] Popover.tsx
- [ ] Checkbox/CheckBox.tsx
- [ ] Flex.tsx
- [ ] CenteredBox.tsx
- [ ] FormLabel/FormLabel.tsx
- [ ] Image/Image.tsx
- [ ] layouts.tsx
- [ ] Link/Link.tsx
- [ ] Table/Table.tsx
- [ ] Collapsible/Collapsible.tsx
- [ ] Divider/Divider.tsx
- [ ] Avatar/Avatar.tsx
- [ ] HR/HR.tsx
- [ ] HintButton.tsx
- [ ] CircleWithLogo.tsx
- [ ] Icon/Icon.tsx
- [ ] Modal.tsx
- [ ] Flex/Flex.tsx
- [ ] Form.tsx

karelianpie:
I can build these stories, will start from top to bottom on that list.

Will create a PR per story

yahiamarzouk:
[...]

PR #1595, opened 2022-10-30
feat: Add `BackButton` story

karelianpie:
## Motivation and Context

## Description

* Implement a story for the `BackButton.tsx` component;
* Move the `BackButton` component to its own folder and refactor imports.

## Test and Deployment Plan

Screenshot 2022-10-30 at 13 14 43

## Reviewers

@crabsinger 

## Related Issue


src/ui/BackButton/BackButton.stories.tsx
```diff
@@ -0,0 +1,16 @@
+import { ComponentStory, ComponentMeta } from '@storybook/react';
+import { withDesign } from 'storybook-addon-designs';
```
crabsinger, line 2: Let's leave out `withDesign` unless we're actually using it.

src/ui/BackButton/BackButton.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+import BackButtonComponent from './BackButton';
+
+export default {
+  title: 'Design System/Components/BackButton',
```
crabsinger, line 7: no longer need the title

src/ui/BackButton/BackButton.stories.tsx
```diff
@@ -0,0 +1,16 @@
...
+  component: BackButtonComponent,
+  decorators: [withDesign, Story => <Story />],
```
crabsinger, line 9: I don't think we need any decorators if we're not using withDesign, right?

crabsinger (changes requested):
@karelianpie thanks for cranking out all these stories:

Seems my comments will be same across all of them:

- let's not use `titles` since these can be inferred now
- lets not use `withDesign` unless were actively pointing to a Figma file.

src/ui/BackButton/BackButton.stories.tsx
karelianpie: Sounds good, removed with https://github.com/coordinape/coordinape/pull/1595/commits/3b733921529d2f7f73f9b810ede7160aaeab6525

src/ui/BackButton/BackButton.stories.tsx
karelianpie: Removed with https://github.com/coordinape/coordinape/pull/1595/commits/3b733921529d2f7f73f9b810ede7160aaeab6525

src/ui/BackButton/BackButton.stories.tsx
karelianpie: You're right! Removed with https://github.com/coordinape/coordinape/pull/1595/commits/3b733921529d2f7f73f9b810ede7160aaeab6525

wingfeatherwave (approved):

PR #1596, opened 2022-10-30
Add Form Story

yahiamarzouk:
## Motivation and Context

 adding the FormComponant story to our Coordinape StoryBook.

## Related Issue

#1570


src/ui/Form/Form.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+    css: {
+      table: {
+        disable: true,
+      },
+    },
+    content: {
+      table: {
+        disable: true,
+      },
+    },
```
karelianpie, line 23: `content` is not a prop for the `Form`, you should disable `as` and `ref` instead

src/ui/Form/Form.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+      dark: '(prefers-color-scheme: dark)',
+      display: 'flex',
+      flexDirection: 'column',
+    }}
+  >
+    <input placeholder="UserName"></input>
+    <input placeholder="Age"></input>
+    <input placeholder="Email"></input>
+    <button type="submit"> Submit </button>
+  </FormComponent>
```
karelianpie, line 39: Screenshot 2022-10-31 at 12 25 54

Could we make the `Form` story look a bit more like a form :)

karelianpie (changes requested):
Good first draft, I'd suggest we use the same components used throughout the app for the `Form` story, i.e. `Panel`, `Text`, `TextField`, `Button`, etc. rather than just `html` elements.

The idea with storybook is to have a starting point when using a component, so I'd just browse the codebase and find how this `Form` is being used across the app and then create a more realistic story for it

src/ui/Form/Form.stories.tsx
yahiamarzouk: what is 'as'

karelianpie: > what is 'as'

It's just a prop

src/ui/Form/Form.stories.tsx
```diff
@@ -0,0 +1,46 @@
...
+      },
+    },
+    css: {
+      table: {
+        disable: true,
+      },
+    },
+    as: {
+      disable: true,
+    },
```
karelianpie, line 21:
```suggestion
    as: {
      table: {
        disable: true,
      },
    },
```
Please refer to the docs [here](https://storybook.js.org/docs/ember/writing-docs/doc-block-argstable#customizing) on how to disable these controls. The same needs to be done for the `ref` prop below, the point of this setting is to hide these controls as they are not needed:

Design_System___Components___Form_-_Form_⋅_Storybook

yahiamarzouk: thank you chef, always learning from you 👍

yahiamarzouk: Screenshot 2022-10-31 at 4 18 41 PM

wingfeatherwave:
@yahiamarzouk I don't really see a need to have this as a story in storybook.  The component itself exists only to pull in stitches.  I'm going to close this story.  Sorry for having not seen this sooner.

karelianpie:
> @yahiamarzouk I don't really see a need to have this as a story in storybook. The component itself exists only to pull in stitches. I'm going to close this story. Sorry for having not seen this sooner.

I think we should have this story as the component is being used across the app, it does not really matter it's just a wrapper component IMO -- people building a form would like to have an example of a `Form` being used. This was the place for that.

PR #1602, opened 2022-10-31
Switch to use a buffering -> scheduled debounce enqueue pattern

crabsinger:
This pattern follows the Give page's pattern rather than using UseEffect to conditionally enqueue saves

### DIFF
--- a/src/pages/ContributionsPage/ContributionsPage.tsx
+++ b/src/pages/ContributionsPage/ContributionsPage.tsx
@@ -144,7 +144,10 @@ const ContributionsPage = () => {
         descriptionField.value
       );
     }
-  }, [saveState, currentContribution?.contribution.id]);
+  }, [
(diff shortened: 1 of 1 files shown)


src/pages/ContributionsPage/ContributionsPage.tsx
```diff
@@ -127,12 +129,27 @@ const ContributionsPage = () => {
...
+  useEffect(() => {
+    // once we become buffering, we need to schedule
+    // this protection of state change in useEffect allows us to fire this only once
+    // so requests don't stack up
+    if (saveState[currentContribution?.contribution.id] == 'buffering') {
+      updateSaveStateForContribution(
+        currentContribution?.contribution.id,
+        'scheduled'
+      );
+      // Should we cancel this too????
```
crabsinger, line 143: @topocount - we left this comment in here...

PR #1603, opened 2022-10-31
fixup/refresh give setting on data change

crabsinger:
- bugfix: setGifts whenever data changes
- Remove never used give and note fields from member object


CryptoGraffe (approved):
Paired on this with crab and confident about the fix

levity:
could we add a test to the suite that would prevent a regression? something that fails when this fix is reverted and passes when it's applied

Issue #1545, opened 2022-10-20
Add Contributions options to the API


reeserj:
**Problem**
What problem is this feature addressing?
The new contributions feature needs to have API read and write access
**Solution**
What's your proposed solution to the problem?
Add the ability to read and write to the contributions data in the API

**Stakeholders**
Who needs to provide input? Who might be impacted? Who needs to approve?
@META-DREAMER

PR #1609, opened 2022-11-01
Update datamodel for read/create contributions API

META-DREAMER:
## Motivation and Context

Allows 3rd party integrators / api users to programatically read and create contributions for a given circle.

## Description

Added columns and appropriate permissions in Hasura for the `api-user` role to create and read contributions. Also added a new column to contributions that tracks which API key was used to create the contribution (if any).

## Test and Deployment Plan

Ensure permissions are correctly configured and contributions created with a given API key are marked as such.

## Reviewers

@topocount @zashton 

## Related Issue

#1545

### DIFF
--- a/hasura/migrations/default/1667278140823_contributions_api/down.sql
+++ b/hasura/migrations/default/1667278140823_contributions_api/down.sql
@@ -1,0 +1,8 @@
+
+alter table "public"."contributions" drop constraint "contributions_created_with_api_key_hash_fkey";
+
+alter table "public"."contributions" drop column "created_with_api_key_hash";
+
+alter table "public"."circle_api_keys" drop column "create_contributions";
+
+alter table "public"."circle_api_keys" drop column "read_contributions";
--- a/hasura/metadata/databases/default/tables/public_circle_api_keys.yaml
+++ b/hasura/metadata/databases/default/tables/public_circle_api_keys.yaml
@@ -1,6 +1,6 @@
 table:
-  schema: public
   name: circle_api_keys
+  schema: public
 object_relationships:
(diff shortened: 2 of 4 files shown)


hasura/migrations/default/1667278140823_contributions_api/down.sql
```diff
@@ -0,0 +1,20 @@
+
+alter table "public"."contributions" drop constraint "contributions_created_with_api_key_hash_fkey";
+
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
```
crabsinger, line 5: can we remove the commented out stuff.

hasura/metadata/databases/default/tables/public_vaults.yaml
```diff
@@ -1,6 +1,6 @@
 table:
-  schema: public
   name: vaults
+  schema: public
```
crabsinger, line 3: Seems a bunch of these changes are just re-ordering the fields... would be preferable to keep that change out of here, or do it first in a separate PR.

META-DREAMER: seems like it was done automatically by Hasura CLI. Related issue: https://github.com/hasura/graphql-engine/issues/8743

will revert in this PR for now

topocount: @crabsinger I think there are a bunch of hasura files that need to be regenerated since the upgrades

topocount (approved):
Thanks @META-DREAMER! We should make this step a standard part of augmenting API I suppose.

PR #1617, opened 2022-11-01
feat: Add `Table` story

karelianpie:
## Motivation and Context

## Description

* Implement a basic story for the `Table` component.

## Test and Deployment Plan

Screenshot 2022-11-01 at 20 46 22

## Reviewers

@crabsinger 

## Related Issue

### DIFF
--- a/src/ui/Table/Table.stories.tsx
+++ b/src/ui/Table/Table.stories.tsx
@@ -1,30 +1,10 @@
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 import faker from 'faker';
-import { withDesign } from 'storybook-addon-designs';
 
 import { Table as TableComponent } from './Table';
 
 export default {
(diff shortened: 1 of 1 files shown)


wingfeatherwave (approved):

levity:
@karelianpie also in this story, we can't randomize the data each time, because Chromatic detects it as a change. You should be able to set the seed for faker so it always generates the same data for the story.

karelianpie:
> @karelianpie also in this story, we can't randomize the data each time, because Chromatic detects it as a change. You should be able to set the seed for faker so it always generates the same data for the story.

Fixed with https://github.com/coordinape/coordinape/pull/1638/commits/b5f89606018a5e24d973d9e197b0ad0f687a2f8b (PR https://github.com/coordinape/coordinape/pull/1638)

PR #1619, opened 2022-11-02
Copy over created date times during end epoch

zashton:
## Motivation and Context

Fixes final allocation datetimes to be inaccurate


CryptoGraffe (approved):

topocount:
Thanks zashton!

PR #1636, opened 2022-11-06
feat: Add `HintButton` story

karelianpie:
## Motivation and Context

## Description

* Implement a basic story for the `HintButton` component.

## Test and Deployment Plan

Screenshot 2022-11-06 at 14 27 57

## Reviewers

@crabsinger 

## Related Issue


levity (changes requested):
Some context about why/where to use this instead of some other Button would be useful

karelianpie:
> Some context about why/where to use this instead of some other Button would be useful

Good question, I was actually not sure why this was needed, @CryptoGraffe since you've created this component do you know?

wingfeatherwave:
Will replace `<HintButton` with `<Link target="_blank" rel="noreferrer"...` and remove it.  Closing PR.

PR #1638, opened 2022-11-07
fix: Chromatic stories

karelianpie:
## Motivation and Context

Randomising data does not work well with Chromatic as it detects it as a change

## Description

Serve always the same assets and data for the `Image` and `Table` stories.

## Test and Deployment Plan

Ran storybook locally and observed always the same image being served as well as the same data for the table

Screenshot 2022-11-07 at 9 40 50
Screenshot 2022-11-07 at 9 50 40

## Reviewers

@levity


levity (approved):

levity:
thanks for the quick fix @karelianpie! note my change to `.seed`; it only needs to be called once per session

PR #1640, opened 2022-11-07
Add profile name to users queries

### LINKED ISSUE #1395: Move names to profiles

sshmm:
## Motivation and Context

Partially Resolves #1395 

## Description

1- fetch profile.name with upon fetching user.name
2- use profile.name in the UI if available. if not use user.name

## Test and Deployment Plan

1- Checked UI to make sure that all user names appears as they should be like in members table and give page..etc

### DIFF
--- a/api-lib/event_triggers/createCircleCRM.ts
+++ b/api-lib/event_triggers/createCircleCRM.ts
@@ -46,6 +46,9 @@ async function handler(req: VercelRequest, res: VercelResponse) {
           },
           {
             name: true,
+            profile: {
+              name: true,
+            },
           },
         ],
       },
@@ -59,7 +62,7 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     await insertCRMRecord({
       CircleName: name,
       ContactInfo: contact ?? '',
-      MemberName: users[0].name,
+      MemberName: users[0].profile.name ?? users[0].name,
       OrgName: organizations_by_pk.name,
       // TODO: there is no way to input these yet
       ContactMethod: '',
--- a/api-lib/event_triggers/utils/handleVouchMsg.ts
+++ b/api-lib/event_triggers/utils/handleVouchMsg.ts
@@ -31,7 +31,9 @@ export default async function handleVouchMsg(
 
   // announce the vouching
   await sendSocialMessage({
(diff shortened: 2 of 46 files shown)


sshmm:
replaced by #1694 due to local git issues after rebasing

PR #1642, opened 2022-11-08
Mobile menu updates

wingfeatherwave:
* Mobile menu updates
* no more default material red as primary in places, most noticeable in our loading animation

Screenshot 2022-11-07 at 3 59 20 PM
Screenshot 2022-11-07 at 4 04 51 PM
Screenshot 2022-11-08 at 2 25 06 PM


levity (approved):
looking sharp!

i'm assuming you intentionally decided against showing the green button when you have tokens to claim?

levity (changes requested):
can you create a Storybook story for the mobile header so we can get some visual checks in CI from Chromatic?

wingfeatherwave:
> i'm assuming you intentionally decided against showing the green button when you have tokens to claim?

In the avatar (profile) menu we show a persistent `/claims` link.  We also show that green button if `(query.data?.claims_aggregate.aggregate?.count || 0) > 0;`. . Which is correct?

levity:
I figured, since we added the green button for visibility when you have
tokens to claim, we could do the same, with a green button at the top of
the mobile menu. But I'm open to alternatives if you think it's too
duplicative when all the links are shown together in a single column in the
mobile menu.

wingfeatherwave:
> can you create a Storybook story for the mobile menu so we can get some visual checks in CI from Chromatic?

Out of curiosity, what would having it as a story give us that is better than looking at the app?  Asking since this is really a compound component, and not something that will be reused. 

Also, I'm planning on unifying the mobile/desktop nav in our nav redesign.

levity:
Maybe I'm stretching the purpose of Storybook here, but now that we have Chromatic set up, it seems to me it'd be great to put lots of our compound components in there just for the purpose of doing automated checks

levity (dismissed):

levity (approved):

PR #1658, opened 2022-11-14
Add Pgives schema

zashton:
## Motivation and Context

Updated schema in preparation for pgive stuff.

## Description

Additional table and columns in epoch table to index pgive data.


levity:
I think that for this sort of data, which isn't used for the core app, it would be better to have it in a separate table that has a foreign key into epochs. This will be more flexible for future changes for permissions, performance optimizations, etc.

levity:
Also, can you create a separate PR just with the changes to the schema & metadata that aren't related to your changes? Thanks!

zashton:
> I think that for this sort of data, which isn't used for the core app, it would be better to have it in a separate table that has a foreign key into epochs. This will be more flexible for future changes for permissions, performance optimizations, etc.

have created a seperate table and added relationship to epochs

hasura/metadata/databases/default/tables/public_epoches.yaml
```diff
@@ -16,6 +16,13 @@ object_relationships:
         remote_table:
           name: circles
           schema: public
+  - name: epoch_pgive_data
```
levity, line 19: can you change the relation name to just `pgive_data`? no need to repeat `epoch_` when you're already starting from an epoch

levity (changes requested):
Just the one rename; all else looks good

levity (approved):

PR #1674, opened 2022-11-16
Allow hiding pending gives during active epoch

### LINKED ISSUE #1523: Hidden GIVE Setting
Admin will be able to hide member give amounts during an active epoch. The map is not available to the user until post-epoch. If GIVE is hidden. GIVE is revealed after Epoch ends. This does not affect data such as epoch description, member notes, and GIVE amounts. This data will be stored and released after the Epoch ends.

sshmm:
## Motivation and Context

resolves #1523

## Description

1- add checkbox to hide pending gives in circle settings
2- update updateCircle mutation to include this option
3- hide Map page, gives popover if that option is enabled during active epoch from regular users only

## Test and Deployment Plan
Go to circle settings and enable this option
image

1- Regular user with this option enabled during active epoch
image
2- Regular users after epoch ending
image
3- Admins are not affected
image


wingfeatherwave (approved):
Tested UI, looks good.  Thanks @sshmm

src/components/MainLayout/CircleNav.tsx
```diff
@@ -38,13 +41,13 @@ export const CircleNav = () => {
         [paths.epoch(circleId), paths.team(circleId), paths.give(circleId)],
       ],
       [paths.members(circleId), 'Members'],
-      [paths.map(circleId), 'Map'],
     ];
-
+    if (showGiveInfoQuery.data?.showGiveInfo || myUser.isCircleAdmin)
```
levity, line 45: @sshmm @wingfeatherwave this prevents a non-admin user from accessing the map at all if `showGiveInfo` is disabled, but that means they can't even look at past epochs... is there a reason i shouldn't revert this?

PR #1680, opened 2022-11-17
Bugfix console warnings nested `button` and `anchor`

wingfeatherwave:
* Console log was getting filled with improper nesting warnings.  Fixed.
* Remove Popover anchor, as it's not actually needed.  I thought it would be helpful in making positioning easier, but the new version of Radix popover makes the positioning easier without `PopoverAnchor`, which can be used if you want the popover to appear someplace different than the trigger.

### DIFF
--- a/src/pages/GivePage/MyGiveRow.tsx
+++ b/src/pages/GivePage/MyGiveRow.tsx
@@ -3,7 +3,7 @@ import { useEffect, useRef, useState } from 'react';
 import { ApeInfoTooltip } from '../../components';
 import { Check, X } from '../../icons/__generated';
 import { IMyUser } from '../../types';
-import { Box, Button, Flex, Text, ToggleButton } from 'ui';
+import { Box, Flex, Text, ToggleButton } from 'ui';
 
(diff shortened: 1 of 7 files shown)


levity (approved):

src/pages/GivePage/MyGiveRow.tsx
```diff
@@ -47,10 +47,18 @@ export const MyGiveRow = ({
     }
   }, [selected, lastSelected]);
   return (
-    <Button
+    <Box
+      tabIndex={0}
       onClick={openEpochStatement}
+      onKeyDown={e => {
```
levity, line 53: Did you mean to have these changes in this pr? I notice you didn't mention them in the description

wingfeatherwave: Yes.  In order to not have nested buttons, I needed to convert this to a div.  Sadly, onClick is not captured on a div when using keyboard... so needed the onkeydown.  and using onkeydown had some strange behavior for the drawer it opened, hence the preventDefault and stopPropagation.  Kind of gnarly

PR #1684, opened 2022-11-21
Minor UI update: inline edit text ui update

wingfeatherwave:
Less obtrusive inline edit text pattern

* add support for svg icon links with nice alignment

Screenshot 2022-11-21 at 12 06 56 PM


levity (approved):
Nice! Looks great

Issue #1653, opened 2022-11-09
Expanded Allocation View


reeserj:
**Problem**
Some Circles want to be able to see what everyone has worked on at a glance without having to click into every person

**Solution** 
An alternative view of the allocation rows. 

https://www.figma.com/file/L4oZwYVUdpgacZtwipGaYe/Coordinape-App?node-id=26659%3A85937&t=PyQ4jhfgV7bLCItj-1

PR #1689, opened 2022-11-23
Grid View for Allocation page

wingfeatherwave:
* Row / Grid css switcher for Give page (default row)
* Tag style updates
* Had to move some tags around in the row view to keep the code sane
* Markdown Headers are now all rendered as the same size `medium` so not to pollute our app with crazy font sizes
* mobile and keyboard nav updates
* add an icon
* adjust md breakpoint

Fixes [#1653](https://github.com/coordinape/coordinape/issues/1653)

Screenshot 2022-11-23 at 3 07 59 PM


levity (approved):
Haven't tested this but the code changes all look pretty tame & unobjectionable. GivePage/index should be split into smaller files but that's outside the scope here ofc.

PR #1544, opened 2022-10-20
Support Multi Chain Login

crabsinger:
## Motivation and Context

Since we use a SIWE pattern to generate a sign-in token and then use that for all further API requests, we can support SIWE across many EVM chains. This PR introduces this pattern and adds support for a number of other EVM chains as supported Login Methods.

- Added a restriction on Claims, Vaults that require you to be on Mainnet or Contract Supported vaults, and also prompt a switch to Mainnet in the browser as well.
- Added a on networkChange callback that reloads the page - this is recommended for avoiding weird cross-network issues.

For now, we only show the Select Network UI if the feature flag is enabled. So we'll launch this but not show the prompt to switch networks.

On a supported Netowrk Login Modal will look like:
Screen Shot 2022-11-28 at 4 19 10 PM

On an unsupported Network login Modal will look like:
Screen Shot 2022-11-28 at 4 19 19 PM

Screen Shot 2022-11-02 at 6 08 11 PM

Follow up tasks:
- [ ] Improve UI and use Uniswap like Chain Selector
- [ ] Write login API test and contract only network hooks test

PR #1690, opened 2022-11-24
refactor to prepare for email login

levity:
## Motivation and Context

We're looking at integrating Magic Connect or Web3Auth to allow users to sign into Coordinape with email addresses. There's some cleanup of our auth code that needs to happen first to clear the way for integration.

## Description

Change most components that are using web3-react directly to use a shim hook instead, so that they can be agnostic to whether the provider, address, chain id, etc. are coming from web3-react or a different source.

## Test and Deployment Plan

Test suite should still pass. App behavior should be the same--these changes should be entirely non-user-facing.

### Notes For Reviewers

Most of the action is in `hooks/useWeb3React`, which contains the shim hook described above, plus a provider component that combines the web3-react provider & the context provider used for tracking where we are in the authentication process. It introduces [zustand](https://github.com/pmndrs/zustand) into the project as a lightweight & easy-to-use alternative to Recoil (no context provider needed!) for sharing state.

This PR also continues the trend away from organizing code by its type (e.g. hooks, components, mutations, selectors), and toward organizing it by its purpose, e.g.: 
  - `hooks/useApiBase -> features/auth/useFinishAuth`
  - `lib/gql/mutations, recoilState/app -> features/auth/useWalletAuth`
  - `services/api -> features/auth/login, features/auth/token`
  
 Please see `features/README.md` and suggest any changes you think would make the rationale clearer.

#### TODO

- [ ] add some tests
- [x] auto-reconnect before rendering WalletAuthModal at all?
- [x] after #1544 lands: move more code into `features/auth`
  - `hooks/login`
  - `WalletAuthModal`
  - `RequireAuth`
  - `services/api`
  - `useAuthToken`


levity:
* main
  * **PR #1544** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1544" target="_blank">Graphite</a> 
    * **PR #1690** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1690" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1690?utm_source=stack-comment).

src/features/auth/RequireAuth.tsx
```diff
@@ -0,0 +1,51 @@
...
+    if (authStep !== 'connect' && !web3Context.active) {
+      setAuthStep('connect');
+      return;
+    }
+
+    if (authStep === 'connect' && web3Context.active) {
+      setAuthStep('sign');
+      finishAuth({ web3Context, authTokens })
+        .then(success => {
+          if (!success) {
```
crabsinger, line 28: nit: assuming success is a bool, then this is fine... but i generally prefer to case on the affirmative, and else on the not affirmative for a bit more readability, and safer handling if there are every multiple cases that could be returned.

```suggestion
          if (success) {
                      setAuthStep('done');
                      }
          else { 
          web3Context.deactivate();
          }
```

src/features/auth/WalletAuthModal.tsx
```diff
@@ -121,19 +124,29 @@ export const WalletAuthModal = ({ open }: { open: boolean }) => {
...
 
   useEffect(() => {
-    if (
-      // safe to refer to window here because we are in useEffect -g
-      window.location.search === AUTO_OPEN_WALLET_DIALOG_PARAMS ||
-      walletAuth.connectorName
-    ) {
-      walletAuth.connectorName && activate(walletAuth.connectorName);
-    }
+    if (walletAuth.connectorName && !web3Context.active)
```
crabsinger, line 127: Do you understand the case when this would be true `if (walletAuth.connectorName)` ?

src/features/auth/WalletAuthModal.tsx
```diff
@@ -121,19 +124,29 @@ export const WalletAuthModal = ({ open }: { open: boolean }) => {
...
+      activate(walletAuth.connectorName);
   }, []);
 
+  const inject = async () => {
```
crabsinger, line 131: I guess I'd prefer to leave this out of `main` when merging this PR, and add it later if there isnt any functionaltiy here yet...

src/features/auth/token.ts
```diff
@@ -0,0 +1,10 @@
+let token: string | undefined;
```
crabsinger, line 1: is this a global, or whats the context of where this variable is now stored?

src/features/auth/useWalletAuth.ts
```diff
@@ -0,0 +1,44 @@
...
+  const { logoutUser } = await client.mutate(
+    { logoutUser: { id: true } },
+    { operationName: 'logout' }
+  );
+  return !!logoutUser?.id;
+};
+
+const updateToken = ({ address, authTokens }: IAuth) => {
+  const token = address && authTokens[address];
+  if (!token && getAuthToken(false)) logout();
```
crabsinger, line 20: what's the point of calling `getAuthToken(false)` here? Just to see if token on line 19 is unavailable, but the token could still be stored and retrievable from getAuthToken() ?

src/hooks/useWeb3React.tsx
```diff
@@ -0,0 +1,126 @@
...
+    ...context,
+    library,
+    active: context.active || !!address,
+    chainId: context.chainId || chainId,
+    account: context.account || address,
+    setProvider,
+    deactivate: context.active ? context.deactivate : () => setProvider(),
+  };
+
+  useEffect(() => {
```
crabsinger, line 89: remove this? or wrap in a debug flag? seems only useful to help remind someone hwo to debug but not wroth keeping in `main`

crabsinger (commented):
@levity looks good - i really like the organizational aspects of moving all this to src/features.

i had a few comments/questions on some things.

I'm a bit hesitant to just add zustand with only a minimal amount of switching to use it, as i dont want to be in a place where we're 50% zustand and 50% recoil and that's actually worse then just fully using recoil... so i guess id be curious to talk more about a plan to remove recoil and switch to zustand, and feel confident its going to happen and has approproate energy behidn it before adding zustand as another state management tool alongside recoil.

src/features/auth/WalletAuthModal.tsx
```diff
@@ -101,6 +114,9 @@ export const WalletAuthModal = ({ open }: { open: boolean }) => {
     }
 
     try {
+      // after this succeeds, consumers of useWeb3React will re-run and
+      // web3Context.active will be true
+      setConnectMessage('');
```
crabsinger, line 119: This actually doesn't work for us... since it hides the message for WalletConnect and Coinbase Wallet....

src/features/auth/WalletAuthModal.tsx
levity: it's true when there's saved walletAuth data, which happens after successful login in `useFinishAuth` [here](https://github.com/coordinape/coordinape/pull/1690/files#diff-9364f778b5d58002aee011b7f30b5acf78b68de864caa758194fb49ea89b8bccR70)

src/features/auth/WalletAuthModal.tsx
levity: i understand. it's just here in this case because i was going back and forth between this and #1704 a lot and it felt right to keep some placeholder here for when i start working on #1705 more

src/features/auth/token.ts
levity: this is not exactly a global, in the sense that you can only access it from outside this file with the exported setter & getter, but it is a singleton. this isn't a change from previous behavior, because the old ApiService was a singleton.

src/features/auth/useWalletAuth.ts
levity: Hmm, I'm carrying over this logic from [before](https://github.com/coordinape/coordinape/pull/1690/files#diff-fe13ed674a354105a9e96354ec08b53830b3a8f3653f90c4eb1311f6a464e2a1L52), but I think it's true that this could just be `if (!token) logout()`, because i can't think of a case where you would be saving a blank token and there isn't already a token stored. Will test removing this

src/hooks/useWeb3React.tsx
levity: Will remove

crabsinger (dismissed):
cool - looks good to merge after the comments are addressed.

levity:
@crabsinger I feel good about adding zustand now. It's better-documented & better-designed than Recoil, I've seen it prove its worth on past projects, and it also happens to have a [tiny bundle size](https://bundlephobia.com/package/zustand@4.1.4).

Also, I can't imagine ever getting to a point where we're 50% recoil, 50% zustand, because the appropriate replacement for Recoil in most of our app is not zustand, but no shared state at all.

src/features/auth/useWalletAuth.ts
levity: Oh, that's right, this is needed when you load the page for the first time and the user's not logged in yet. `updateToken` is called with blank auth info because the `rWalletAuth` atom is being initialized.

PR #1704, opened 2022-11-29
Magic Connect support

levity:
Support email login with Magic Connect

- [x] add Magic to login dialog
- [x] add REACT_APP_MAGIC_API_KEY to vercel env
- [x] support networks other than mainnet
- [x] fix login persistence
- [x] set `connectorName` in `rWalletAuth` for Magic
  - maybe this will allow using `activate` for reconnecting instead of a separate component
- [x] test making vault transactions
  - [x] when trying to create a vault, clicking on the Magic tx dialog closes the vault create form drawer

for later:
- [ ] track profiles that are created with email login
- [ ] login flow is not quite right yet: sometimes the app proceeds, but then after a few seconds, the Magic prompt appears


levity:
* main
  * **PR #1544** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1544" target="_blank">Graphite</a> 
    * **PR #1690** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1690" target="_blank">Graphite</a> 
      * **PR #1705** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1705" target="_blank">Graphite</a> 
      * **PR #1704** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1704" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1704?utm_source=stack-comment).

src/features/auth/useWalletAuth.ts
```diff
@@ -21,12 +32,28 @@ const updateToken = ({ address, authTokens }: IAuth) => {
...
 };
 
+const AUTH_STORAGE_KEY = 'capeAuth';
+
+const saveAuth = (auth: IAuth) =>
+  localStorage.setItem(AUTH_STORAGE_KEY, JSON.stringify(auth));
+const getSavedAuth = (): IAuth => {
+  try {
+    const auth = localStorage.getItem(AUTH_STORAGE_KEY);
+    assert(auth);
```
crabsinger, line 42: assert is just here to verify the fetched local storage auth data object is not an empty {} ?

JSON.parse would otherwise accept and return it even though there's no auth data there?

crabsinger (commented):
code looks generally good - didn't test because I don't have a magic auth key. Should we all get our own dev key and our own account? Or can we provision a dev key and pass around/

src/features/auth/useWalletAuth.ts
levity: oh, this is leftover from when i removed assertDef a few months ago: 7fe85945

looking at this more closely now, it's true we could drop the assert, since the try/catch will handle all the weird cases.

PR #1698, opened 2022-11-28
 Bug: Dework tasks not showing up on Allocations card view

MooSeMoos:
## Motivation and Context

resolves #1691 

## Description

-Allow Dework and wonder tasks to show up on allocations card view 
-the start date of the epoch GiveDrawer and EpochStatementDrawer is the end of the previous epoch like ContributionsPage

## Test and Deployment Plan
Test 1 
open circle --> create epoch --> add tasks from (wonderverse Or Dework) connection --> check Allocations card view
test 2
open circle --> go to contributions page -->check the letest contribution and date  --> go to Allocations card view and check

## Screenshots (if appropriate):
test 1
Screenshot from 2022-11-28 17-38-52
Screenshot from 2022-11-28 17-38-43
Screenshot from 2022-11-28 17-37-02

test 2
Screenshot from 2022-11-28 17-51-45
Screenshot from 2022-11-28 17-51-35
Screenshot from 2022-11-28 17-51-26
Screenshot from 2022-11-28 17-51-17

## Reviewers

@levity 
@sshmm 

## Related Issue

#1691

PR #1709, opened 2022-11-29
bugfix: stop querying contributions from time 0

topocount:
Contribution providers do not want to support multi-year queries due to
adversarial concerns such as scraping. Our contribution hook would fall
back to epoch timestamp zero if no start date for the query was
supplied. This is now fixed and the query falls back to a month before
the current epochs start date if no prior epoch exists.

I've also tightened up the API to make the dates required params so that
no hidden fallbacks are possible. This is at the expense of the legacy
code in the ProfileCard, which utilizes recoil and irresponsibly tries
to render a result even if no epoch is currently active.

It'd be preferable to pass in a required epoch prop or two that the
component can rely on, but given that this is a legacy view, it's not
worth the maintenance. The solution, then is nominally unreachable hack
that would yield a minimally impactful query for integrated
contributions

### DIFF
--- a/src/components/ProfileCard/ProfileCard.tsx
+++ b/src/components/ProfileCard/ProfileCard.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 
+import { DateTime } from 'luxon';
+
 import { makeStyles, Button } from '@material-ui/core';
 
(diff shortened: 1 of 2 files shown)


levity (approved):

levity:
Can you please check out #1698 to make sure this fix will apply correctly to that? Thanks 🙏

PR #1712, opened 2022-11-29
Cleanup auth flow

crabsinger:
This fixes a React console error around updating an unmounted component.

image

In short, we call setConnectMessage after the modal is closed throwing this error.

Also removes the ?open_wallet functionality and logic which is unused.

### DIFF
--- a/src/components/WalletAuthModal.tsx
+++ b/src/components/WalletAuthModal.tsx
@@ -12,9 +12,7 @@ import { Box, Button, Text, Modal, Flex, HR, Select } from '../ui';
 import { EConnectorNames, WALLET_ICONS } from 'config/constants';
 import isFeatureEnabled from 'config/features';
 import { useApeSnackbar } from 'hooks';
-import { useWalletAuth } from 'recoilState/app';
 import { connectors } from 'utils/connectors';
-import { AUTO_OPEN_WALLET_DIALOG_PARAMS } from 'utils/domain';
 import { switchNetwork } from 'utils/provider';
 
 export const WalletAuthModal = ({ open }: { open: boolean }) => {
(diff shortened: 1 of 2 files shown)


src/components/WalletAuthModal.tsx
```diff
@@ -116,20 +114,8 @@ export const WalletAuthModal = ({ open }: { open: boolean }) => {
...
-    setConnectMessage('');
   };
 
-  useEffect(() => {
-    if (
-      // safe to refer to window here because we are in useEffect -g
-      window.location.search === AUTO_OPEN_WALLET_DIALOG_PARAMS ||
-      walletAuth.connectorName
-    ) {
-      walletAuth.connectorName && activate(walletAuth.connectorName);
```
levity, line 129: if i'm not mistaken, this is what allows the app to automatically reconnect to the last logged-in wallet, so it shouldn't be removed.

we're thinking alike though -- i took the auto-open stuff out in my branch here: d338a48

crabsinger: ok, yeah, you're right. just looked into this more and on page reload, or fresh visit, the walletAuth.connectorName is present as `injected` but the walletAuth is not activated. so we need this, otherwise you have to click "metamask" every page reload.

src/components/WalletAuthModal.tsx
```diff
@@ -101,6 +98,7 @@ export const WalletAuthModal = ({ open }: { open: boolean }) => {
     }
 
     try {
+      setConnectMessage('');
```
crabsinger, line 101: @levity this line is still needed, because the WalletAuthModal closes and then this tries to update the text on it (unmounted now) and throws a console error. I think its easiest to just bring this over to your PR rather than keep this one open since you moved all this code.

crabsinger: pushed these fixes onto your branch: https://github.com/coordinape/coordinape/pull/1690/commits/6b781f07c101b32708e3bb0665907900538804ed, so I'll go ahead and clsoe this PR. @levity

PR #1713, opened 2022-11-30
link to create a real circle in the sample org alert bar

CryptoGraffe:
## Motivation and Context
Users have expressed that they don't know how to "leave" or "progress out of" the sample circle experience.

## Description

New "Create real Circle" button in the sample org header


CryptoGraffe:
Tweaking styling a bit w/ ivy before this ships

crabsinger (approved):

PR #1714, opened 2022-11-30
fix date range in give drawer not including contributions from before epoch started

z-br:
## Motivation and Context

Give Drawer contribution list was only inlcuding contributions that happened _during_ the epoch, instead of since the last epoch. Same for epoch statement drawer. @MooSeMoos fixed this in #1698 but I'm breaking it off into a separate simpler PR that is easier to test since we were hitting this in our give circle. My fix is also slightly different. 

## Description

Change the start time in query to be the end of previous epoch


levity:
#1698 just got merged, sorry for the extra work

PR #1717, opened 2022-11-30
Remove deprecated createUser handler

zashton:
## Motivation and Context

Removing the hasura createUser handler as createUser is no longer used anywhere in the FE (currently using createUsers instead) and I realised it seems to be broken right now while trying to write tests for it.


crabsinger (approved):
lgtm

PR #1720, opened 2022-12-01
Implement a Debug Logger

topocount:
## Motivation

Why implement a new logger class? In short, to improve dev experience
when it comes to:
- implementing nuanced, complex logic that might need to be extended or
  debugged further later
- easily muting logging in unit tests (by aggressively mocking this class)
- runtime debugging locally by using the DEBUG environment variable

This is all possible because the class uses the debug package
under the hood: https://github.com/debug-js/debug#debug

## Test Plan

unit tests are included and demonstrate how the logger can be used

add this code to `src/App.ts`:
```js
import { DebugLogger } from 'common-lib/log';
const logger = new DebugLogger('App');
logger.log('hello, world');
logger.error('derp');
```

Peek 2022-12-01 11-25

### DIFF
--- a/src/common-lib/log.ts
+++ b/src/common-lib/log.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-console */
 import debug from 'debug';
 
-export class Logger {
+export class DebugLogger {
   log: ReturnType<typeof debug>;
   error: ReturnType<typeof debug>;
   constructor(location: string) {
--- a/src/common-lib/log.ts
+++ b/src/common-lib/log.ts
@@ -1,0 +1,16 @@
+/* eslint-disable no-console */
+import debug from 'debug';
+
+export class Logger {
+  log: ReturnType<typeof debug>;
+  error: ReturnType<typeof debug>;
+  constructor(location: string) {
+    // set all locations under our `coordinape` namespace
+    const domain = 'coordinape:' + location;
+    this.log = debug(domain + ':log');
+    // log to stdout
+    this.log.log = console.log;
+    // print errors to stderr (debug's default)
+    this.error = debug(domain + ':error');
+  }
+}
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,9 @@ LOCAL_POSTGRES_USER=postgres
 LOCAL_POSTGRES_PASSWORD=postgrespassword
 LOCAL_POSTGRES_DATABASE=postgres
 
+# enable app debug and error logs
+DEBUG=coordinape*
+
 # Create a bot here: https://discord.com/developers/applications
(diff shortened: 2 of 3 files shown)


topocount:
* main
  * **PR #1720** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1720" target="_blank">Graphite</a>  👈
    * **PR #1721** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1721" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1720?utm_source=stack-comment).

topocount:
**Graphite Merge Job**

Current status: 🔄 Waiting for checks to pass

This pull request is currently waiting for CI and mergeability checks to pass before being merged.

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1720?utm_source=stack-comment).


Job Reference: ZUTgtAZ48xmFtp8HlPbJ

src/common-lib/log.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export class DebugLogger {
+  log: ReturnType<typeof debug>;
+  error: ReturnType<typeof debug>;
+  constructor(location: string) {
+    // set all locations under our `coordinape` namespace
+    const domain = 'coordinape:' + location;
+    this.log = debug(domain + ':log');
+    // log to stdout
+    this.log.log = console.log;
```
levity, line 12: hey, sorry i didn't get to reviewing this earlier. what's the purpose of this additional property? it's not sufficient to call `this.log` directly to log to console?

levity:
i can see that this would help standardize the log names we're using, which are currently all over the place, in addition to creating a site for more sophisticated behavior in the future. (though... [YAGNI](https://www.martinfowler.com/bliki/Yagni.html))

but since we're already using `debug` to log in a number of places in the code, i would have suggested that if we want to insert our own logic, we continue to conform to the `debug` API so it can be a more-or-less drop-in replacement. 

but having a mix between direct `debug` usage and this Logger might lead to confusion. for example, adding that ENV variable will also turn on logging for hardhat setup and lib/vaults/contracts.

src/common-lib/log.ts
levity: ah, found the explanation in https://github.com/debug-js/debug#output-streams

but that also says:
```js
log.log = console.log.bind(console); // don't forget to bind to console!
```

topocount: It worked without the bind step when I was testing this out. Initially had it, but then I figured out we didn't need it.

PR #1724, opened 2022-12-01
Feature/bulk user delete

### LINKED ISSUE #807: Bulk Deletion of Users
# Enable API Support for bulk deletion of users via Graphql API

This task should be done using hasura's action handler

The GQL mutation should accept an input of `circle_id` + an array of `addresses`

Validate that there are no duplicates of `address` among the array of `addresses` in the input. 
if there are any duplicates throw an 422 error and indicate the specific `addresses` that has issue.

To delete a user set the `deleted_at` column of the user to current timestamp
also delete records from `teammates` table where user is == `user_id` or == `team_mate_id`.
The event trigger will handle the cleanup of the accounting if the circle is in an active epoch. (Do test if this is working correctly)

for the response return a `success` boolean.

FilipeAleixo:
resolves #807


api/hasura/actions/_handlers/deleteUserBulk.ts
```diff
@@ -0,0 +1,131 @@
...
+        },
+        { id: true },
+      ],
+    },
+    {
+      operationName: 'deleteUserBulk_getCurrentEpoch',
+    }
+  );
+
+  if(!currentEpoch) {
```
zashton, line 42: Admin should be able to delete users even when there is no ongoing epoch

api/hasura/actions/_handlers/deleteUserBulk.ts
```diff
@@ -0,0 +1,131 @@
...
+      },
+      422
+    );
+  }
+
+  // Check that all the addresses exist
+  for (const address of addresses) {
+    const {
+      users: [existingUser],
+    } = await adminClient.query(
```
zashton, line 70: we can optimise this by combining it into 1 query request instead

api/hasura/actions/_handlers/deleteUserBulk.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (!existingUser) {
+      errorResponseWithStatusCode(res, { message: `User with address ${address} does not exist.` }, 422);
+      return;
+    }
+
+    existingUserIds.push(existingUser.id)
+  }
+
+  for (const userId of existingUserIds) {
+    await adminClient.mutate(
```
zashton, line 99: we could optimize by merging the mutate into 1 call instead and so if the mutation fails the whole transaction will rollback together

zashton (changes requested):
Thanks for the PR, highlighted some improvements. We would also need to add an unit test for this action handler . You could refer from`api-test/hasura/actions/handlers`

hasura/metadata/actions.graphql
```diff
@@ -1,89 +1,135 @@
 type Mutation {
-  adminUpdateUser(payload: AdminUpdateUserInput!): UserResponse
+  adminUpdateUser(
```
zashton, line 2: can it be because of your hasura-cli version difference hence the formatting changes ?

FilipeAleixo:
Thanks for the feedback! will make those changes

hasura/metadata/actions.graphql
FilipeAleixo: I'm not sure where these differences in formatting come from. I'm using `yarn hasura console` so it's using hasura-cli 2.10.1. Hasura docker image is also the same. The only thing that might differ is my OS - I'm on Ubuntu 20.04

api/hasura/actions/_handlers/deleteUserBulk.ts
```diff
@@ -0,0 +1,142 @@
...
+    session_variables: sessionVariables,
+  } = composeHasuraActionRequestBody(deleteUserBulkInput).parse(req.body);
+
+  const { circle_id, addresses } = payload;
+  const existingUserIds : number[] = [];
+  const uniqueAddresses = [...new Set(addresses.map(a => a.toLowerCase()))];
+
+  // Check whether circle is in active epoch
+  const {
+    epochs: [currentEpoch],
```
zashton, line 24: think this query is also no longer needed anymore

FilipeAleixo: The circle doesn't need to be in an active epoch in order to delete the users?

api/hasura/actions/_handlers/deleteUserBulk.ts
FilipeAleixo: Done in https://github.com/coordinape/coordinape/pull/1724/commits/32611e6fb3b5e855f0eebc4d3835e8518e7f4f6e

api/hasura/actions/_handlers/deleteUserBulk.ts
FilipeAleixo: Done in https://github.com/coordinape/coordinape/pull/1724/commits/cde0b2926d95f14a72e4b763672381134b3a5557

FilipeAleixo:
> We would also need to add an unit test for this action handler . You could refer from`api-test/hasura/actions/handlers`

@zashton I'm currently trying to set up the test but on trying to run `yarn test:ci` I keep getting "Waiting for services on ports 3007, 8087, 8547...", even though all the services are running. Did you go through this issue before?

levity:
> > We would also need to add an unit test for this action handler . You could refer from`api-test/hasura/actions/handlers`
> 
> @zashton I'm currently trying to set up the test but on trying to run `yarn test:ci` I keep getting "Waiting for services on ports 3007, 8087, 8547...", even though all the services are running. Did you go through this issue before?

Do you see any errors appear on startup? (I know it's tricky because the web server clears the console-- there may be a way to prevent that in your terminal app; I know how to do that in macOS iTerm but not Windows)

You can also try running `yarn test:up` in one terminal and `yarn test:ci` in another terminal (or just `yarn test` for only the Jest tests)

api/hasura/actions/_handlers/deleteUserBulk.ts
zashton: nope, you can delete users anytime

zashton:
> > We would also need to add an unit test for this action handler . You could refer from`api-test/hasura/actions/handlers`
> 
> @zashton I'm currently trying to set up the test but on trying to run `yarn test:ci` I keep getting "Waiting for services on ports 3007, 8087, 8547...", even though all the services are running. Did you go through this issue before?

@FilipeAleixo have you tried @levity 's suggestion, did it work for you ? 
It usually works for me else a stop & restart of `yarn test:ci`would fix it.

FilipeAleixo:
@zashton I've managed to go past the waiting for the services problem I had, but currently I'm being faced with this error for the test I needed to base myself off of, which I don't know how to solve
image

crabsinger:
> @zashton I've managed to go past the waiting for the services problem I had, but currently I'm being faced with this error for the test I needed to base myself off of, which I don't know how to solve image

Hmm - i saw this same error today.

Is it still happenign to you? Maybe try a full restart of your system and see if it persists? Seems like soemthing weird and itnermittent when it happened to me.

Can you also rebase this branch off main to update it @FilipeAleixo ?

crabsinger:
@sshmm when you have a chance, you could take over this PR, get it rebased and up to date, and then resolve any remaining errors. I think this was mostly done before @FilipeAleixo ran into some issues.

sshmm:
> Thanks for the PR, highlighted some improvements. We would also need to add an unit test for this action handler . You could refer from`api-test/hasura/actions/handlers`

done with some other improvements. please re review

CryptoGraffe:
@zashton are you able to re-review this?

zashton (approved):

zashton:
@sshmm looks good covered all flows, thanks

PR #1624, opened 2022-11-03
pricePerShare in Hasura remote schema

levity:
The API handler was split out ahead of #1520 to resolve difficulties encountered during deployment. Now that those are resolved, it's all in a single branch again. See #1520 for more context.

PR #1725, opened 2022-12-01
use price_per_share from remote schema

levity:
In #1520 and #1624 we added a Hasura remote schema so the frontend can get pricePerShare values for vaults in a Hasura query instead of having to directly query the blockchain. This PR changes the distributions & claims code to use that.

### DIFF
--- a/src/common-lib/distributions.ts
+++ b/src/common-lib/distributions.ts
@@ -1,26 +1,4 @@
-import { FixedNumber } from 'ethers';
 import { formatUnits } from 'ethers/lib/utils';
-
-const calc = ({
-  address,
-  decimals,
-  fixedGifts,
-  pricePerShare,
-}: {
-  address: string;
-  decimals: number;
-  fixedGifts?: Record<string, string>;
-  pricePerShare?: FixedNumber;
-}) => {
-  return {
-    fixedPaymentAmt: getUserClaimedFixedPaymentAmt({
-      decimals,
-      fixedGifts,
-      address,
-    }),
-    pricePerShare: pricePerShare?.toUnsafeFloat() || 1,
-  };
-};
 
 const getUserClaimedFixedPaymentAmt = ({
   decimals,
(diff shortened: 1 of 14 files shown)


levity:
* main
  * **PR #1725** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1725" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1725?utm_source=stack-comment).

src/common-lib/distributions.ts
```diff
@@ -62,28 +41,23 @@ export const claimsUnwrappedAmount = ({
...
   if (circleDistClaimAmount) {
     const claimAmt = circleDistClaimAmount || 0;
-    const { fixedPaymentAmt, pricePerShare } = calc({
+    const fixedPaymentAmt = getUserClaimedFixedPaymentAmt({
       address,
       decimals: circleDistDecimals,
       fixedGifts: circleFixedGifts,
-      pricePerShare: circleDistPricePerShare,
     });
-    fixedPayment = fixedPaymentAmt * pricePerShare;
```
levity, line 71: @zashton i think i found a bug here? it was multiplying fixed payment amount by circle distribution's price per share. can you confirm that this was a bug, & if so, what do you think the impact was on users?

src/pages/DistributionsPage/DistributionsPage.test.tsx
```diff
@@ -72,26 +70,14 @@ jest.mock('./queries', () => ({
   useSubmitDistribution: jest.fn(),
 }));
 
-let queryClient: QueryClient;
```
levity, line 75: removed this code because it didn't fix the sporadic Jest crash that happens when running this test file

src/common-lib/distributions.ts
zashton: hmmm should be correct, there are 4 cases here
combined dist (fixed + circle dist)
circle dist only .
fixed dist only
circle + fixed dist seperate distributions

that line handles the case of combined distribution where fixed+circle dist is using the same vault hence pricePerShare is the same. So if it is only a circle dist `fixedPaymentAmt` will be 0 because `circleFixedGifts` is empty and so `fixedPayment` will be also 0

the following `if (fixedGifts)` to handle when the distribution is only a fixed payment distribution .

Your update should have the same end result as the original code

levity:
This page breaks with a GraphQL error due to the remote server: https://coordinape-5zu4rbki1-coordinape.vercel.app/circles/2610/distributions/7385

src/common-lib/distributions.ts
levity: ah, of course, thanks for clearing that up!

PR #1733, opened 2022-12-05
feat: Hedgey Integration, adding locked_token_distributions table and…

bark4mark:
## Motivation and Context

We require a new table (locked_token_distributions) and the ability to update an existing circle_integration.

## Description

This is the backend changes for the Hedgey distribution

## Test and Deployment Plan

A developer should apply the migrations and ensure they don't fail, there is no functionality affected by this change.

## Reviewers

@levity @crabsinger


hasura/metadata/databases/default/tables/public_locked_token_distributions.yaml
```diff
@@ -0,0 +1,53 @@
...
+insert_permissions:
+  - role: user
+    permission:
+      check:
+        epoch:
+          circle:
+            users:
+              profile:
+                id:
+                  _eq: X-Hasura-User-Id
```
levity, line 17: The insert and update checks should also validate that the user is a circle admin (role = 1)

hasura/migrations/default/1670234780227_create_table_public_locked_token_distributions/up.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE "public"."locked_token_distributions" ("id" bigserial NOT NULL, "epoch_id" bigint NOT NULL, "gift_amount" numeric NOT NULL, "tx_hash" varchar, "distribution_json" jsonb NOT NULL, "distributed_by" varchar NOT NULL, PRIMARY KEY ("id") , FOREIGN KEY ("epoch_id") REFERENCES "public"."epoches"("id") ON UPDATE cascade ON DELETE cascade);
```
levity, line 1: What values does `distributed_by` contain?

levity:
Thanks for making these small PRs!

hasura/migrations/default/1670234780227_create_table_public_locked_token_distributions/up.sql
bark4mark: This is the address that kicked off the distribution, we will display it in the locked token distribution table.

bark4mark:
I have answered the two questions, thank you :grin:

hasura/migrations/default/1670234780227_create_table_public_locked_token_distributions/up.sql
levity: could we make it into a foreign key linking to `profiles.id`? or do you want the flexibility of being able to add an address that doesn't correspond to a Coordinape user?

levity (commented):
Followup question: https://github.com/coordinape/coordinape/pull/1733#discussion_r1042734490

hasura/migrations/default/1670234780227_create_table_public_locked_token_distributions/up.sql
bark4mark: @levity yes that makes sense, I've updated the database to have a foreign key and added the rule to have it populated by the session variable (similar to how the distributions table is configured).

levity:
I'm going to merge this to staging first to make sure the migrations take; if all goes well, we can push to prod tomorrow

PR #1735, opened 2022-12-05
Create profiles with names mutations

### LINKED ISSUE #1395: Move names to profiles

sshmm:
## Motivation and Context

resolves #1395

## Description

1- Add a modal to ask users to enter a profile name that will be used in all circles
2- Change createusers mtations 'manual-address-entry' to check if the names exists in profiles and nominees
3- create users and profiles as follows:
 - if the user has a profile without a name the entered name will be the profile name
 - if the user doesn't have a profile, create a profile with that name
 - if the name exists for other addresses inform the user
 4- create Nominee only if the nominee name is not used by other nominees or users

### Test 1
1- Log in to coordinape
expected:
a modal should appear to enter user name and only accepts unique names

### Test 2
1- Create a nominee with a unique name
2- try to add user with that name
expected:
you should get an error about the situation

### Test 3
1- complete the vouching for the nominee in Test 2
expected: a profile with the nominee name should be added

### Test 4
1- create a user with a new name
expected: a profile with the user name should be added

### Test 5
1- create a user with a name in profiles table
expected: you should get an error

image

Screenshot from 2022-12-05 15-40-05

Screenshot from 2022-12-04 23-25-34

inform Admins of ignored names entry
image


api/hasura/actions/_handlers/createUsers.ts
```diff
@@ -135,6 +137,162 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
+          names.length > 1
+            ? 'names used by other coordinape users'
+            : 'a name used by another coordinape user'
+        }: ${names}`,
+      },
+      422
+    );
+  }
+
+  const { profiles } = await adminClient.query(
```
zashton, line 179: You probably can save on the query call if you merge the create & update profile mutation into an upsert like this. Think this should work the same.

```
import { profiles_constraint, profiles_update_column } from '../../../../api-lib/gql/__generated__/zeus';

const createProfilesMutations = adminClient.mutate({
    insert_profiles:[
      {
        objects: newUsers.map( user => {
          return {
            address: user.address.toLowerCase(),
            name: user.name,
          }
        }),
        on_conflict: {
          constraint: profiles_constraint.profiles_address_key,
          update_columns: [profiles_update_column.name],
          where: {
            name: {_is_null: true}
          }
        }
      },
      {
        returning: {
          id: true,
        }
      }
    ]
  },
  {
    operationName: 'createUsers_createProfiles',
  });
  ```

File: api/hasura/actions/_handlers/createUserMutation.ts [hunk 2]
```diff
@@ -38,6 +42,63 @@ export async function createUserMutation(
...
+  if (!addressProfile && nameProfile) {
+    throw new UnprocessableError(
+      'This user name is used by another coordinape user'
+    );
+  }
+
+  let createProfileMutation = null;
+
+  if (!addressProfile && !nameProfile) {
+    createProfileMutation = await adminClient.mutate(
```
zashton, line 59: its possible to do it as an upsert as well here

src/lib/zod/index.ts
```diff
@@ -122,6 +122,12 @@ export const generateApiKeyInputSchema = z
   })
   .strict();
 
+export const updateProfileNameSchemaInput = z
+  .object({
+    name: z.string(),
```
zashton, line 127: should we impose a character limit ?

zashton (changes requested):
nice work on the profile name changes, left a few suggestions.

api/hasura/actions/_handlers/createUsers.ts
```diff
@@ -135,6 +137,162 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
+  //check if names are used by other coordinape users
+  const { profiles: existingNames } = await adminClient.query(
+    {
+      profiles: [
+        {
+          where: {
+            _or: newUsers.map(user => {
+              return {
+                _and: [
+                  { name: { _eq: user.name } },
```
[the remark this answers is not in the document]
sshmm: I am using citext for names so ilike is not necessary

levity (commented):
Oops! I left a bunch of comments 2 days ago but never submitted the review. If anything I've said since then conflicts with a comment here, ignore the comment. :)

zashton (dismissed):
Went through all the test cases, works with no issues. 👍

zashton:
@sshmm note that all the e2e cypress tests are failing

zashton (dismissed):

levity (approved):
nice work! sorry it took so long to get thru review 😅

PR #1740, opened 2022-12-07
Create Hasura Role for the discord bot

topocount:
This role formalizes the access the discord bot will have to the db,
     since it's administering an adjacent schema to persist state.

Specifically, the discord bot is now able to find all circles a user is
an admin of after they link their discord account to their coordinape
account. This is needed for the circle-linking flow.

PR #1741, opened 2022-12-07
Create table to store api tokens for the discord bot

topocount:
The schema and permissions support the following flow:

1. The discord bot receives a request to administer a circle via the
   discord bot. This might entail the following:
   - adding or removing users via discord roles
   - assigning the circle to a channel
   - configure circle notifications in discord
2. the bot inserts a row in the discord.circle_api_tokens table
   containing a `circle_id` and the `channel_snowflake` and begins listening for the token to be added
   to that row via a subscription
3. the bot generates a link that will open the coordinape app on the
   discord route with the param `id` representing the row id in the
   circle_api_tokens table.
4. The user opens the link in a browser and authenticates or is
   authenticated with the address that has admin perms on the circle.
5. In the frontend, a circle api token is generated and passed into an
   action handler along with the circle_id provided, and the id of the
   token.
6. The handler attempts to update the row created by the bot with the
   new id. If no row exists with a null token field and the
   update fails, the key is deleted and an error is shown to the user
   in a toast on redirect.
7. If the update is successful, the user is notified the circle is
   linked to the channel in both the app in a toast and in discord
   because the subscription returns affirmatively.

We want to ensure the links generated to complete the flow are
single-use and do not cause unintended side-effects if the link is
accessed more than once. This flow guarantees that the links will be
single-use.

This schema and prescribed flow would not be possible for updating the
api token, so it's best to request all permissions the bot needs up
front, rather than having the permissions granted to the bot evolve
without compromising on the single-use aspect.

We could consider generating and passing around a single-use identifier
but the flow gets a bit more complex. This could also be added later
without much concern for additional lift. [YAGNI](https://builtin.com/software-engineering-perspectives/yagni) for now seems prudent
though

### Todos
- [ ] Update the Zeus schema

### DIFF
--- a/hasura/migrations/default/1670795162160_alter_table_discord_roles_circles_alter_column_role/up.sql
+++ b/hasura/migrations/default/1670795162160_alter_table_discord_roles_circles_alter_column_role/up.sql
@@ -1,0 +1,1 @@
+alter table "discord"."roles_circles" rename column "role" to "server_role";
--- a/api-lib/config.ts
+++ b/api-lib/config.ts
@@ -72,6 +72,10 @@ export const MIXPANEL_PROJECT_TOKEN: string = getEnvValue(
   ''
 );
 
+export const HASURA_DISCORD_SECRET: string = getEnvValue(
+  'HASURA_DISCORD_SECRET'
+);
+
 export const DISCORD_BOT_CLIENT_ID: string = getEnvValue(
   'DISCORD_BOT_CLIENT_ID',
   'no_token'
--- a/api-lib/gql/__generated__/zeus/const.ts
+++ b/api-lib/gql/__generated__/zeus/const.ts
@@ -1564,6 +1564,67 @@ export const AllTypesProps: Record<string, any> = {
     _neq: 'date',
     _nin: 'date',
   },
+  discord_circle_api_tokens_aggregate_fields: {
+    count: {
+      columns: 'discord_circle_api_tokens_select_column',
+    },
+  },
+  discord_circle_api_tokens_bool_exp: {
+    _and: 'discord_circle_api_tokens_bool_exp',
+    _not: 'discord_circle_api_tokens_bool_exp',
+    _or: 'discord_circle_api_tokens_bool_exp',
+    channel_snowflake: 'String_comparison_exp',
+    circle_id: 'bigint_comparison_exp',
+    created_at: 'timestamptz_comparison_exp',
+    id: 'bigint_comparison_exp',
+    token: 'String_comparison_exp',
+  },
+  discord_circle_api_tokens_constraint: true,
+  discord_circle_api_tokens_inc_input: {
+    circle_id: 'bigint',
+    id: 'bigint',
+  },
+  discord_circle_api_tokens_insert_input: {
+    circle_id: 'bigint',
+    created_at: 'timestamptz',
+    id: 'bigint',
+  },
+  discord_circle_api_tokens_on_conflict: {
+    constraint: 'discord_circle_api_tokens_constraint',
+    update_columns: 'discord_circle_api_tokens_update_column',
+    where: 'discord_circle_api_tokens_bool_exp',
+  },
+  discord_circle_api_tokens_order_by: {
+    channel_snowflake: 'order_by',
+    circle_id: 'order_by',
+    created_at: 'order_by',
+    id: 'order_by',
+    token: 'order_by',
+  },
+  discord_circle_api_tokens_pk_columns_input: {
+    id: 'bigint',
+  },
+  discord_circle_api_tokens_select_column: true,
+  discord_circle_api_tokens_set_input: {
+    circle_id: 'bigint',
+    created_at: 'timestamptz',
+    id: 'bigint',
+  },
+  discord_circle_api_tokens_stream_cursor_input: {
+    initial_value: 'discord_circle_api_tokens_stream_cursor_value_input',
+    ordering: 'cursor_ordering',
+  },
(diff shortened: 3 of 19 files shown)


topocount:
* main
  * **PR #1740** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1740" target="_blank">Graphite</a> 
    * **PR #1741** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1741" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1741?utm_source=stack-comment).

topocount:
TODO: update the new zeus schema

levity (dismissed):

karelianpie:
I'm not sure about the steps 2-6, I thought after out discussion that the flow would be

1. The bot generates a link that will open the coordinape app on the discord route with the params `circle_id` and `channel_snowflake`;
2. The user clicks on the link and is redirected from discord to the browser;
3. In the frontend, a circle api token is generated and passed into an action handler along with the circle_id provided, and the id of the token;
4. The handler attempts to insert a new row, if there's already a row with `circle_id` and `channel_snowflake` then an error message is displayed letting the user know that the circle is already linked to the channel. If there is no row with the given `circle_id` and `channel_snowflake` and the insert is successful, the user is notified the circle is linked to the channel in both the app in a toast and in discord.

I'm not sure why we need the subscription, can't this be done synchronously through a mutation?

I'm afraid that we will also need the user to authorise discord yet again in this step too. The reason being that we need the discord admin user snowflake as well, this should not be sent as a param otherwise anyone could try and impersonate a circle admin. Therefore, I believe we will need to have another user authorisation, once we have the user token we can make a request to the discord api to get their snowflake as we do for the user linking, once we have their snowflake we can initiate this flow because we can validate that the user making the request is indeed the discord user linked to the profile in coordinape that as admin permission to the circle.

> This schema and prescribed flow would not be possible for updating the
api token, so it's best to request all permissions the bot needs up
front, rather than having the permissions granted to the bot evolve
without compromising on the single-use aspect.

This is a bit of a concern as well, unless we know all permissions that we need upfront. What if new things are added to the circle, how would we update the tokens?

topocount:
@karelianpie the flow you describe above makes the generated links non-idempotent, and will generate another token each time the link is accessed in an authenticated context. We want the links to be strictly single-use.

> What if new things are added to the circle, how would we update the tokens?

If the discord bot doesnt have the perms, it can always request more. We'd just pass the existing row `id` into an update handler that deletes the exiting row in a database transaction that subsequently adds the new one. I'm trying to do you a favor here by making it plain that this concern is not in scope for your work :wink: but if you want to do it, I'm not going to protest.

My point was that this flow cannot be used to create new token unless the existing one is explicitly unlinked first. This is because the existing row would throw a constraint error, since circle_ids and tokens have a 1:1 relationship in this table. (to minimize ambiguity)

> I'm not sure why we need the subscription, can't this be done synchronously through a mutation?

the subscription is needed only for the bot to listen for a token to be added for a given circle. Apologies if my wording was confusing there.

> I'm afraid that we will also need the user to authorize discord yet again in this step too. The reason being that we need the discord admin user snowflake as well.

This won't be necessary, because as @levity suggested, the discord_bot role will be updated have access to user perm levels and can verify a linked coordinape user's circle role before proceeding to generate a link.

Please let me know if you have any further questions.

karelianpie (approved):

karelianpie:
Can this be merged @topocount?

karelianpie:
Ok, so in the discord front:

1. the bot can already request the user to select the circles they want to manage (we still need to get the real circles):

Screenshot 2022-12-20 at 4 01 57 pm

2. Create a new channel and role for the circle(s) selected:

Screenshot 2022-12-20 at 4 03 05 pm

3. In the new channel prompt the user to link the circle, once they click in Link it's ready to send the mutation with `circleId` and `channel_snowflake`:

Screenshot 2022-12-20 at 4 00 07 pm

Once this goes in we can and begin listening for the token to be added to that row via a subscription and continue with the flow

/cc @reeserj

hasura/migrations/default/1670795162160_alter_table_discord_roles_circles_alter_column_role/up.sql
```diff
@@ -0,0 +1 @@
+alter table "discord"."roles_circles" rename column "role" to "server_role";
```
crabsinger, line 1: Can we move this into the first migration where we create the table and remove the second `alter table` migration?

karelianpie: The first migration in this PR created a different table, `"discord"."circle_api_tokens"`, so we should keep this `alter table` as that table creation happened a while ago.

karelianpie:
> @karelianpie I can help get this merged. There was one remaining comment on the duplicate migrations and it needs to be rebased and cleaned up again. If you can make those changes I'll approve and merge. (If you need to push to a different branch because you don't have write to this branch just lmk.)

I do not have write access to this, but was able to rebase locally without any merge conflicts.

Assuming it's all good, let me know @crabsinger

karelianpie:
Let's close this and continue here https://github.com/coordinape/coordinape/pull/1788

I've added a bit more logic to handle the circle linking

karelianpie:
Closing as I've rebased with `main` and added a bit more logic to handle the circle linking here https://github.com/coordinape/coordinape/pull/1788

PR #1758, opened 2022-12-10
Move web3 event handlers into global

crabsinger:
We want the web3 event handlers to be setup globally, not just via the auth modal, so moving these into Web3React hook setup.

Additionally adding a reload on accountsChanged hook.

### DIFF
--- a/src/hooks/useWeb3React.tsx
+++ b/src/hooks/useWeb3React.tsx
@@ -1,3 +1,5 @@
+import { useEffect } from 'react';
+
 import type { JsonRpcProvider } from '@ethersproject/providers';
 import { Web3Provider } from '@ethersproject/providers';
 import {
@@ -62,6 +64,7 @@ export function Web3ReactProvider({
     <OriginalWeb3ReactProvider getLibrary={getLibrary}>
       {children}
       <MagicModalFixer />
+      <Web3EventHooks />
     </OriginalWeb3ReactProvider>
   );
 }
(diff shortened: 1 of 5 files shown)


src/hooks/useWeb3React.tsx
```diff
@@ -81,3 +84,30 @@ function getLibrary(provider: any): Web3Provider {
...
+        // When a Provider makes its initial connection, it emits a "network"
+        // event with a null oldNetwork along with the newNetwork. So, if the
+        // oldNetwork exists, it represents a changing network
+        if (oldNetwork) {
+          window.location.reload();
+        }
+      });
+
+      // Web3Provider doesn't work with accountsChanged
+      ethereum.on('accountsChanged', () => {
```
crabsinger, line 105: @topocount not sure why but using `provider.on` doesn't work here. Any idea why the Ethers provider wouldn't pass through this event? But using the plain object does work?

topocount: @crabsinger I'm not sure what

> using the plain object does work

means. Could you sketch the working example?

crabsinger: `ethereum.on('accountsChanged', ()` works
`provider.on('network',` does not work.

@topocount does that make sense? Wrapping the `window.ethereum` object in a Ethers Web3Provider breaks the functionality.

crabsinger: https://github.com/ethers-io/ethers.js/issues/1396#issuecomment-806380431 answer here!

levity (dismissed):

levity (approved):

PR #1768, opened 2022-12-15
Verify deleteCircle unit tests

zashton:
## Motivation and Context

to verify https://github.com/coordinape/coordinape/pull/1754/files#r1046564492


zashton:
@crabsinger 
verified the unit test is running the actual handler code, 
my guess is that if you modified the actual handler code you will need to restart `yarn test:up`.
`mockUserClient` is simulating making the api call from that specified user with his authToken.
image

PR #1775, opened 2022-12-20
Delete old give page

CryptoGraffe:
## Motivation and Context

Old give page is no longer linked or used, we can finally jettison it into the sun. This changes /givebeta to /give , deletes some recoil stuff, and changes the cypress test to use the new give page. 

## Description

Delete the page and some of the dependencies that are no longer used. 

## Test and Deployment Plan

Make sure nav and allocating work ok.


cypress/e2e/give.cy.ts
```diff
@@ -0,0 +1,80 @@
...
+      circleId = q.circles[0].id;
+    });
+  });
+  it('can login and visit give', () => {
+    cy.visit(`/circles/${circleId}/give`);
+    cy.login();
+    cy.url({ timeout: 120000 }).should('include', '/give');
+    cy.contains('thank your teammates').click();
+  });
+  // it('can select teammates', () => {
```
crabsinger, line 25: should we make this try the "+ Collaborator" thing when you hover on a row?

src/pages/GivePage/queries.ts
```diff
@@ -122,6 +122,141 @@ export type CircleAllocationText = Awaited<
...
+            user: {
+              circle_id: { _eq: selectedCircleId },
+              address: { _ilike: address },
+            },
+          },
+        },
+        {
+          team_mate_id: true,
+          teammate: {
+            // id: true,
```
crabsinger, line 221: why commented? remove if not needed?

crabsinger (dismissed):
@CryptoGraffe two small nits - otherwise looks good

cypress/e2e/give.cy.ts
CryptoGraffe: did it

src/pages/GivePage/queries.ts
CryptoGraffe: this whole query can be deleted , nice catch

levity (approved):

PR #1769, opened 2022-12-15
Hedgey Locked token distribution frontend

bark4mark:
## Motivation and Context

This is a follow on to #1733 and #1734 and adds the frontend to allow users to distribute locked tokens using Hedgey.

## Description

- Circle Admin
  - Added Hedgey Integration

- Distribution page
  - Added Hedgey option to gift circle
  - Hedgey options should only show if hedgey is enabled and no locked token distribution has been done for this epoch
  - If a locked token distribution has been done, the etherscan button should be shown and should take you to the batchmint transaction (in a new tab)

- Claims Page
  - Added a new table for locked token distribution
  - Table has individual gift amount, a link to the transaciton and a link to the hedgey portfolio (should open in new tab)

## Test and Deployment Plan

There is a meeting booked for 15/12/22 to talk through testing and QA.

## Screenshots (if appropriate):

Circle Integration

Screenshot from 2022-12-15 10-12-49

Claims table

Screenshot from 2022-12-15 10-13-04

Source selection

Screenshot from 2022-12-15 10-13-39

Locked token gift circle

Screenshot from 2022-12-15 10-13-47

## Related Issue

#1733 and #1734

PR #1774, opened 2022-12-16
WIP - Epoch Form Refinement

topocount:
WIP - Epoch Form Refinement

This is refactor of the Epoch Form, which provides a more familiar
interface for scheduling repeating epochs

State management using the incumbent primitives is very slow and breaks
sometimes, so this is not testable yet.

PR #1782, opened 2022-12-21
Locked token distribution frontend

levity:
Continuing #1769

DO NOT MERGE THIS -- it's stacked on WIP epoch changes so it can be tested on staging. See #1837 instead


src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -418,6 +581,73 @@ export function DistributionForm({
     return `Avail. ${commify(round(humanNumber, 2))}`;
   };
 
+  const getVaultOptions = (options?: {
```
levity, line 584: @bark4mark why do we need two toggles here? under what conditions should `includeHedgey` and `includeConnectWallet` have different values?

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -78,6 +99,19 @@ export function DistributionForm({
...
+    useState({} as any);
+
+  const loadPreviousLockedTokenDistribution = () => {
+    getPreviousLockedTokenDistribution(epoch.id).then(result =>
+      setPreviousLockedTokenDistribution(result)
+    );
+  };
+
+  useEffect(() => {
+    loadPreviousLockedTokenDistribution();
```
levity, line 112: this should not run unless the hedgey integration is enabled. maybe there's a way to move it into a subcomponent as I mentioned in [this comment](https://github.com/coordinape/coordinape/pull/1769#discussion_r1051263568)

levity: or, as an alternative, add it into the existing `getEpochData` query so that we can avoid sending a separate API request

btw, the naming is a bit misleading -- the `getPreviousDistribution` method uses "previous" in a different sense, because it's looking for a distribution from the previous _epoch_

src/pages/DistributionsPage/DistributionForm.tsx
levity: never mind, i understood it after digging into the code more. please check out my changes to the method & let me know if i'm missing somethign

src/pages/DistributionsPage/useLockedTokenDistributions.tsx
```diff
@@ -0,0 +1,103 @@
...
+  return async ({
+    amount,
+    gifts,
+    vault,
+    tokenContractAddress,
+    hedgeyLockPeriod,
+    hedgeyTransferable,
+    epochId,
+    totalGive,
+  }: any) => {
```
levity, line 32: it will help debugging a lot if this `any` is replaced with proper types

crabsinger:
FYI - just pushed this to our staging environment.

src/pages/DistributionsPage/useLockedTokenDistributions.tsx
bark4mark: I've updated this to have the types

levity:
I've rebased this branch off of `epoch-form-refinements` (#1774).

I had to do some manual conflict resolution, so I saved the original pre-rebase commit history at [this tag](https://github.com/coordinape/coordinape/releases/tag/locked-token-distribution-frontend-before-rebase) in case anything went missing and needs to be retrieved.

src/pages/ClaimsPage/queries.ts
```diff
@@ -4,6 +4,31 @@ import { Contracts } from 'lib/vaults';
 
 import type { Awaited } from 'types/shim';
 
+export const getLockedTokenGifts = async (profileId: number) => {
+  const { locked_token_distribution_gifts } = await client.query({
+    locked_token_distribution_gifts: [
+      {
+        where: {
+          profile_id: { _eq: profileId },
```
levity, line 12: you can only see rows that you own, due to the row select permissions in Hasura, so this profile_id check isn't needed

src/pages/DistributionsPage/DistributionForm.tsx
levity: addressed in afe64b0f

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -1,10 +1,17 @@
 import assert from 'assert';
```
levity, line 1: i still wish that the Hedgey-specific code in this file were kept more separate from the rest of it. But in the interest of getting this into production ASAP, we can consider that refactoring (plus adding more tests) to be follow-up work

levity:
@bark4mark i still wish that the Hedgey-specific code in DistributionForm.tsx were kept more separate from the rest of it. But in the interest of getting this into production ASAP, we can consider that refactoring (plus adding more tests) to be follow-up work.

PR #1785, opened 2022-12-22
Show users in csv if deleted after epoch ends

crabsinger:
Fixes bug where give sent doesn't match give received.

If you deelte a user after an epoch ends, they should still get old give, so show them inthe lsit but as deleted.


CryptoGraffe (approved):
Looks good. This will fix the current user problem. I'll put in a future ticket for the distribution table percentages not matching this.

PR #1787, opened 2022-12-23
fixing flaky tests that were failing due to data from other tests

CryptoGraffe:
## Motivation and Context

Test was using indexes instead of user names to fetch rows in the give page. Switched to using names which is more reliable after other tests goof with the users in this circle.


levity (approved):
I'm a bit nervous about hardcoding names but am ok with this to get CI green again asap

PR #1789, opened 2022-12-23
remove nominees.name column and fix ui

sshmm:
## Motivation and Context

remove name column from nominees table as the name now exists in profile table

## Description

1- remove nominees.column
2- remove nominees.name and replace it by nominees.profile.name
3- update create and update nominees mutations

## Test and Deployment Plan
1- Create nominee at members table
2- check the nominee name at nominees table and in discord message
3- vouch for the nominee check the user name

image

We need to run these sql statements before merging:
 in order to create profiles for nominees without profiles:
`INSERT INTO profiles(address, name)`
`SELECT nominees.address, nominees.name`
`FROM nominees`
`WHERE nominees.address NOT IN (`
  `SELECT profiles.address `
  `FROM profiles `
  `GROUP BY profiles.address`
`) AND nominees.ended = False  ON CONFLICT DO NOTHING`

update profiles.name with nominees.name if profiles.name is null
`UPDATE profiles`
`SET name = nominees.name`
`FROM nominees`
`WHERE nominees.address IN (`
  `SELECT nominees.address `
  `FROM nominees `
  `GROUP BY nominees.address,nominees.ended`
  `HAVING COUNT(distinct LOWER(nominees.name)) = 1 AND nominees.ended = False`
`) `
`AND nominees.address = profiles.address AND profiles.name IS NULL`


levity:
Can you add the query we need to run to the PR description?

hasura/migrations/default/1671750861500_alter_table_public_nominees_drop_column_name/down.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."nominees" alter column "name" drop not null;
```
levity, line 1:
```suggestion
```

levity:
I ran some modified versions of the SQL:
```sql
insert into profiles (address, name) select address, min(name) from nominees 
where address not in (select address from profiles) group by address;
-- inserted 50 rows

select p.address, n.name from profiles p, nominees n 
where p.address = n.address 
and p.name is null 
and p.address not in (select address from users) 
and n.name not in (select name from profiles);
-- returned 0 rows, meaning there are no profiles that can be assigned names from nominees 
-- without overwriting a user name, so there's nothing to be done there
```

PR #1793, opened 2022-12-27
Allow editing profile names

### LINKED ISSUE #1738: Allow editing username in profiles
**Problem**
Once we ship the Username feature there's not a way to edit the username

**Solution**
Build a way to edit the username from the profile

**Specification**
What precisely needs to get built? (include any requirements or design files)
Add an Edit user name button to the profile page

Part of #1395

sshmm:
## Motivation and Context
resolves #1738

## Description
1- changed editProfileForm to have user name input field and edited the schema for that
2 show error if the name is already in use

## Test and Deployment Plan
1- Go to profile page and click editprofile
2- update your name and click save

image


src/components/EditProfileModal.tsx
```diff
@@ -141,11 +167,33 @@ export const EditProfileModal = ({
...
+              Profile Image
+            </Text>
+            <AvatarUpload original={myProfile.avatar} />
+          </Flex>
+          <Flex column css={{ alignItems: 'center' }}>
+            <Text p css={sectionHeader}>
+              Profile Name{' '}
+              <Tooltip
+                content={
+                  <div>You can&apos;t edit your name during active epochs</div>
```
CryptoGraffe, line 182: Is there consensus around this? What if you are ALWAYS in an active recurring epoch?

CryptoGraffe: Seems like this warning should only show during an active epoch

src/pages/ProfilePage/queries.ts
```diff
@@ -0,0 +1,42 @@
...
+              epochs: [
+                {
+                  where: {
+                    _and: [
+                      { ended: { _eq: false } },
+                      { start_date: { _lt: 'now' } },
+                      {
+                        circle: {
+                          organization: {
+                            name: { _neq: 'Sample DAO' },
```
CryptoGraffe, line 25: This should use the sample boolean instead of this string

src/pages/ProfilePage/queries.ts
```diff
@@ -0,0 +1,42 @@
...
+                  },
+                },
+                { id: true },
+              ],
+            },
+          },
+        ],
+      },
+    ],
+  });
```
CryptoGraffe, line 39: would be nice to have an operationName in here

CryptoGraffe (commented):
A couple minor comments here, with some more discussion in discord. Seems to work ok for me when testing.

src/components/EditProfileModal.tsx
CryptoGraffe: Is this *also* enforced on the server side?

crabsinger: @sshmm did you see these?

crabsinger: 👍

crabsinger (approved):

PR #1795, opened 2022-12-29
inform nominators about ignored names

### LINKED ISSUE #1737: When a user is invited or nominated and they have an existing different username, notify about it
**Problem**
Once we have moved to one username per profile (#1395), there can be a conflict with an admin providing a new username when inviting or nominating. We need to detect this on the backend and return information such that the inviter/nominator can discover that despite them inviting an address with a particular name, that profile/address already had a valid name.

sshmm:
## Motivation and Context

resolves #1737

## Description

1- if the user has a profile with a name in our system. inform the nominator that this name will be user

## Test and Deployment Plan
1- Test 1
- Nominate a member with a unique name
- Nominate the same address in a different circle with a different name
expected: an alert will inform the nominator of the name that will be user

Screenshot from 2022-12-29 14-56-49

### DIFF
--- a/src/pages/NewNominationPage/NewNominationPage.tsx
+++ b/src/pages/NewNominationPage/NewNominationPage.tsx
@@ -305,7 +305,7 @@ export const NewNominationPage = () => {
                 <Flex>
                   <Check color="successDark" size="lg" css={{ mr: '$md' }} />
                   <Text size="large">
-                    You have successfully Nominated {nomineeName} to the circle
+                    You have successfully nominated {nomineeName} to the circle
                   </Text>
                 </Flex>
               </Panel>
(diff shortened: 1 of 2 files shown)


src/pages/NewNominationPage/NewNominationPage.tsx
```diff
@@ -297,14 +300,30 @@ export const NewNominationPage = () => {
...
-                  You have successfully Nominated {nomineeName} to the circle
-                </Text>
-              </Flex>
-            </Panel>
+            <>
+              <Panel success css={{ mt: '$xl' }}>
+                <Flex>
+                  <Check color="successDark" size="lg" css={{ mr: '$md' }} />
+                  <Text size="large">
+                    You have successfully Nominated {nomineeName} to the circle
```
levity, line 308:
```suggestion
                    You have successfully nominated {nomineeName} to the circle
```

levity (approved):

PR #1803, opened 2023-01-05
Create new toast system

crabsinger:
## Motivation and Context

We've been using notistack for our toast system, however this is built coupled on Material UI and doesn't support themes and stitches. We replace notistack with a `react-toastify` based toast implementation.

## Description

* remove notistack, and replace with react-toastify
* style light and dark themes for our toasts
* introduce three toast variants `showDefault`, `showError`, and `showSuccess` (we used to only have showInfo and showError)
* switch old showInfo toasts with a "saved" or "success" message to use `showSuccess` instead.

Current Theme:

Screenshot 2023-01-13 at 10 59 19 AM

**Dark Theme:**

Screenshot 2023-01-13 at 10 11 51 AM

## Reviewers

@wingfeatherwave  @CryptoGraffe 

## Testing
Here are some useful buttons for invoking and testing these toasts:

```
<Box>
<Button
            onClick={() => {
              showError(
                'Oh snap, you successfully read this important alert message. This example text is going to run a bit longer so that you can see how spacing within an alert works with this kind of content. Be sure to use margin utilities to keep things nice and tidy.'
              );
            }}
            color="primary"
            pill
          >
            Error Multiline
          </Button>
          <Button
            onClick={() => {
              showSuccess(
                'Well done, you successfully read this important alert message. This example text is going to run a bit longer so that you can see how spacing within an alert works with this kind of content. Be sure to use margin utilities to keep things nice and tidy.'
              );
            }}
            color="primary"
            pill
          >
            Success Multiline
          </Button>
          <Button
            onClick={() => {
              showDefault(
                ' This is a long info message, but its really great because you are really great, really really. Yes you are Im so great too. Were all great. We neven need some more text. What kind of made up words can I create like globularus and the textimonisnyess'
              );
            }}
            color="primary"
            pill
          >
            Info Multiline
          </Button>
          <Button
            onClick={() => {
              showError('The file flowbite-figma-pro was permanently deleted.');
            }}
            color="primary"
            outlined
            pill
          >
            Error
          </Button>
          <Button
            onClick={() => {
              showSuccess(
                'The action that you have done was a success! Well done'
              );
            }}
            color="primary"
            outlined
            pill
          >
            Success
          </Button>
          <Button
            // as={NavLink}
            // to={paths.createCircle}
            onClick={() => {
              showDefault('🦄 Wow so easy!');
            }}
            color="primary"
            outlined
          >
            Info
          </Button>
        </Box>
```

## Todos

- Add accessibility handling (ability to tab and 'Enter' to dismiss toasts)


src/components/ToastContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+    maxWidth: '640px',
+    width: '80%',
+  },
+  '.Toastify__toast-icon': {
+    marginInlineEnd: '20px',
+  },
+  '.Toastify__progress-bar': {
+    height: '1px',
+  },
+  '.Toastify__toast--info': {},
```
crabsinger, line 20: using no border for `info` ...

src/components/ToastContainer.tsx
```diff
@@ -0,0 +1,51 @@
...
+  '.Toastify__toast--success': {
+    borderColor: '$successColor',
+  },
+  '.Toastify__toast--error': {
+    borderColor: '$errorColor',
+  },
+  '.Toastify__toast--success *': {
+    color: 'var(--toastify-text-color-success)',
```
crabsinger, line 28: @wingfeatherwave are there fixed color variable names we can just use for these?

src/stitches.config.ts
```diff
@@ -374,6 +374,12 @@ export const dark = createTheme({
     mapNodeFade: '#FFFFFF20',
     mapLink: '#FFFFFF15',
     mapLinkDim: '#FFFFFF08',
+
+    // react-toastify
```
crabsinger, line 378: comments or no comments?

src/stitches.config.ts
```diff
@@ -239,6 +231,14 @@ export const {
     transitions: {
       quick: 'all 0.2s',
     },
+    toastify: {
+      'toast-width': '80% !important',
+      'text-color-error': newFigmaColors.error6 + ' !important',
+      'text-color-success': newFigmaColors.success6 + ' !important',
```
crabsinger, line 237: we could also do some of these via direct CSS in useToast or ToastContainer... so maybe its better to do it all in one place? But the variable changing is a bit lighter weight.

src/components/ToastContainer.tsx
wingfeatherwave: I think using the toastify values is the right thing to do since success and error color isn't different between dark/light.  BTW, I believe you can write this as `$toastify-text-color-success` instead of `var(--toastify-text-color-success)`

src/stitches.config.ts
wingfeatherwave: Not sure what you're asking here.  I think we should use scoped names if possible, or reuse things like $toastify-text-color-error.  As I think errorColor is too generic

src/stitches.config.ts
wingfeatherwave: I like having the stitches variables wherever possible.  I'm really surprised we need the `!important`s in there.  What's the point of them using css variables if they can't be easily overwritten.  Is there a cascade issue going on, like maybe their css needs to be imported at a different spot?

src/components/ToastContainer.tsx
wingfeatherwave: As discussed, I think we will want a border for the default state as there are some overlapping grey/white problems.

wingfeatherwave:
I made a lot of comments here... perhaps we can pair on finishing all this up when I return to Portland.

src/components/ToastContainer.tsx
crabsinger: I just tried to switch to $ vars and it isn't working...

Screenshot 2023-01-12 at 10 14 10 AM

I think these are used by CSS itself not Stitches, so they need to be in the plain old CSS syntax... or I'm doing something else wrong.

crabsinger:
@wingfeatherwave this is ready for final approval

wingfeatherwave (approved):

PR #1808, opened 2023-01-06
fix epoch null displayed at circles page

sshmm:
## Motivation and Context

1- display Upcoming epoch instead of epoch null at circles page
2- refetch orgs query after adding epochs

## screenshots and tests
image

### DIFF
--- a/src/pages/CirclesPage/CirclesPage.tsx
+++ b/src/pages/CirclesPage/CirclesPage.tsx
@@ -340,7 +340,15 @@ export const CircleRow = ({ circle, onButtonClick, state }: CircleRowProps) => {
             )}
           </Text>
           {epoch && startDate && endDate && (
-            <Text>{epoch.description ?? `Epoch ${epoch.number}`}</Text>
+            <Text>
+              {epoch.description
+                ? epoch.description
+                : epoch.number !== null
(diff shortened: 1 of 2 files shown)


src/pages/CirclesPage/CirclesPage.tsx
```diff
@@ -340,7 +340,15 @@ export const CircleRow = ({ circle, onButtonClick, state }: CircleRowProps) => {
             )}
           </Text>
           {epoch && startDate && endDate && (
-            <Text>{epoch.description ?? `Epoch ${epoch.number}`}</Text>
+            <Text>
+              {epoch.description
```
crabsinger, line 344: Thanks for the quick fix. I wonder if we could make this into a helper function, or otherwise have this be clearer to read rather than the nested ternarys.... @sshmm

crabsinger (approved):

src/pages/CirclesPage/CirclesPage.tsx
sshmm: done

CryptoGraffe (approved):

crabsinger:
@sshmm thanks - i like the helper function pattern a lot more. also would be easy to test in the future.

PR #1818, opened 2023-01-11
Script used for backfilling pGives

zashton:
## Motivation and Context

Generation of historical pGive data in preparation for retrieval from SBT app

## Description

Added an backend handler to trigger the backfilling of pGives data by circles

## Test and Deployment Plan

is triggered by cron. 

set `BACKFILL_TO` in the .env file to the previous month's last day e.g `BACKFILL_TO=2022-12-31` to enable the backfilling.
populates `epoch_pgive_data` `member_epoch_pgives` tables accordingly

## Reviewers

@topocount


api/hasura/actions/_handlers/backfillPgive.ts
```diff
@@ -0,0 +1,418 @@
...
+        __alias: {
+          ...ops,
+        },
+      },
+      {
+        operationName: 'insert_pGiveData',
+      }
+    );
+  }
+  res.status(200).json({ success: true });
```
CryptoGraffe, line 332: This should return something like a count processed or something so you know when it's done or there are no more circles to process, otherwise how do you know when to stop backfilling? Also, how is this really going to be used? Just call in a loop until all circles have been handled? One concern is if a single circle takes >30s vercel will time us out.

api/hasura/actions/_handlers/backfillPgive.ts
```diff
@@ -0,0 +1,418 @@
...
+    ],
+  });
+  return circles;
+};
+
+const roundNumbers = (value: number) => {
+  return Math.round(value * 100) / 100;
+};
+
+export default verifyHasuraRequestMiddleware(handler);
```
CryptoGraffe, line 418: Anyone can call this? Should be admin only

crabsinger: I'm thinking if this takes a long time it could be a DDOS vector if anyone can call it, so maybe some scheme for restricting who can invoke this to protect against that.

topocount (commented):
@zashton this is really nicely composed. I apologize for the late turnaround on this request, but could you write some basic tests for this just to demonstrate some expected table outputs? I don't think you'd need to adjust the seed script, but it'd be good to have a baseline.

api/hasura/actions/_handlers/backfillPgive.ts
zashton: So i've moved this to a cron fully and batched it, the cron i've set it to timeout `120` seconds.
I'd tested the previous verions of the code with prod data before and the handler could run like 15-150 circles (depending the length of epochs it had) in 1 go . 

I've also added a `PGIVE_CIRCLE_MAX_PER_CRON` env var and defaulted it to 10 circles.

Theres a sub filter that filters away circles that already had pgive records generated , so the cron is going to keep processing until this filters to nothing.
```
circles: [
      {
        limit: parseInt(process.env.PGIVE_CIRCLE_MAX_PER_CRON || '') || 10,
        where: {
          epochs: {
            ended: { _eq: true },
            end_date: { _gte: startFrom.toISO(), _lte: endTo.toISO() },
            _not: {
              /* this filters away epochs that already has pgive data generated */
              pgive_data: {},
            },
          },
        },
      },
      {
        id: true,
      },
    ],
```

api/hasura/actions/_handlers/backfillPgive.ts
zashton: have moved it to a cron task so this should be solved

api/hasura/actions/_handlers/backfillPgive.ts
```diff
@@ -0,0 +1,418 @@
...
+} from '../../../../src/lib/zod';
+Settings.defaultZone = 'utc';
+
+/* These are the main variables used in the algo to calculate the pGive totals   */
+
+const BASE_ACTIVE_POINTS = 100;
+const MAX_BASE_ACTIVE_TOTAL = 3000;
+const PER_ACTIVE_MONTH_BONUS = 4;
+const MAX_ACTIVE_MONTH_BONUS = 48;
+const MAX_NOTE_BONUS_PER_USER = 30;
```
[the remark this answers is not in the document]
zashton: yeah great idea, have just added env vars for them

api/hasura/actions/_handlers/backfillPgive.ts
```diff
@@ -0,0 +1,418 @@
...
+
+async function handler(req: VercelRequest, res: VercelResponse) {
+  const {
+    input: { payload },
+  } = composeHasuraActionRequestBody(backfillPgiveInput).parse(req.body);
+  const { circle_limit } = payload;
```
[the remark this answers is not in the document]
zashton: yeah the hasura handlers could max only run up to 45 secs before timing out. 
I've now moved it to a cron job and have added `PGIVE_CIRCLE_MAX_PER_CRON` env var and default it to 10.

api/hasura/actions/_handlers/backfillPgive.ts
```diff
@@ -0,0 +1,418 @@
...
+      {
+        limit,
+        order_by: [{ id: order_by.asc }],
+        where: {
+          _and: [
+            {
+              _not: {
+                epochs: {
+                  pgive_data: {},
+                },
```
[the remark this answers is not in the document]
zashton: The backfilling processes circle by circle, all the ended epochs in the circle at one go and creates an `epoch_pgive_data` for each of them and also `member_epoch_pgives` for each recipient if there were pgive values for that epoch , the `epoch_pgive_data` db table has a unique key on the `epoch_id` so it cannot be inserted twice. On top of that i've now added a filter `const epochs = circle.epochs.filter(e => !e.pgive_data)` in the processing logic just incase.

```
export const getCirclesNoPgiveWithDateFilter = async (
  startFrom: DateTime,
  endTo: DateTime
): Promise<Array<number>> => {
  const { circles } = await adminClient.query({
    circles: [
      {
        limit: parseInt(process.env.PGIVE_CIRCLE_MAX_PER_CRON || '') || 10,
        where: {
          epochs: {
            ended: { _eq: true },
            end_date: { _gte: startFrom.toISO(), _lte: endTo.toISO() },
            _not: {
              /* this filters away epochs that already has pgive data generated */
              pgive_data: {},
            },
          },
        },
      },
      {
        id: true,
      },
    ],
  });
  return circles.length ? circles.map(c => c.id) : [];
};

```

zashton:
thanks for the review @crabsinger @CryptoGraffe @topocount , i've made changes according to your suggestions. Previously was using a handler because it was easier to test. now i've moved it to a full cron.

zashton:
@topocount i've updated this PR with the monthly pgive generation as well since its using the same algo processing. i've tweaked it slightly to include `dateFrom`, `endTo` parameters in the same cron so it doesn't process the same epochs that the backfilling did. And added a `BACKFILL_TO` env var as a flag to start backfilling in the cron. WIll update this with tests soon

topocount (approved):

PR #1819, opened 2023-01-11
Refactor Time and Date Pickers to raw elements

topocount:
The integration isn't quite working yet because the datetime picker element doesn't handle time zones. So while we're successfully saving start_times, I'm not sure we're submitting them with the correct timestamp.

Secondly, displaying existing epoch start times and modifying them hasn't successfully been integrated yet either 
So we're still using one field value but we need to make sure the local time zone is appended to it before it is submitted to the api

TODOs:
1. Handle submitting:     // BEFORE SUBMITTING: we need to convert the time from the UI time picker into a time with the local timezone. Or correct absolute time.
2. Handle reading existing values: When we fetch data, we're getting a time in UTC, and need to convert it to local time. Probably by calling set zone and then passing it to the element. Then just call .toISO on it and pass it to the UI datetime picker.

image

### DIFF
--- a/src/components/FormDatePicker.tsx
+++ b/src/components/FormDatePicker.tsx
@@ -1,31 +1,14 @@
 import React from 'react';
 
-import { DateType } from '@date-io/type';
+import { FieldValues } from 'react-hook-form';
 
-import { DatePicker } from '@material-ui/pickers';
+import { FormInputField, TFormInputField } from 'components/index';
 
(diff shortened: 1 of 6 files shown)


topocount:
@wingfeatherwave just a heads up The DatePicker component is a bit wonky currently, and expands outside it's prescribed footprint.

topocount:
Closing for now, since we're on track to not need this.

Issue #1816, opened 2023-01-10
implement "monthly" logic in the createEpoch handler


topocount:
The TS types exist for this already. We just need to flesh out the logic and check that the selected weekday isn't greater than 7.

PR #1773, opened 2022-12-16
Implement New CreateEpoch API

topocount:
This new API includes functionality for custom and one-off epochs to be
scheduled. monthly epochs (epochs that start and end on the same weekday
each month) will be implemented after the intial integration for
one-offs and custom repeating epochs is complete and they are beginning
to be tested

Test plan:
- unit tests are provided
- existing e2e tests should still pass.

Merge plan:
This work will be merged into main after this is approved for merge.

PR #1772, opened 2022-12-16
Update jest typedefs to match jest version and zod

topocount:
## Motivation and Context

These dependencies are out of date. and need to be updated for future
PRs in this stack.

## Test and Deployment Plan

Test Suite should pass

PR #1771, opened 2022-12-16
Optimize TestManager for backend test development

topocount:
- make interactive mode utilize nodemon
- move seed script invocation to only run before cypress tests
- make stop_services cleanup script delete db volume on quit

test plan: existing tests should all pass

PR #1829, opened 2023-01-17
Implement Epoch Monthly backend

topocount:
## Motivation and Context

This implements the backend for creating a repeating epoch that starts
on the same day of the same week each month.

## Description

This logic accepts a start date, end date, and week number. The start
date and end date must be the same weekday. They also must fall on the
week number of the month. The first week of the month is week 0.

The week number is a necessary extra parameter because it can't
necessarily be inferred from the date. If a following date overflows
into the following month, the inferred week number would cause all
following epochs to start at the beginning of the following month than
the end of the current month, as was originally intended

## Test Plan

unit tests are provided

## Merge Plan

This can merge after #1773 is merged

## Related Issue

#1816

### DIFF
--- a/api-test/hasura/actions/_handlers/createEpoch.test.ts
+++ b/api-test/hasura/actions/_handlers/createEpoch.test.ts
@@ -609,59 +609,6 @@ describe('createEpoch', () => {
   });
 
   describe('monthly input', () => {
-    describe('findSameDayNextMonth', () => {
-      it('handles month beginnings correctly', () => {
-        let date = DateTime.fromISO('2023-01-01T20:28:00.000Z');
-        let result = findSameDayNextMonth(date, {
-          week: 0,
-        });
-        expect(result.toISO()).toBe('2023-02-05T20:28:00.000Z');
-
-        date = DateTime.fromISO('2023-02-06T20:28:00.000Z');
-        result = findSameDayNextMonth(date, {
-          week: 0,
-        });
-        expect(result.toISO()).toBe('2023-03-06T20:28:00.000Z');
-
-        date = DateTime.fromISO('2023-03-06T20:28:00.000Z');
-        result = findSameDayNextMonth(date, {
-          week: 0,
-        });
-        expect(result.toISO()).toBe('2023-04-03T20:28:00.000Z');
-      });
-
-      it('handles month endings correctly', () => {
-        const date = DateTime.fromISO('2023-01-31T20:28:00.000Z');
-
-        const result = findSameDayNextMonth(date, {
-          week: 4,
-        });
-        // truncated from 5th week to last week
-        expect(result.toISO()).toBe('2023-02-28T20:28:00.000Z');
-      });
-      test('handles weekly boundaries correctly', () => {
-        // week 0
-        // 1st Friday each month
-        let start = DateTime.fromISO('2023-04-07T00:00:00.000Z');
-        let result = findSameDayNextMonth(start, { week: 0 });
-        expect(result.toISO()).toBe('2023-05-05T00:00:00.000Z');
-        // week 5
-        // 6th Friday each month truncated to last Friday
-        result = findSameDayNextMonth(start, { week: 5 });
-        expect(result.toISO()).toBe('2023-04-28T00:00:00.000Z');
-
-        // week 1
-        // 2nd Saturday each month
-        start = DateTime.fromISO('2023-04-08T00:00:00.000Z');
-        result = findSameDayNextMonth(start, { week: 1 });
-        expect(result.toISO()).toBe('2023-05-13T00:00:00.000Z');
-        // week 6
-        // 6th Saturday each month truncated to the last Saturday
-        result = findSameDayNextMonth(start, { week: 6 });
-        expect(result.toISO()).toBe('2023-04-29T00:00:00.000Z');
-      });
-    });
-
     test('errors on malformed input', async () => {
       const now = DateTime.now();
       const params = {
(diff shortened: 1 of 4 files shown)


topocount:
* main
  * **PR #1771** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1771" target="_blank">Graphite</a> 
    * **PR #1772** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1772" target="_blank">Graphite</a> 
      * **PR #1773** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1773" target="_blank">Graphite</a> 
        * **PR #1774** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1774" target="_blank">Graphite</a> 
          * **PR #1829** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1829" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1829?utm_source=stack-comment).

api/hasura/actions/_handlers/createEpoch.ts
```diff
@@ -213,4 +230,42 @@ async function checkOverlappingEpoch({
   }
 }
 
+export function findSameDayNextMonth(
+  start: DateTime,
+  { week }: { week: number }
```
CryptoGraffe, line 235: curious about these two diff args types, one is object (w/ one field) one is straight straight param

api/hasura/actions/_handlers/createEpoch.ts
```diff
@@ -108,6 +108,23 @@ async function handler(request: VercelRequest, response: VercelResponse) {
   insertNewEpoch(response, input);
 }
 
+function validateMonthlyInput(
+  input: z.infer<typeof zMonthlyInputSchema>
+): ErrorReturn {
+  const { start_date, end_date } = input;
```
CryptoGraffe, line 114: Why does end_date need to be input in this case? Shouldn't we just calc it?

api-test/hasura/actions/_handlers/createEpoch.test.ts
```diff
@@ -545,13 +548,182 @@ describe('createEpoch', () => {
...
+            type: 'monthly',
+            week: 6,
+          },
+          start_date: expect.stringContaining(
+            params.start_date.substring(0, 16)
+          ),
+          end_date: expect.stringContaining(params.end_date.substring(0, 16)),
+        })
+      );
+    });
```
CryptoGraffe, line 727: nice job on the thorough testing

api-test/hasura/actions/_handlers/createEpoch.test.ts
```diff
@@ -545,13 +548,182 @@ describe('createEpoch', () => {
...
+        let date = DateTime.fromISO('2023-01-31T20:28:00.000Z');
+
+        let result = findSameDayNextMonth(date, {
+          week: 4,
+        });
+        expect(result.toISO()).toBe('2023-03-07T20:28:00.000Z');
+
+        date = DateTime.fromISO('2023-02-07T20:28:00.000Z');
+        result = findSameDayNextMonth(date, {
+          week: 5,
```
CryptoGraffe, line 581: I'm confused about the week arg i guess, If its the 5th week, why is the test from 02-07 and why do we need the week? Shouldn't the week of the month be calculated from the start date rather than provided?

api/hasura/actions/_handlers/createEpoch.ts
CryptoGraffe: I saw the comment in the PR and i think i understand, but i'm worried about user confusion, maybe we need to discuss . Does the user have to specify a week number?

CryptoGraffe:
FWIW google implements this by doing "monthly on the 4th tuesday" and then if you pick a 5th tuesday it changes it to "monthly on the last tuesday" instead of 5th/6th etc. Maybe this is too hard to change at this point. I'm trying to figure out wher ethe `week` field is incorporated into the front end but I haven't figure it out yet. When I choose Monthly there is no date entry and it just chooses today and says "Every month on the 24th" , so i'm not sure how to choose a different date or to tweak the week parameter, which afaict is client driven at this point. Let me know if it would be easier to talk this through a bit.

CryptoGraffe:
Realizing now this is only the backend impl, so the lack of FE makes sense (It's in the dang name!)

api/hasura/actions/_handlers/createEpoch.ts
topocount: @CryptoGraffe a date is picked from a datepicker and the UI informs them of the pattern like "The epoch repeats on the **2nd Tuesday** of each month"

api/hasura/actions/_handlers/createEpoch.ts
topocount: @CryptoGraffe this is just Typescript notion. I'm destructuring the param I need on the left side and defining the param on the right.

topocount:
Since my git commit message amendment just caused tests to be skipped on HEAD, here's the link to the last tests that ran: https://github.com/coordinape/coordinape/actions/runs/4048036564

src/common-lib/epochs.ts
```diff
@@ -0,0 +1,54 @@
...
+  return nextEndDate;
+}
+
+/**
+ * @dev This function is not safe for generating subsequent epoch start dates
+ * in an existing cycle of repeating monthly epoichs
+ * and is only meant to provide a porcelain interface for users creating a new
+ * set of repeating epochs, either in the UI or via the API.
+ */
+export function findMonthlyEndDate(start: DateTime): DateTime {
```
levity, line 51: this isn't used?

api/hasura/actions/_handlers/createEpoch.ts
CryptoGraffe: What I meant is why doesn't the function just consistently take the arg it needs (eg start and week) so the destructuring or whatever would happen in the caller. Not important just curious why.

PR #1842, opened 2023-01-20
fix/add operation names

crabsinger:
## Add operation_name to all Graphql requests

Having an `operation_name` is very useful for debugging and performance
monitoring of our requests. We add a TS requirement that makes operation_name
required in the makeThunder client for FE and admin requests, and then add names to all requests that
didn't have them.

### DIFF
--- a/api/hasura/actions/_handlers/updateUser.ts
+++ b/api/hasura/actions/_handlers/updateUser.ts
@@ -26,24 +26,27 @@ async function handler(req: VercelRequest, res: VercelResponse) {
 
   const {
     users: [user],
-  } = await adminClient.query({
-    users: [
-      {
-        limit: 1,
-        where: {
-          address: { _ilike: address },
-          circle_id: { _eq: circle_id },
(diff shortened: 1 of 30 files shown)


CryptoGraffe (dismissed):

api/hasura/actions/_handlers/updateUser.ts
```diff
@@ -26,24 +26,27 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
-      {
-        id: true,
-        fixed_non_receiver: true,
-        give_token_received: true,
-      },
-    ],
-  });
+      ],
+    },
+    { operationName: 'getUsers' }
```
crabsinger, line 48: not sure on a nomenclature, but maybe this should have a __filename postfix for generic operation names

CryptoGraffe: ok i changed this here

CryptoGraffe (approved):
Made a couple tweaks and merging

PR #1851, opened 2023-01-23
show allocation table for all chains

### LINKED ISSUE #1846: Bug: Can't access CSV when not on Main net
Problem:
I can't access a distribution page when not connected to main net. 

Image

Proposed Solution:
Allow access to the Distribution page, but disable the Give Circle and Fixed Payment forms/actions.
This way, folks could still use the "Export CSV" action without being forced to switch to Main Net.

Image

A nice to have would be a message in the forms that says 
"Gift Circle payments disabled on Chain ID: [chain.id]" & "Fixed Payments are disabled on Chain ID: [chain.id]"

sshmm:
## Motivation and Context
fixes #1846

## Description

1- vaults related features are disabled
2- Admins can still enter the allocation amount to export it on CSV

image

### DIFF
--- a/src/pages/DistributionsPage/DistributionForm.tsx
+++ b/src/pages/DistributionsPage/DistributionForm.tsx
@@ -553,6 +553,7 @@ export function DistributionForm({
   };
 
   const shouldDisableGiftInput = () => {
+    if (!contracts) return true;
     if (existingLockedTokenDistribution?.tx_hash) return true;
     if (hedgeyIntegration?.data.enabled) return false;
     else
@@ -660,7 +661,6 @@ export function DistributionForm({
                   includeHedgey: hedgeyIntegration?.data.enabled,
                   includeConnectWallet: false,
                 })}
-                disabled={!contracts}
               ></Select>
             </Box>
             <Box css={{ width: '100%' }}>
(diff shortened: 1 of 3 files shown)


crabsinger:
@sshmm this looks good but the tests broke - can you fix this tests and ping me for re-review

src/pages/DistributionsPage/DistributionForm.tsx
```diff
@@ -897,7 +901,7 @@ export function DistributionForm({
           ) : vaults[0] ? (
             <Button
               color="secondary"
-              disabled={giftSubmitting || !sufficientGiftTokens}
+              disabled={giftSubmitting || !sufficientGiftTokens || !contracts}
```
crabsinger, line 904: Ithink this is breaking the tests because it adds another invariant where contracts being defined makes this button NOT disabled, where it should be disabled.

crabsinger: actually, nvm, that doesn't make sense.

sshmm:
> @sshmm this looks good but the tests broke - can you fix this tests and ping me for re-review

done

crabsinger (approved):

PR #1864, opened 2023-01-24
Remove user name field from create circle form

sshmm:
## Motivation and Context

Remove user name field from create circle form because profile name is used in all circles now

## Description

1- Remove name Field
2- fetch data by react-query and remove recoil queries
3- handle landing on the form page immediately after setting the profile name

## Test and Deployment Plan
1- go to app.coordinape.come/new-circle and create a circle
expected : the circle will be created successfully and your name will be used there


sshmm:
converting to draft until fixing test issues

src/hooks/useConnectedAddress.ts
```diff
@@ -2,7 +2,7 @@ import { useEffect, useState } from 'react';
 
 import { useWeb3React } from 'hooks/useWeb3React';
 
-export default function useConnectedAddress() {
+export function useConnectedAddress() {
```
crabsinger, line 5: was there a reason for this change?

crabsinger (changes requested):
@sshmm this looks good 

one question on the useConnectedAddress changes

second is a blocker: seems the logo image is not saving when set during the circle create form.

src/hooks/useConnectedAddress.ts
sshmm: to be able to import useConnectedAddress like this 
`import { useConnectedAddress } from 'hooks/useConnectedAddress';`
I had troubles mocking it in unit test and was getting this error:
`Error: Uncaught [TypeError: (0 , _useConnectedAddress.default) is not a function]`

levity: hmm... could you try:

```js
jest.mock('hooks/useConnectedAddress', () => {
  return {
    default: () => 'abc',
  };
});
```

sshmm: unfortunately it gave the same error as before. I reverted the changes and used another fix please check it

crabsinger: sounds good

crabsinger (approved):

PR #1867, opened 2023-01-25
New stitches friendly loading modal

CryptoGraffe:
## Motivation and Context

Old loader was always blue , new one is more fun and can use stitches colors

Note: this look makes it look blurry/ghosty but it isn't

https://www.loom.com/share/86aab23feeea41049634f6b35e6831e4

The modal for this still uses some MUI we need to remove the Modal


CryptoGraffe:
@wingfeatherwave maybe you have an easy way for us to remove the mui modal here

wingfeatherwave (dismissed):

wingfeatherwave (approved):
got rid of mui modal

PR #1879, opened 2023-01-28
dismiss profile when click outside

CryptoGraffe:
## Motivation and Context

When you navigate to anything outside the profile or click outside the profile box , it should dismiss.

### DIFF
--- a/src/features/nav/NavProfile.tsx
+++ b/src/features/nav/NavProfile.tsx
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+/* eslint-disable no-console */
+import React, { useEffect, useRef, useState } from 'react';
 
 import { Network } from '../../components';
 import { CreateUserNameForm } from '../../components/MainLayout/CreateUserNameForm';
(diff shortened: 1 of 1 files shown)


src/features/nav/NavProfile.tsx
```diff
@@ -1,4 +1,5 @@
-import React, { useState } from 'react';
+/* eslint-disable no-console */
```
crabsinger, line 1: @CryptoGraffe can we remove this line...

src/features/nav/NavProfile.tsx
```diff
@@ -20,8 +21,27 @@ export const NavProfile = ({
...
+    const handleClickOutside = (event: MouseEvent) => {
+      if (
+        ref.current &&
+        event.target &&
+        !ref.current.contains(event.target as Node)
+      ) {
+        setOpen(false);
+      }
+    };
+    document.addEventListener('click', handleClickOutside, true);
```
crabsinger, line 36: clever solution... does this cause two re-renders on a click though?

PR #1880, opened 2023-01-28
Track contribution created events in Mixpanel

crabsinger:
Track contribution created events in Mixpanel

Using a event_trigger pattern on the INSERT.

Verified these show in Mixpanel (logged to user and circle)

image

### DIFF
--- a/api/hasura/event_triggers/eventManager.ts
+++ b/api/hasura/event_triggers/eventManager.ts
@@ -1,8 +1,9 @@
-import { VercelRequest, VercelResponse, VercelApiHandler } from '@vercel/node';
+import { VercelApiHandler, VercelRequest, VercelResponse } from '@vercel/node';
 
 import checkNomineeDiscord from '../../../api-lib/event_triggers/checkNomineeDiscord';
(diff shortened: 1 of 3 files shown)


api/hasura/event_triggers/eventManager.ts
```diff
@@ -23,22 +24,23 @@ import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
 type HandlerDict = { [handlerName: string]: VercelApiHandler };
 
 const HANDLERS: HandlerDict = {
+  checkNomineeDiscord,
```
crabsinger, line 27: I sorted these

crabsinger:
@CryptoGraffe I think this works, except I am not seeing my data in MIXPANEL staging when I run with that key set in my env, but i think this works?

crabsinger:
> @CryptoGraffe I think this works, except I am not seeing my data in MIXPANEL staging when I run with that key set in my env, but i think this works?

was using the key not the token, now its working.

CryptoGraffe (approved):

PR #1918, opened 2023-02-06
Implement Epoch Cron Logic For New Schema

topocount:
This PR contains the epoch cron that will create repeating epochs using
the new `repeat_data` schema. A timzone has been added to the schema to
minimize the complexity of the calculations on the backend. However, the
input of this timzone is pretty unopinionated. If an unsupported zone is
entered, or just a nonsense string is, the timezone is configured to be
UTC, which eliminates any shifts due to DST. Otherwise, DSTs will be
observed when the timezone is applied to a datetime.

Merge Plan

This implementation is meant to be merged before the migration and
cutover, since the code supporting the old epoch logic is still present
and the encasing cron logic will function exactly the same until
the `repeat_data` column is populated. At that point, the new repeating
logic will kick in.

### DIFF
--- a/api/hasura/cron/_epochs.test.ts
+++ b/api/hasura/cron/_epochs.test.ts
@@ -1,5 +1,5 @@
 import faker from 'faker';
-import { DateTime } from 'luxon';
+import { DateTime, Interval } from 'luxon';
 
 import { adminClient } from '../../../api-lib/gql/adminClient';
 import { sendSocialMessage } from '../../../api-lib/sendSocialMessage';
@@ -10,6 +10,7 @@ import {
   EpochsToNotify,
   endEpoch,
   makeNextStartDate,
+  calculateNextEpoch,
 } from './epochs';
 
 jest.mock('../../../api-lib/gql/adminClient', () => ({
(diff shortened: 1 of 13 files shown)


api/hasura/cron/epochs.ts
```diff
@@ -445,7 +461,110 @@ export function makeNextStartDate(
   });
 }
 
-async function createNextEpoch(epoch: {
+export async function createNextEpoch(epoch: {
+  id: number;
+  start_date: string;
+  end_date: string;
+  circle_id: number;
+  repeat_data: unknown;
```
crabsinger, line 469: Could we type this more specifically?

crabsinger (approved):
@topocount this looks good to me. Creating and editing epochs works fine testing locally, and CI is passing.

One question on if we can add tighter typing to an`unknown`.

otherwise approving.

api/hasura/cron/_epochs.test.ts
```diff
@@ -1,5 +1,5 @@
 import faker from 'faker';
```
levity, line 1: can you merge this file with the one in `api-test/hasura`?

topocount:
> One question on if we can add tighter typing to an`unknown`.

@crabsinger thanks for the review! I typed this as `unknown` because it's at worst a random `jsonb` payload from the database, and I don't want to "trust" the `createEpoch` and `updateEpoch` hooks to _always_ provide correct data, so I parse it here and reject it if it's not "expectably" typed:

https://github.com/coordinape/coordinape/blob/6570521d7e2a25b88df03c3eb6c4cd872a214b8b/api/hasura/cron/epochs.ts#L473-L478

My thinking is that at the worst case in the future, we've migrated away from an old schema and an older, long-repeating epoch cycle is no longer supported. This logic will just cause it to cease repeating. We might consider adding a discord alert for this case in the future.

api/hasura/cron/_epochs.test.ts
topocount: Actually @levity we can't/shouldn't merge these files because the gql client is stubbed in these tests but I can colocate it with a more descriptive rename

PR #1921, opened 2023-02-06
Create activities table

crabsinger:
# Create an activities table to build an activities feed

- create database migrations
- setup hasura events
- create event_trigger and test

## Testing:
- create a contribution, add a user, create an epoch, and expect to see the activities table auto populated.
- also added tests for the createActivity file

## Questions:
- we wanted the activities table to hold `actor_profile_id` and `target_profile_id` but that data isn't always passed through, so we need to pass it through or fetch it later.
For users and epochs, we don't have the actor_profile....

## API:

Standard graphql using relations to expose nested activity data types

**Request**
```graphql
query Activities {
  activities {
    epoch {
      id
    }
    action
    user {
      address
      entrance
    }
    contribution {
      description
      id
    }
  }
}
```

**Response**


hasura/migrations/default/1675459146542_create_activities_table/up.sql
```diff
@@ -0,0 +1,45 @@
+CREATE TABLE "public"."activities" (
```
crabsinger, line 1: @levity i dont need you to fully review this code (unless you want to) but since this is a big new DB architecture, I invite your feedback and thoughts before this is merged as its much easier to incorporate changes earlier rather than later.

api-test/hasura/event_triggers/createActivity.payloads.ts
```diff
@@ -0,0 +1,107 @@
+export const payloads = {
```
CryptoGraffe, line 1: is there a better naming convention so that this is obviously for a test?

hasura/metadata/databases/default/tables/public_contributions.yaml
```diff
@@ -90,6 +90,26 @@ select_permissions:
               _is_null: true
       allow_aggregations: true
 event_triggers:
+  - name: contributionInsertCreateActivity
```
CryptoGraffe, line 93: might be nice to rename all these to have activity in the front so they are grouped in the UI

hasura/migrations/default/1675459146542_create_activities_table/up.sql
```diff
@@ -0,0 +1,45 @@
...
+    "epoch_id" bigint,
+    "user_id" bigint,
+
+    "action" varchar(100) NOT NULL,
+
+    "created_at" timestamptz NOT NULL DEFAULT now(),
+    "updated_at" timestamptz NOT NULL DEFAULT now(),
+
+  PRIMARY KEY ("id") ,
+  FOREIGN KEY ("circle_id") REFERENCES "public"."circles"("id") ON UPDATE CASCADE ON DELETE set null,
```
CryptoGraffe, line 17: If we are going to cascade (vs restrict) perhaps we should just delete this activity instead of orphan it? idk

api-test/hasura/event_triggers/createActivity.payloads.ts
CryptoGraffe: maybe even a subdirectory

hasura/migrations/default/1675459146542_create_activities_table/up.sql
levity: this looks good to me -- i can see that avoiding more exotic polymorphism patterns would probably be most compatible with Hasura, and also GraphQL generally -- it seems like [the spec is still in flux](https://github.com/graphql/graphql-spec/issues?q=polymorphism+is%3Aopen)

api-lib/gql/mutations.ts
```diff
@@ -533,3 +533,15 @@ export async function insertInteractionEvents(
   );
   return;
 }
+
+export async function insertActivity(
```
levity, line 537: same comment here as in `api-lib/gql/queries`... actually, in this case, the mutation is so straightforward i wouldn't bother splitting it out into a separate function at all

api-lib/gql/queries.ts
```diff
@@ -63,6 +63,50 @@ export async function getCircles() {
...
+          users: [
+            { where: { id: { _eq: user_id } } },
+            {
+              profile: { id: true },
+            },
+          ],
+        },
+      ],
+    },
+    {
```
levity, line 83: please avoid unnecessary newlines where possible

hasura/migrations/default/1675459146542_create_activities_table/up.sql
```diff
@@ -0,0 +1,45 @@
...
+    "id" bigserial NOT NULL,
+    "circle_id" bigint,
+    "organization_id" bigint NOT NULL,
+    "actor_profile_id" bigint,
+    "target_profile_id" bigint,
+    "contribution_id" bigint,
+    "epoch_id" bigint,
+    "user_id" bigint,
+
+    "action" varchar(100) NOT NULL,
```
levity, line 11: this field needs stricter validation -- it should fail to insert if it isn't in a specified set of strings

levity (changes requested):

hasura/metadata/databases/default/tables/public_contributions.yaml
crabsinger: ok, how about we name the trigger `activityHandler` since we probably will want to support more than just update in that same file.

and then we can rename these to `activityHandlerContributionInsert`

how's that sound?

hasura/migrations/default/1675459146542_create_activities_table/up.sql
crabsinger: yeah, that sounds reasonable. I've updated this to cascade deletes.

api-lib/gql/queries.ts
```diff
@@ -63,6 +63,50 @@ export async function getCircles() {
   ).circles;
 }
 
+export async function getOrgAndProfile(user_id: number, circle_id: number) {
```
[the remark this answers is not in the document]
crabsinger: done

api-test/hasura/event_triggers/createActivity.test.ts
```diff
@@ -0,0 +1,149 @@
...
+
+jest.mock('../../../api-lib/gql/adminClient', () => ({
+  adminClient: { query: jest.fn() },
+}));
+
+jest.mock('../../../api-lib/gql/mutations', () => ({
+  insertActivity: jest.fn(),
+}));
+
+describe('Create Activity Event Trigger', () => {
```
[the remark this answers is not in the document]
crabsinger: ok, yeah, this output is plenty nice:

Screenshot 2023-02-15 at 9 31 04 AM

api/hasura/event_triggers/eventManager.ts
```diff
@@ -24,6 +25,9 @@ import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
 type HandlerDict = { [handlerName: string]: VercelApiHandler };
 
 const HANDLERS: HandlerDict = {
+  usersInsertCreateActivity: createActivity,
```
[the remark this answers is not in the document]
crabsinger: 1 handler, but three event_triggers that fire that same handler. each trigger is tied to a table action "contribution_insert" or something like that. We're going to eventually have a lot of triggers, but hopefully only one handler. @levity

hasura/migrations/default/1675459146542_create_activities_table/up.sql
crabsinger: talked with graffe - holding off on this for now.

crabsinger:
Ok, addressed all the feedback. I think this is ready to merge. I think our types are a bit weird right now, as I am seeing some diffs with `yarn generate` around GuildInfoInput added and Discord stuff being removed... so not sure about that...

api/hasura/event_triggers/eventManager.ts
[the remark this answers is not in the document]
levity: what i meant was, is it possible to define a single handler here named `actvityHandler` and have all three new `event_triggers` YAML sections all refer to that same handler with 
```diff
-  - name: activityHandlerContributionInsert
+  - name: activityHandler
``` 

but i just checked the docs and the event trigger name has to be unique, i guess 🤷

api-lib/event_triggers/activity/activityHandler.ts
```diff
@@ -0,0 +1,142 @@
+import type { VercelRequest, VercelResponse } from '@vercel/node';
```
levity, line 1: now that you've created a folder named `activity`, having the files also be prefixed with `activity` seems redundant to me, and none of the other files have `Handler` in their name, so how about renaming to `index.ts` and `mutations.ts`?

crabsinger: ok, addressed all the feedback.

levity (dismissed):

CryptoGraffe (approved):

CryptoGraffe (approved):

PR #1924, opened 2023-02-07
disable New button when new contribution is still saving

### LINKED ISSUE #1890: Contributions get lost if entered between network requests
**Summary:**

See GIF: https://app.usebubbles.com/nd5EMG6ZMvamsAoenLLxwh/contributions-get-lost-on-quick-new-button

Creating contributions via the New button in quick succession will overwrite previous Contributions.

All contributions are recorded, and none our lost.

4. **Steps to recreate:**

Enter contributions quickly, or setup a delay inthe hasura handler to make it easier to reproduce the slowness, then entere contributions.

levity:
## Motivation and Context

Does not fully deal with #1890 (with a workaround that prevents the race condition from occurring, rather than a more involved rewrite to allow concurrent saving of new contributions)

This is currently stacked on #1923 but could also be merged directly to main.

### DIFF
--- a/api-test/helpers/users.ts
+++ b/api-test/helpers/users.ts
@@ -1,14 +1,22 @@
+import assert from 'assert';
+
 import faker from 'faker';
 
 import { GraphQLTypes } from '../../api-lib/gql/__generated__/zeus';
 
 import type { GQLClientType } from './common';
+import { createProfile } from './profiles';
 
 export async function createUser(
   client: GQLClientType,
   object: Partial<GraphQLTypes['users_insert_input']>
-): Promise<{ id: number }> {
-  const { insert_users_one } = await client.mutate(
+) {
+  if (!object.address) {
+    const profile = await createProfile(client);
+    object.address = profile.address;
+  }
+
+  const { insert_users_one: user } = await client.mutate(
     {
       insert_users_one: [
         {
@@ -20,14 +28,17 @@ export async function createUser(
             entrance: object.entrance ?? '?',
           },
         },
-        { id: true },
+        {
+          id: true,
+          address: true,
+          circle_id: true,
+          role: true,
+          profile: { id: true, address: true, name: true },
+        },
       ],
     },
     { operationName: 'createUser' }
   );
-  if (!insert_users_one) {
-    throw new Error('User not created');
-  }
-
-  return insert_users_one;
+  assert(user, 'User not created');
+  return user;
 }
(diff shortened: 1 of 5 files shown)


levity:
* main
  * **PR #1923** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1923" target="_blank">Graphite</a> 
    * **PR #1924** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1924" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1924?utm_source=stack-comment).

crabsinger:
@levity approving and merging, as this is an improvement, however this doesn't resolve the bug because if I close the panel, and create a new contribution quickly I can still reproduce the race condition. We could disable that possibility by also limiting the parent "Add Contribution" button while in flight... but I think at some point actually fixing this may be worth it.

Anyhow, with your fix it should cover the most common case of contributions getting lost.

crabsinger (approved):

levity:
oh, that's a good point. we could extend the disabling behavior to that
other button as well without much effort. i intentionally held off on
"actually fixing this" because the concurrency logic is pretty hard to
follow because of how it hardcodes the assumption that there's only one
"in-flight" new contribution at a time and it's so coupled to the rest of
the contributions data and its linked-list structure. it would probably
take me at least half as much time as writing the feature from scratch. i
was initially tempted to just use incrementing negative IDs for new
contributions and keep everything else the same as much as possible, but
thinking about the amount of QA needed, on top of what's already been sunk
into this, had me want to just cut the Gordian knot, so to speak.

crabsinger:
I think just disablign the other button as well is a good enough solution
for me for now... - wanna crank that out?

levity:
Did it in 1155cda2 after seeing your comment

yahiamarzouk:
Can't we just add a delay func to adding new contribution before saving?
Just want experience on this subject

levity:
The problem is that it's possible to submit two requests to save new contributions at the same time, so that the second one overwrites the first one. Adding a delay to both saves wouldn't change the problem. Adding a delay to just the second save would make the problem less likely to happen, but it wouldn't guarantee that the issue would be fixed. It would also increase the risk that the user could close the page before their second contribution got saved

yahiamarzouk:
> The problem is that it's possible to submit two requests to save new contributions at the same time, so that the second one overwrites the first one. Adding a delay to both saves wouldn't change the problem. Adding a delay to just the second save would make the problem less likely to happen, but it wouldn't guarantee that the issue would be fixed. It would also increase the risk that the user could close the page before their second contribution got saved

I very wonder how can we fix that

levity:
The code assumes there’s only one new contribution saving at any time, so
it uses the special value of 0 for the ID of new contributions. The code
would have to be rewritten not to use just one special value.

PR #1946, opened 2023-02-15
Create User Circle Event Handlers

topocount:
These handlers will be used to dispatch notification info to the discord
bot. Handlers have not been implemented for discord webhooks or
telegram, but it can be done with minimal lift if desired.

## Test Plan

This code is still not accessible from within prod and requires integration
with the bot, so that's how it will be tested for now.

### DIFF
--- a/api-lib/handleUserAddedMsg.ts
+++ b/api-lib/handleUserAddedMsg.ts
@@ -1,0 +1,80 @@
+import { isFeatureEnabled } from '../src/config/features';
+
+import * as queries from './gql/queries';
+import {
+  Channels,
+  DiscordUserAddedOrRemoved,
+  sendSocialMessage,
+} from './sendSocialMessage';
+import { Awaited } from './ts4.5shim';
+import { EventTriggerPayload } from './types';
+
+type GetChannelsProps = {
+  data: EventTriggerPayload<'users', 'INSERT' | 'UPDATE'>['event']['data'];
+  circle: Awaited<ReturnType<typeof queries.getCircle>>['circles_by_pk'];
+  profiles: Awaited<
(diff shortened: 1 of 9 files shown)


api-lib/handleUserRemovedMsg.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    return {
+      isDiscordBot: true,
+      discordBot: {
+        type: 'user-removed',
+        channelId,
+        roleId,
+        discordId: user?.user_snowflake,
+        address: data.new.address,
+        circleName: circle?.name ?? 'Unknown',
```
karelianpie, line 45: In here we'd like to have `methodOfRemoval` and `refunds`

https://github.com/coordinape/coordinape/pull/1948

karelianpie (changes requested):
Missing the `method` prop, discord bot messages refactored with https://github.com/coordinape/discord-bot/pull/48

api-lib/handleUserAddedMsg.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    return {
+      isDiscordBot: true,
+      discordBot: {
+        type: 'user-added',
+        channelId,
+        roleId,
+        discordId: user?.user_snowflake,
+        address: data.new.address,
+        circleName: circle?.name ?? 'Unknown',
```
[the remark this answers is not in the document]
topocount: followup: talked this over with @reeserj and we're removing this feature from the current set of deliverables, since it's not supported in our core app at this time.

api-lib/handleUserRemovedMsg.ts
topocount: same as above

PR #1948, opened 2023-02-15
feat: Add a few more properties

karelianpie:
Split `DiscordUserAdded` and `DiscordUserRemoved` as we need different properties for them.

### Data missing for DiscordUserAdded
```typescript
type DiscordUserAdded = DiscordEpochEvent & {
  // ...
  methodOfAddition: string;
};
```
### Data missing for DiscordUserRemoved
```typescript
type DiscordUserRemoved = DiscordEpochEvent & {
  // ...
  methodOfRemoval: string;
  refunds: { username: string; give: number }[];
};
```

### DIFF
--- a/api-lib/handleUserAddedMsg.ts
+++ b/api-lib/handleUserAddedMsg.ts
@@ -3,7 +3,7 @@ import { isFeatureEnabled } from '../src/config/features';
 import * as queries from './gql/queries';
 import {
   Channels,
-  DiscordUserAddedOrRemoved,
+  DiscordUserAdded,
   sendSocialMessage,
 } from './sendSocialMessage';
 import { Awaited } from './ts4.5shim';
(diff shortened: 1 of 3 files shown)


topocount (commented):
@karelianpie we don't need the refunds in this handler. The `OptOut` handler runs when a user is removed from a giving circle for any reasons. This is a design feature of reactive event handling. Data emergently composes, rather than passing monolithic data blocks around.

karelianpie:
> @karelianpie we don't need the refunds in this handler. The `OptOut` handler runs when a user is removed from a giving circle for any reasons. This is a design feature of reactive event handling. Data emergently composes, rather than passing monolithic data blocks around.
> 
> This also allows us to genercize the method of removal/addition to `method` and continue utilizing a single type across addition/removal events.

I see, that's a problem for the way @reeserj originally set up the messages:

Screenshot 2023-02-16 at 8 43 28 am

I've refactored the discord message with https://github.com/coordinape/discord-bot/pull/48

karelianpie:
Not applicable

PR #1950, opened 2023-02-15
ensNames validation cron event

### LINKED ISSUE #1925: Create ENS profile name cron check
Write a cron that runs every hour and validates that all profile names are valid ENS names if they end in .eth.

If a profile name is not valid, add an asterisk`*` to the end of the profile name. If it's invalid (because of uniqueness, keep adding asterisks until you get to a valid field).

Run the cron every hour.

Check if we can batch the lookups to Infura rather than doing one request per name.

abstract code:
```
const ens_profiles = SELECT * from profiles
WHERE profiles.name ends with '.eth'.

ens_profiles.each({profile =>
if profile.valid(): next
else: handle_invalid_name()
```

Validate the job runs in less than 30 seconds, and batch if needed.
Validate we have plenty of Infura API calls on our account.

sshmm:
## Motivation and Context

resolves #1925

## Description

1- cron that triggers every hour to validate ens names
2- unresolved names will be trailed by *

## Test and Deployment Plan
1- add users with names that ends with .eth
expected: those names will be trailed by * if they are unresolved or resolved to a different address


api/hasura/cron/ensNames.ts
```diff
@@ -0,0 +1,82 @@
...
+      },
+      {
+        operationName: 'endEpoch_insertAndDeleteGifts',
+      }
+    );
+  } catch (e: any) {
+    if (e.response.errors[0].message.includes('Uniqueness violation')) {
+      updateEnsName({ id: id, name: modifiedName });
+    } else {
+      Promise.reject(e);
```
levity, line 77: i think you meant `return Promise.reject(e)` here. just creating a promise value but not returning it does nothing.

note, though, that if one of these rejects, then `await Promise.all` in the handler will throw, and the handler doesn't have a try/catch block. is this intentional?

levity (changes requested):

api/hasura/cron/ensNames.ts
sshmm: yes verifyHasuraRequestMiddleware has a try catch block

levity (approved):

PR #1951, opened 2023-02-16
Upgrade Hasura

topocount:
Our local image of hasura is has fallen nearly 10 increments behind our cloud console, so this is a proactive measure to avoid any issues due to inconsistencies

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,7 +13,7 @@ services:
 
   # Hasura
   graphql-engine:
-    image: hasura/graphql-engine:v2.13.2.cli-migrations-v3
+    image: hasura/graphql-engine:v2.19.0.cli-migrations-v3
     volumes:
       - ./hasura/migrations:/hasura-migrations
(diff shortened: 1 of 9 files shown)


docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
 
   # Hasura
   graphql-engine:
-    image: hasura/graphql-engine:v2.13.2.cli-migrations-v3
+    image: hasura/graphql-engine:v2.19.0.cli-migrations-v3
```
crabsinger: Our cloud is running on v2.18.0-cloud.2, so I propose we upgrade to that, but not beyond it for now.
Screenshot 2023-02-20 at 11 03 44 AM

crabsinger: i've pushed a commit with 2.18

crabsinger (approved):
looks good -ran this locally and seems fine.

CryptoGraffe (approved):

PR #1952, opened 2023-02-16
Discord Bot: Add column for alert notification config

topocount:
This is meant to be populated as follows:
```ts
{
  [alert_type]: boolean
}
```
that way, it's simple to utilize postgres, and generated gql types to
read configs and modify configs in this JSONB column.


karelianpie (approved):
LGTM!

PR #1953, opened 2023-02-16
validate add members list ens names

sshmm:
## Motivation and Context

issue raised by @crabsinger :
if you use the Add Members page, and use a ENS .eth in the Name field, we won't verify it, and then it will be set on the user. 

## Description

1- verify the list of names on FE and BE

## Test and Deployment Plan
go to member page and try to add a mix of addresses with correct and wrong ens names

front-end
Screenshot from 2023-02-16 17-53-04

back-end if front-end is bypassed
Screenshot from 2023-02-16 18-23-17

### DIFF
--- a/api-lib/validateENS.ts
+++ b/api-lib/validateENS.ts
@@ -1,0 +1,12 @@
+import { getProvider } from './provider';
+
+export const isValidENS = async (name: string, address?: string) => {
+  const resolvedAddress = await getProvider(1).resolveName(name);
+  if (
+    !resolvedAddress ||
+    resolvedAddress.toLowerCase() !== address?.toLowerCase()
+  ) {
+    return false;
+  }
+  return true;
+};
--- a/api/hasura/actions/_handlers/createUsers.ts
+++ b/api/hasura/actions/_handlers/createUsers.ts
@@ -12,6 +12,7 @@ import {
   errorResponseWithStatusCode,
   InternalServerError,
 } from '../../../../api-lib/HttpError';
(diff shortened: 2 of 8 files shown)


crabsinger (approved):
This works and looks decent @sshmm. However, since we are doing ENS validation in 4 different places now:

```
src/pages/AddMembersPage/NewMemberList.tsx|140 col 54| const resolvedAddress = await provider().resolveName(m.name);
src/components/EditProfileModal.tsx|136 col 48| const resolvedAddress = await provider().resolveName(params.name);
src/components/MainLayout/CreateUserNameForm.tsx|66 col 48| const resolvedAddress = await provider().resolveName(data.name);
api/hasura/actions/_handlers/updateProfile.ts|42 col 50| const resolvedAddress = await getProvider(1).resolveName(name);
api/hasura/actions/_handlers/createUsers.ts|55 col 54| const resolvedAddress = await getProvider(1).resolveName(user.name);
api/hasura/cron/ensNames.ts|12 col 52| const resolvedAddress = await getProvider(1).resolveName(profile.name);

```

I am curious if we could clean this up a bit by having a helper function that does the validation and then just calling that in each of these different places. Allthough we would to have custom logic for handlign the errors and messages depending on where it is invoked.

but maybe something like isValidEns(member) or something might be nice to abstract out the direct call `resolveName` with provider.

crabsinger (approved):
@sshmm looks good - nice work refactoring. I think in the future, one shared function would be ideal, but for now, having one across api and one for frontend with zod makes sense.

PR #1956, opened 2023-02-17
Implement Variable Dust Allowance

topocount:
Because the evm deals in fixed-point arithmetic, the order of magnitude
of the dust allowance should increase with the quantity of decimals a
token has.

For example, it does not make sense for DAI (18 decimals) to have the
same dust allowance as USDC (6 decimals) since that would mean we are
allowing 12 orders of magnitude less dust for DAI, making it much harder
to submit a distribution.

Test Plan

Unit Tests are included and existing tests are updated.

### DIFF
--- a/src/lib/merkle-distributor/index.test.ts
+++ b/src/lib/merkle-distributor/index.test.ts
@@ -1,7 +1,7 @@
 import { BigNumber } from '@ethersproject/bignumber';
 import padStart from 'lodash/padStart';
 
-import { createDistribution } from '.';
+import { createDistribution, getMaxAllowableDust } from '.';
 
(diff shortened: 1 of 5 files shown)


levity (changes requested):
Hmm. I originally very intentionally set the dust limit to a tiny, tiny amount relative to an 18-decimal-place token. It shouldn't scale with token decimals, because it's a compensation for the underlying fixed-point math system, not the token abstraction on top of it. It should be the smallest possible number that we can get away with. I would support setting it to 100 or something instead of 20, or an even higher number if we can construct a test case that requires it, but definitely not 2*10^13. If we need 2*10^13 it means we are making rounding errors. For the same reasons, for a token with a low number of decimals, a dust limit of 2 will be too low.

PR #1913, opened 2023-02-04
circle activity page

CryptoGraffe:
## Motivation and Context

Contribution activity stream for circle. 

Turn on with `localStorage.setItem('feature:activity',true)`

PR #1967, opened 2023-02-23
Add created_by to epoches

crabsinger:
Add created_by to epoches to track which user profile was the one who
created the epoch.


crabsinger:
* main
  * **PR #1913** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1913" target="_blank">Graphite</a> 
    * **PR #1967** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1967" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1967?utm_source=stack-comment).

api/hasura/actions/_handlers/createEpoch.ts
```diff
@@ -128,7 +129,7 @@ async function handler(request: VercelRequest, response: VercelResponse) {
     return;
   }
 
-  insertNewEpoch(response, input);
+  insertNewEpoch(response, input, creatorProfileId);
```
crabsinger, line 132: this is kinda messy, but the input type is what comes from the FE, so don't want to add to that... curious if there's a better paradigm.

hasura/schema/user/schema.graphql
```diff
@@ -5372,36 +5378,6 @@ type epochs {
   ): [pending_token_gifts!]!
   grant: numeric!
   id: bigint!
-
-  """
-  An array relationship
-  """
-  member_epoch_pgives(
```
CryptoGraffe, line 5379: Is this stuff happening even when you do a clean restart?

api-lib/event_triggers/activity/index.ts
```diff
@@ -45,7 +45,7 @@ export default async function handler(req: VercelRequest, res: VercelResponse) {
           event: {
             op: operation,
             data: {
-              new: { id, circle_id, created_at },
+              new: { id, circle_id, created_at, created_by },
```
CryptoGraffe, line 48: wondering if created_by should be created_by_profile or something since we have a lot of ambiguity about profile vs user in the app

api-lib/gql/__generated__/zeus/const.ts
```diff
@@ -2669,6 +2660,8 @@ export const AllTypesProps: Record<string, any> = {
     circle: 'circles_bool_exp',
     circle_id: 'Int_comparison_exp',
     created_at: 'timestamp_comparison_exp',
+    created_by: 'Int_comparison_exp',
```
CryptoGraffe, line 2663: Actually now i'm noticing both are in here, and not totally understanding

CryptoGraffe (commented):
I only partially reviewed this. Would be nice to chat more when we are both online

hasura/schema/user/schema.graphql
crabsinger: yes it is 😢

api-lib/gql/__generated__/zeus/const.ts
crabsinger: i also added a relationship `created_by_profile` to the profiles table using `created_by` as the column key.

hasura/schema/user/schema.graphql
CryptoGraffe: you fixed it yay

CryptoGraffe (approved):
Looks good! I changed the epoch activity to have optional actor_profile_id since there will be historical epochs (and future recurring epochs) with no profile id

CryptoGraffe (approved):

PR #1972, opened 2023-02-24
Remove old Epoch Action handlers

topocount:
This is dead code now that we've successfully migrated over to the new
Epoch API

test plan: unit tests should pass


crabsinger (approved):
lgtm - just waiting for CI to pass.

crabsinger (changes requested):
@topocount just noticed this: https://github.com/coordinape/coordinape/blob/main/api/hasura/cron/epochs.ts#L556 needs to also be removed/updated

crabsinger (approved):

PR #1993, opened 2023-03-01
a few small fixes

levity:
just some things that were worth quickly fixing as i was working on org member access

PR #1994, opened 2023-03-01
store profile data globally

levity:
this stores profile data locally. i've added it to ~the existing zustand auth store for now~ react-query, and to avoid bikeshedding about the structure of local data storage, i am just storing the entire `profiles_by_pk` subtree of the fetchManifest query.

in the short term, this will be used to determine whether the user is a member of a given org, without incurring additional requests. in the longer term it will also help us continue to move away from Recoil.

### DIFF
--- a/src/features/auth/useFinishAuth.ts
+++ b/src/features/auth/useFinishAuth.ts
@@ -1,6 +1,7 @@
 import assert from 'assert';
 
 import * as Sentry from '@sentry/react';
+import { useQueryClient } from 'react-query';
 
 import { DebugLogger } from '../../common-lib/log';
 import { useApiBase } from 'hooks';
@@ -9,6 +10,7 @@ import { useWeb3React } from 'hooks/useWeb3React';
 
 import { findConnectorName } from './connectors';
 import { login } from './login';
+import { QUERY_KEY_LOGIN_DATA } from './useLoginData';
 import { useLogout } from './useLogout';
 import { useSavedAuth } from './useSavedAuth';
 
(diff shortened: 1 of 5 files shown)


levity:
* main
  * **PR #1993** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1993" target="_blank">Graphite</a> 
    * **PR #1994** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1994" target="_blank">Graphite</a>  👈
      * **PR #1995** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/1995" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/1994?utm_source=stack-comment).

src/features/auth/store.ts
```diff
@@ -1,12 +1,19 @@
 import type { JsonRpcProvider } from '@ethersproject/providers';
 import create from 'zustand';
 
+import type { QueryManifest } from 'hooks/useApiBase';
```
levity, line 4: i don't love the dependency chain here... but i could probably resolve it by pulling fetchManifest into this feature

CryptoGraffe (approved):
I approved but I'm nervous about storing the results of the profile query w/ no invalidation pattern for it. Would rather just store the id and/or addr

levity:
In persistent local storage, I’m storing only the ID and address. This is
storing the results of the profile query in Zustand only until the page is
closed/reloaded

levity:
It runs every time the page is loaded from scratch. I know the “login”
naming is probably misleading, but this runs whether you’re logging in for
the first time or reusing an auth token that is still valid

CryptoGraffe (changes requested):
I'm removing approval until the invalidation stuff is sorted

levity:
@CryptoGraffe invalidation stuff has been addressed

CryptoGraffe (approved):

PR #1996, opened 2023-03-02
Co in yo circle

wingfeatherwave:
* use our brand co for coordinape user
* it seems that the s3 image is unreliable, as it can be deleted or expire?? so I'm using an image I'm adding `/public`

Screenshot 2023-03-01 at 4 12 58 PM


CryptoGraffe:
I suspect the bio gets wiped at end of epoch. Have you tested that? We can improve that in a follow-on, maybe an exception for the wiping in the end epoch cron.

CryptoGraffe (approved):

levity:
Since we have control over the Coordinape user's data, another option here would have been to update the avatar properties for all Coordinape users. Then no changes to Avatar.tsx, etc. would have been necessary

PR #2001, opened 2023-03-03
Fix React remove scrollbar console warning

wingfeatherwave:
Get rid of that annoying console warning.

@levity Is it ok to add this dep to upgrade the version, and then modify the lock file manually?


yarn.lock
```diff
@@ -20336,7 +20336,7 @@ react-remove-scroll@2.5.5:
   resolved "https://registry.yarnpkg.com/react-remove-scroll/-/react-remove-scroll-2.5.5.tgz#1e31a1260df08887a8a0e46d09271b52b3a37e77"
   integrity sha512-ImKhrzJJsyXJfBZ4bzu8Bwpka14c/fQt0k+cyFp/PBhTfyDnU5hjOtM4AG/0AMyy8oKzOTR0lDgJIM7pYXI0kw==
   dependencies:
-    react-remove-scroll-bar "^2.3.3"
+    react-remove-scroll-bar "^2.3.4"
```
wingfeatherwave, line 20339: I made this change manually, so I'm not sure if it's ok, or will be overwritten later.

crabsinger: would an upgrade cuz this to be overwritten? just running `yarn` and such doesn't overwrite this on my local env...

levity:
yes, i think so

CryptoGraffe (dismissed):

wingfeatherwave:
used resolutions 🫰

levity (approved):

PR #2002, opened 2023-03-03
Update user name mutations

### LINKED ISSUE #1929: Profile Name Rollout 4/4
1. Add a DB migration to remove the users.name field.

```
ALTER TABLE users DROP COLUMN name;
```

2. Make sure the Hasura metadata and Zeus schema files are updated to reflect the removal of this column.

3. Make sure profiles are always created with a unique placeholder name, like a substring of the address, or "New User N", where N is a unique incrementing number. (done in b97a2423)

4. Add add a NON-NULL constraint on profile.name

```
ALTER TABLE profiles
ALTER COLUMN name SET NOT NULL;
```

Run it

sshmm:
## Motivation and Context

resolves #1929

## Description

1- removed users.name
2- removed insert_users_trigger
3- fixed create and insert user mutations
4- fixed tests

## Test and Deployment Plan
1- Unit Test should pass without issues
2- Try adding users from members page


crabsinger:
@sshmm CI is still failing on this, but just one failing test

hasura/migrations/default/1677779258033_alter_table_public_users_drop_column_name/down.sql
```diff
@@ -0,0 +1,3 @@
+comment on column "public"."users"."name" is E'Members of a circle';
+alter table "public"."users" alter column "name" drop not null;
```
crabsinger, line 2: why do we need this line? doesn't the "add column" already set this as ? @sshmm

crabsinger (approved):
one question on the down migration, but probably will enver be run (its a down) so merging.

PR #2014, opened 2023-03-13
Login fix

levity:
This fixes a regression I introduced in my commits last week: after those changes, we were not checking that the stored auth token matched the currently connected account, and not clearing tokens correctly on logout.

to fix this, i restored the multi-account storage in useSavedAuth, then added a check in useFinishAuth to mark the correct address as the one whose token we should use.

I also ended up removing the Recoil dependency of useSavedAuth because it's more straightforward to use localStorage directly.


levity:
> @levity getting an error loading this page on vercel: image

Whoops! Please try again after this latest deployment finishes.

sshmm (approved):
looks good to me. I tested it with multiple addresses and reviewed the code

PR #2016, opened 2023-03-13
Epoch timezone test fix

levity:
I noticed that there were some tests with try/catch blocks so that even though assertions were failing, the test wasn't failing.

Looking more closely at the tests, I see that there was not only a minor issue with an assertion not accounting for shifting forward due to Daylight Saving Time, there's also a real bug which would prevent the creation of recurring epochs when Daylight Saving Time shifts back. I created a quick copy test to indicate this issue. Since it won't happen in production till November, we have some time to address it.

EDIT: Oops -- this is happening for users in Australia NOW. They shift back on Apr 2.


levity:
> Since it won't happen in production till November, we have some time to address it.

Oops -- this is happening for users in Australia NOW. They shift _back_ on Apr 2.

levity:
* main
  * **PR #2016** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/2016" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/2016?utm_source=stack-comment).

levity:
should be fixed in #2051

crabsinger:
@levity can I close this? or do you plan to come back to it?

levity:
You can close it but it does represent an unaddressed bug. Maybe mention it in a new issue?

PR #1995, opened 2023-03-01
allow org members to view circle members pages

levity:
This PR makes the circle members page accessible to org members. This means:

- As an org member (not a circle member), you can mouse over circles in the org page and see a link to the Members page, which you can click on
- When you do, the Members page will be loaded, and the side nav will show the circle you've selected (which doesn't show otherwise, since you're not a member
- The Members page has limited content: there's no admin button or Leave Circle button, of course

### Testing

Since there is no way in the UI to add a member directly to an org yet, you can do this from the REPL: run `await createMembersFromCircles(orgId)`, where `orgId` is the org you want to add members to. All circle members in that org will be made org members as well.

If you want to scrutinize the permissions setup, you can remove yourself from all circles in the org, and you should still be able to see each circle's members page in the limited way described above.

If you want to test this in the preview app, you need to also deploy this code to staging, and then run the REPL against staging (by setting NODE_HASURA_URL and HASURA_GRAPHQL_ADMIN_SECRET to staging values)

### Todo

- [x] stop using useSelectedCircle
- [x] combine queries on members page
- [x] set up permissions so org members can view circle, users, profiles
- [x] move MemberRow to its own file
- [x] select correct org in side nav when viewing a circle you're not part of
- [x] show non-member circles in side nav only when selected
- [x] fix circle overview page (deferred to future PR)
- [x] get map page working (deferred to future PR)
- [x] debug failing assertion when switching from non-member page to give page for a circle you're in
  - root cause: useSelectedCircle called before it's updated for the new circle

PR #2018, opened 2023-03-14
Allow activities and contributions to circle org members

crabsinger:
## Motivation and Context
Allow activities and contributions  select permission to circle org members

## Description

Update Hasura perms.

## Test and Deployment Plan

With activity enabled, add a user as org member but not circle member, and you should see contributions from circles, and activities for circles.

### DIFF
--- a/hasura/metadata/databases/default/tables/public_activities.yaml
+++ b/hasura/metadata/databases/default/tables/public_activities.yaml
@@ -48,11 +48,21 @@ select_permissions:
         - created_at
         - updated_at
       filter:
-        _and:
+        _or:
           - organization:
               circles:
                 users:
-                  profile:
-                    id:
+                  _and:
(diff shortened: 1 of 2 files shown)


CryptoGraffe (approved):

hasura/metadata/databases/default/tables/public_activities.yaml
```diff
@@ -48,11 +48,21 @@ select_permissions:
...
+                  _and:
+                    - profile:
+                        id:
+                          _eq: X-Hasura-User-Id
+                    - deleted_at:
+                        _is_null: true
+          - organization:
+              members:
+                _and:
+                  - id:
```
levity, line 64: this is a bug, because that's the org_member row id, not the profile id; will fix in #1995

PR #2027, opened 2023-03-15
Minor spacing fix for sidenav

wingfeatherwave:
Screenshot 2023-03-15 at 1 35 59 PM

added some bottom padding to the logo so overflow isn't right at the edge of the logo.


CryptoGraffe (approved):
thx for doing this

PR #2029, opened 2023-03-16
Org view for circle map page

levity:
Allow members of the org a circle is in to view that circle's map page

Plan: make it work by fetching the necessary data & loading it into Recoil when the map page is loaded, since for an org member, it is not prefetched on page load like it is for circle members. Work through any issues with permissions or assumptions in the code that the user is a member of the circle.

### DIFF
--- a/src/routes/routes.tsx
+++ b/src/routes/routes.tsx
@@ -66,15 +66,15 @@ const LoggedInRoutes = () => {
       <Route path="circles/:circleId" element={<OrgRouteHandler />}>
         <Route path="activity" element={<CircleActivityPage />} />
         <Route path="members" element={<MembersPage />} />
+        <Route path="map" element={<MapRouteHandler />}>
+          <Route path="" element={<LazyAssetMapPage />} />
+        </Route>
       </Route>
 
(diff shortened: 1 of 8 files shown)


src/routes/routes.tsx
```diff
@@ -66,15 +66,15 @@ const LoggedInRoutes = () => {
       <Route path="circles/:circleId" element={<OrgRouteHandler />}>
         <Route path="activity" element={<CircleActivityPage />} />
         <Route path="members" element={<MembersPage />} />
+        <Route path="map" element={<MapRouteHandler />}>
+          <Route path="" element={<LazyAssetMapPage />} />
```
tiff0000, line 70: is this path="" ? i'm not familiar with existing routes, just wanted to bring attention to making sure this was intended.

levity: yep, that's intentional. it means that once the parent route "map" is matched, this route should be executed for all nested routes. in this case it's just a way of running MapRouteHandler with LazyAssetMapPage nested inside it.

levity:
The map code has changed a lot since this was opened -- will start from scratch when I pick this up again.

PR #2033, opened 2023-03-16
try setting up retry 2 on cypress runs

crabsinger:
Not sure if this will work with our setup, but hopefully it does


levity:
our existing cypress tests will probably always break on retries because we're only seeding the data once on test suite setup, so the 2nd+ runs aren't starting with the same data as the 1st run

PR #2041, opened 2023-03-22
add event to add new users to org members

sshmm:
## Motivation and Context

add event to add new users to org members

## Description

1- the event will be triggered upon adding a new user or updating deleted_at column
2- the user will be added to org_members for this org if it's a new user or deleted_at column updated to be null
3- deleted users will be removed from org_members if they are deleted from all circles in that org

## Test and Deployment Plan
1- inserUserOrgMember.test.ts is added to test the description.
2- it can be tested from FE by adding users to a circle that you are an admin and check Org Members Table and check again after deleting them


hasura/metadata/databases/default/tables/public_users.yaml
```diff
@@ -188,6 +188,22 @@ event_triggers:
...
+      insert:
+        columns: '*'
+      update:
+        columns:
+          - deleted_at
+    retry_conf:
+      interval_sec: 10
+      num_retries: 5
+      timeout_sec: 60
+    webhook: '{{HASURA_API_BASE_URL}}/event_triggers/eventManager?event=insertUserOrgMember'
```
crabsinger, line 203: @sshmm It looks like you are doing two different checks via the same event_trigger:
`insertOrgMember` on add user to circle and 
`deleteOrgMember` on delete user from circle.

I think it would be cleaner to have two smaller, simpler, separate event_triggers to handle each of these cases separately.

hasura/metadata/databases/default/tables/public_users.yaml
```diff
@@ -188,6 +188,22 @@ event_triggers:
       paused: true
       schedule: 0 0 * * *
       timeout: 60
+  - name: insertUserOrgMember
+    definition:
+      enable_manual: false
+      insert:
+        columns: '*'
+      update:
```
crabsinger, line 196: We can have this be used for a separate event_trigger of `deleteOrgMember`

api-lib/event_triggers/insertUserOrgMember.ts
```diff
@@ -0,0 +1,129 @@
+import assert from 'assert';
```
crabsinger, line 1: maybe rename `insertOrgMember`

crabsinger (changes requested):
@sshmm left some comments

api-lib/event_triggers/insertUserOrgMember.ts
crabsinger: @sshmm can we rename to this

api-test/hasura/actions/_handlers/insertUserOrgMember.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import {
+  createCircle,
+  createProfile,
+  createUser,
+  mockUserClient,
+  createOrganization,
+} from '../../../helpers';
+import { getUniqueAddress } from '../../../helpers/getUniqueAddress';
+
+let address1, address2, profile1, profile2, circle1, circle2, client, org;
```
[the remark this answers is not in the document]
sshmm: they are just two users in the same org. no priviledge

hasura/metadata/databases/default/tables/public_users.yaml
sshmm: after removing the deleting part, insertUserOrgMember became much simpler

api-test/hasura/actions/_handlers/insertOrgMember.test.ts
```diff
@@ -34,7 +34,7 @@ beforeAll(async () => {
 });
 
 test('Check that the created user has been added to org_members', async () => {
-  await new Promise(res => setTimeout(res, 1000));
+  await new Promise(res => setTimeout(res, 2000));
```
levity, line 37: this kind of wait should be unnecessary... if it passes with a delay and fails without it, it's an indication that a promise / async function call is not being awaited properly

api-lib/event_triggers/insertOrgMember.ts
```diff
@@ -0,0 +1,77 @@
...
+  org_members_update_column,
+} from '../gql/__generated__/zeus';
+import { adminClient } from '../gql/adminClient';
+import { EventTriggerPayload } from '../types';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  try {
+    const payload: EventTriggerPayload<'users', 'INSERT' | 'UPDATE'> = req.body;
+    const user = payload.event.data.new;
+    const query = await adminClient.query(
```
levity, line 16: what you're trying to do here needs only a single query, because all the relationships are in the schema already (pseudocode below):
```
users_by_pk(id: userId)
  profile
    org_members(where: organization -> circle -> user -> id == userId)
```

api-lib/event_triggers/insertOrgMember.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    if (user.deleted_at === null) {
+      if (!org_members[0] || org_members[0].deleted_at !== null) {
+        adminClient.mutate(
+          {
+            insert_org_members_one: [
+              {
+                object: {
+                  profile_id: query.profiles[0].id,
+                  org_id: query.circles_by_pk?.organization_id,
```
crabsinger, line 53: can we move `query.circles_by_pk?.organization_id` to a variable for a bit cleaner code

hasura/metadata/databases/default/tables/public_users.yaml
crabsinger: if we delete a circle user, we don't do anything. so we don't need the `update`

hasura/metadata/databases/default/tables/public_users.yaml
```diff
@@ -188,6 +188,22 @@ event_triggers:
       paused: true
       schedule: 0 0 * * *
       timeout: 60
+  - name: insertOrgMember
+    definition:
+      enable_manual: false
+      insert:
+        columns: '*'
+      update:
```
crabsinger, line 196: if we undelete a circle user, (UDPATE) then we want to re-add them to the org. so this is correct.

api-lib/event_triggers/insertOrgMember.ts
```diff
@@ -0,0 +1,77 @@
...
+              ],
+            },
+          },
+          { id: true, deleted_at: true },
+        ],
+      },
+      { operationName: 'insertOrgMember_getOrgMember' }
+    );
+
+    if (user.deleted_at === null) {
```
levity, line 45: can you add a check in the beginning of the handler? basically if user.deleted_at is not null, we don't have to do anything now

api-lib/event_triggers/insertOrgMember.ts
```diff
@@ -0,0 +1,77 @@
...
+      if (!org_members[0] || org_members[0].deleted_at !== null) {
+        adminClient.mutate(
```
levity, line 47: missing `await`

api-test/hasura/actions/_handlers/insertOrgMember.test.ts
```diff
@@ -0,0 +1,98 @@
+import { adminClient } from '../../../../api-lib/gql/adminClient';
```
levity, line 1: please move this to `api-lib/event_triggers/insertOrgMember.test.ts`

levity: `api-test` is only for code in `api`, because Vercel tries to make everything the `api` directory into an API route (an alternative would have been to put test files in there but add `_` to the beginning of their names)

api-lib/event_triggers/insertOrgMember.ts
sshmm: done. 
do you think I should add `else` statement that has a different res.status(200) for that case or the current one is suitable for both cases

levity (approved):
@sshmm i did some refactoring; please note the changes to the test file and the early return in the handler (i prefer the early return to having a big block of code in an if-clause)

levity (dismissed):

PR #2047, opened 2023-03-23
add circle and/or org name to Mixpanel events

levity:


crabsinger (approved):
looks good, we could add a test too... ;)

CryptoGraffe (approved):

PR #2065, opened 2023-03-30
Remove all orgs vaults routes

sshmm:
## Motivation and Context

Removed all organization vaults content

## Description
Error:
1- creating a new circle with a new org (without vaults) will have a helper banner with a link to vaults page
image
2- Clicking this link will crash the page with the following error
Screenshot from 2023-03-29 16-45-58
3- it's caused by trying to access an org that's not fetched yet.

Fix:
1- Removed /vaults route and content
2- made all vaults routes specific to organizations

## Test and Deployment Plan
1- create a new circle with a new org
2- navigate  to Epochs/Overview page and click create a new vault

### DIFF
--- a/src/components/MainLayout/getMainHeaderData.ts
+++ b/src/components/MainLayout/getMainHeaderData.ts
@@ -5,8 +5,8 @@ import { useQuery } from 'react-query';
 import useConnectedAddress from 'hooks/useConnectedAddress';
 import { useWeb3React } from 'hooks/useWeb3React';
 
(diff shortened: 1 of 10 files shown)


sshmm:
> @sshmm CI is failing.

@crabsinger  @levity  do you think I should keep the failing test as it's searching for the removed vaults page? I can't think of a way to replace it

levity:
You can change the test to go to any other page instead, and look for content on that page

levity (approved):

PR #2067, opened 2023-03-30
Fix triggering addReaction twice

sshmm:
## Motivation and Context

Upon adding a reaction to an activity the following error is received:

"instrument.ts:124 Error: Uniqueness violation. duplicate key value violates unique constraint "reactions_profile_id_activity_id_reaction_key"
    at index.ts:54:17'

## Description
Cause:
1- Create Reaction was triggered twice 
Fix:
Moved onClick to the checkbox input 

## Test and Deployment Plan

Go to activity stream, add a reaction,

Expect: 1 createMutation request fired

### DIFF
--- a/src/features/activities/reactions/ReactionCounts.tsx
+++ b/src/features/activities/reactions/ReactionCounts.tsx
@@ -25,7 +25,6 @@ export const ReactionCounts = ({
             color="reaction"
             data-myreaction={!!rg.myReaction}
             key={rg.reaction}
-            myReaction={rg.myReaction}
             onClick={() => {
               rg.myReaction
                 ? deleteReaction(rg.myReaction)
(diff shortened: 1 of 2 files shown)


src/features/activities/reactions/ReactionOptions.tsx
```diff
@@ -91,9 +91,6 @@ export const ReactionOptions = ({
             <ReactionButton
               color="reaction"
               key={r}
-              onClick={() =>
```
crabsinger, line 94: I think its best practice to leave the onClick on the top level button. Otherwise, it's possible you click on the button but the input isn't clicked and so the UI doesn't behave as expected.

So even though this commit fixes it, I think we should understand why and where the other addReaction is being triggered from, and fix this in another way, rather than just moving the onClick to the input. @sshmm

src/features/activities/reactions/ReactionCounts.tsx
```diff
@@ -25,7 +25,6 @@ export const ReactionCounts = ({
             color="reaction"
             data-myreaction={!!rg.myReaction}
             key={rg.reaction}
-            myReaction={rg.myReaction}
```
crabsinger, line 28: @CryptoGraffe I couldn't find any use for this, and it caused a react error, but curious if you recall a reason for it.

crabsinger (approved):

PR #2068, opened 2023-04-04
allow epochs to start 100 days in the past

### LINKED ISSUE #2060: Allow epochs to start in the past
Problem:
Epochs are hard to start immediately. Also, with our current restrictions on Epoch start, it's difficult to create a recurring epoch where the recurring period has already started. 
(e.g. - I want to have a monthly epoch starting on the 1st of the month, but today is the 15th. This means that I have to make a single epoch for today through the end of this month. THEN make a a recurring epoch that starts on the 1st of the next month and repeats.)

Solution:
Allow epochs to start in the past. 
- make sure that all other current rules still apply (like not allowing epochs to overlap).
- limit to no more than 100 days into the past

Image

sshmm:
## Motivation and Context

resolves #2060

## Test and Deployment Plan

Make sure that current unit test passes.
1- created an epoch that starts in the past
2- created a future epoch and periodic epochs that starts in the past to check collision

### DIFF
--- a/src/pages/HistoryPage/EpochForm.tsx
+++ b/src/pages/HistoryPage/EpochForm.tsx
@@ -269,7 +269,7 @@ const getZodParser = (
       },
       {
         path: ['start_date'],
-        message: 'Start date can not be earlier than 100 days',
+        message: 'Start date can not be earlier than 100 days ago',
       }
     )
     .refine(
--- a/src/pages/HistoryPage/EpochForm.tsx
+++ b/src/pages/HistoryPage/EpochForm.tsx
@@ -155,7 +155,7 @@ const getCollisionMessage = (
 ) => {
   if (
     newInterval.overlaps(e.interval) ||
-    (e.repeatEnum === 'none' && !newRepeat)
+    (e.repeatEnum === 'none' && (!newRepeat || newRepeat.type === 'one-off'))
   ) {
     return newInterval.overlaps(e.interval)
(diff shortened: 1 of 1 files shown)


src/pages/HistoryPage/EpochForm.tsx
```diff
@@ -263,13 +263,13 @@ const getZodParser = (
...
         return (
-          start_date > DateTime.now().setZone() ||
+          start_date > DateTime.now().setZone().minus({ days: 100 }) ||
           (currentEpoch && source?.epoch?.id === currentEpoch.id)
         );
       },
       {
         path: ['start_date'],
-        message: 'Start date must be in the future',
+        message: 'Start date can not be earlier than 100 days',
```
levity, line 272:
```suggestion
        message: 'Start date can not be earlier than 100 days ago',
```

src/pages/HistoryPage/EpochForm.tsx
```diff
@@ -263,13 +263,13 @@ const getZodParser = (
...
         return (
-          start_date > DateTime.now().setZone() ||
+          start_date > DateTime.now().setZone().minus({ days: 100 }) ||
           (currentEpoch && source?.epoch?.id === currentEpoch.id)
         );
       },
       {
         path: ['start_date'],
-        message: 'Start date must be in the future',
+        message: 'Start date can not be earlier than 100 days ago',
```
levity, line 272:
```suggestion
        message: 'Start date cannot be earlier than 100 days ago',
```

levity (approved):

PR #2071, opened 2023-04-04
Parallelize sendSocialMessage actions

### LINKED ISSUE #2063: Parallelize actions in sendSocialMessage
Parallelize the different actions in sendSocialMessage and make sure that one of them failing doesn't prevent the rest from occurring

source #2042

sshmm:
## Motivation and Context

resolves #2063

## Description

parallelized the different actions in sendSocialMessage and make sure that one of them failing doesn't prevent the rest from occurring

### DIFF
--- a/api-lib/sendSocialMessage.test.ts
+++ b/api-lib/sendSocialMessage.test.ts
@@ -1,0 +1,59 @@
+import { createCircle } from '../api-test/helpers';
+
+import { adminClient } from './gql/adminClient';
+import { sendSocialMessage } from './sendSocialMessage';
+import { Awaited } from './ts4.5shim';
+
+describe('sendSocialMessage', () => {
+  let circle: Awaited<ReturnType<typeof createCircle>>;
+
+  test('Test Failed Telegram message', async () => {
+    circle = await createCircle(adminClient, {
+      telegram_id: 'coordinapeTest123',
+    });
+    await expect(
+      sendSocialMessage({
+        message: 'Test',
+        circleId: circle.id,
+        channels: { telegram: true },
+        sanitize: false,
+      })
+    ).rejects.toThrowError(/"Bad Request: chat not found"/);
+  });
+
+  test('Test Failed discord webhook message', async () => {
+    circle = await createCircle(adminClient, {
+      discord_webhook: 'https://.com/api/webhooks/coordinapeTest123',
+    });
+    await expect(
+      sendSocialMessage({
+        message: 'Test',
+        circleId: circle.id,
+        channels: { discord: true },
+        sanitize: false,
+      })
+    ).rejects.toThrowError(
+      /request to https:\/\/.com\/api\/webhooks\/coordinapeTest123 failed, reason: getaddrinfo ENOTFOUND/
+    );
+  });
+
+  test('Test Failed discord bot message', async () => {
+    circle = await createCircle(adminClient);
+    await expect(
+      sendSocialMessage({
+        message: 'Test',
+        circleId: circle.id,
+        channels: {
+          isDiscordBot: true,
+          discordBot: {
+            channelId: 'Coordinape123',
+            roleId: 'Test123',
+            type: 'daily-update' as const,
+            message: 'Test',
+          },
+        },
+        sanitize: false,
+      })
+    ).rejects.toThrowError(/"Value \\"Coordinape123\\" is not snowflake."/);
+  });
+});
--- a/api-lib/sendSocialMessage.test.ts
+++ b/api-lib/sendSocialMessage.test.ts
@@ -1,16 +1,48 @@
+import * as nodeFetch from 'node-fetch';
+
 import { createCircle } from '../api-test/helpers';
 
 import { adminClient } from './gql/adminClient';
 import { sendSocialMessage } from './sendSocialMessage';
 import { Awaited } from './ts4.5shim';
 
+const fetchOrig = jest.requireActual('node-fetch').default;
+
+const expectedResponse = {
+  ok: false,
+  status: 404,
+  statusText: 'Not Found',
+  headers: new Headers(),
+  text: jest.fn().mockResolvedValue(''),
+  json: () =>
+    Promise.resolve({
+      success: false,
+      error: 'Unexpected Id',
+    }),
+  buffer: undefined,
+  size: undefined,
+  textConverted: undefined,
+  timeout: undefined,
+} as Response;
+
 describe('sendSocialMessage', () => {
+  const consoleSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
   let circle: Awaited<ReturnType<typeof createCircle>>;
 
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
   test('Test Failed Telegram message', async () => {
     circle = await createCircle(adminClient, {
       telegram_id: 'coordinapeTest123',
     });
+
+    jest
+      .spyOn(nodeFetch, 'default')
+      .mockImplementationOnce(fetchOrig)
+      .mockImplementationOnce(() => Promise.resolve(expectedResponse));
+
     await expect(
       sendSocialMessage({
         message: 'Test',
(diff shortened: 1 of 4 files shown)


crabsinger (changes requested):
@sshmm a few improvements Im imagining:

1. Add a few tests to help us ensure its working as expected.
2. Have each of these `fetch`es moved into a method. The code uses .then() to handle the response and errors, but it doesn't include a .catch() block for proper error handling. Instead of throwing errors inside the .then() block, you should return a rejected promise and handle it in a .catch() block. This ensures that any errors that occur during the fetch process are handled gracefully.

3. Wait for all requests to complete before returning:
The function currently sends requests in parallel but doesn't wait for them to complete before returning. This can lead to unexpected behavior, especially when errors occur. Use Promise.all() to wait for all requests to complete before returning.

Here is a code example of this pattern:

```ts
export async function sendSocialMessage({
  // ...
}) {
  // ...

  const sendDiscordBotMessage = async () => {
    try {
      const { type } = channels.discordBot || {};
      const res = await fetch(`https://coordinape-discord-bot.herokuapp.com/api/epoch/${type}`, {
        method: 'POST',
        body: JSON.stringify(channels.discordBot),
        headers: {
          'Content-Type': 'application/json',
          'x-coordinape-bot-secret': COORDINAPE_BOT_SECRET,
        },
      });

      if (!res.ok) {
        throw new Error(JSON.stringify(await res.json()));
      }
    } catch (error) {
      // Handle error
    }
  };

  const sendDiscordMessage = async () => {
    // ...
  };

  const sendTelegramMessage = async () => {
    // ...
  };

  // Execute all requests in parallel and wait for them to complete
  await Promise.all([
    sendDiscordBotMessage(),
    sendDiscordMessage(),
    sendTelegramMessage(),
  ]);
}

```

api-test/helpers/circles.ts
```diff
@@ -10,6 +10,8 @@ const createCircleSchemaInput = z
   .object({
     name: z.string().min(3).max(255),
     organization_id: z.number().int().positive(),
+    telegram_id: z.string().optional(),
```
crabsinger, line 13: @sshmm do these need to be part of this PR?

crabsinger (dismissed):
ok, i think this looks good to me. @levity any additional feedback you have, or is this what you were imagining around parallelizing this?

api-test/helpers/circles.ts
levity: yeah, they're for the test

api-lib/sendSocialMessage.ts
```diff
@@ -131,23 +135,40 @@ export async function sendSocialMessage({
...
-      }
-    );
+  const requests = [];
 
-    if (!res.ok) {
-      throw new Error(JSON.stringify(await res.json()));
-    }
+  if (channels?.isDiscordBot && channels?.discordBot) {
+    const { type } = channels.discordBot || {};
+    const updateDiscordBot = fetch(`${DISCORD_BOT_EPOCH_URL}${type}`, {
```
levity, line 142: there's quite a bit of duplication in the code (fetch..catch..error..json..) in these three blocks. can you make a function that takes parameters like the URL and a label like "discord-bot", to reduce the duplication?

api-lib/sendSocialMessage.ts
```diff
@@ -176,16 +216,58 @@ export async function sendSocialMessage({
...
+  }
+
+  const responses = await Promise.allSettled(requests);
 
-    if (!res.ok) {
-      throw new Error(JSON.stringify(await res.json()));
-    }
+  const errors = responses.filter(isRejected);
+
+  if (errors.length > 0) {
```
levity, line 258: now, what to do if there are some errors and some successes is an interesting question. i think we should probably not have the whole thing throw an error if we have at least one success.

my thinking here is that having at least one success means at least one message got through. if we were to repeat this job due to having one error, that message would be duplicated.

the tradeoff is, if one of the errors was due to a transient condition, we would miss the opportunity to retry it and have it work the second time.

so there's a product choice here between "too few messages--but at least one success", and "too many messages".

levity (changes requested):

api-lib/sendSocialMessage.ts
levity: after discussion, i think this is probably fine as-is. if we do send duplicate messages to a community in one channel because of errors with another channel, that will at least help raise awareness that there's a problem

api-lib/sendSocialMessage.ts
levity: btw, when you make this function, please use `async() => { try { await ... } catch }` syntax instead of `.then and `.catch`

api-lib/sendSocialMessage.ts
```diff
@@ -176,16 +174,105 @@ export async function sendSocialMessage({
...
+  url: RequestInfo;
+  headers?: HeaderInit;
+  body: BodyInit;
+  label: string;
+  notifyOrg: boolean;
+  channelId?: string;
+  circleId: number;
+}) => {
+  try {
+    const updateWebhook = await fetch(url, {
```
levity, line 228: this shouldn't be called `updateWebhook` anymore because it's general-purpose now; how about just `response` or `resp`?

api-lib/sendSocialMessage.ts
```diff
@@ -176,16 +174,105 @@ export async function sendSocialMessage({
...
+    console.error(`Error sending social messages: ${errorMessages.join('\n')}`);
+    throw new Error(errorMessages.join('\n'));
+  }
+}
+
+const isRejected = (
+  response: PromiseSettledResult<unknown>
+): response is PromiseRejectedResult => response.status === 'rejected';
+
+const fetchUrl = async ({
```
levity, line 210: how about just calling this `update` since that's what it's doing?

api-lib/sendSocialMessage.test.ts
```diff
@@ -0,0 +1,102 @@
...
+    jest.clearAllMocks();
+  });
+  test('Test Failed Telegram message', async () => {
+    circle = await createCircle(adminClient, {
+      telegram_id: 'coordinapeTest123',
+    });
+
+    jest
+      .spyOn(nodeFetch, 'default')
+      .mockImplementationOnce(fetchOrig)
```
levity, line 39: hmm, i don't understand why you're doing this--you shouldn't have the test send a real request

api-lib/sendSocialMessage.ts
```diff
@@ -176,16 +174,105 @@ export async function sendSocialMessage({
...
+    const updateWebhook = await fetch(url, {
+      method: 'POST',
+      body: body,
       headers: {
         'Content-Type': 'application/json',
+        ...headers,
       },
     });
+    if (!updateWebhook.ok) {
+      if (channelId) {
```
levity, line 237: this code is the same as in the `catch` block. won't you print errors twice in this case, since you throw at the end of this block again? maybe it's sufficient to just throw and not log here?

api-lib/sendSocialMessage.ts
```diff
@@ -176,16 +174,105 @@ export async function sendSocialMessage({
...
+      }
 
-    if (!res.ok) {
-      throw new Error(JSON.stringify(await res.json()));
+      const json = await updateWebhook.json();
+      throw new Error(JSON.stringify(json));
     }
+  } catch (err: any) {
+    if (channelId) {
+      console.error(
```
levity, line 261: these two `console.error` lines are identical except for `channelId`. so it could just be one:
```js
console.error('...' + (channelId ? `Channel Id: ${channelId}` : ''));
```

api-lib/sendSocialMessage.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  headers: new Headers(),
+  text: jest.fn().mockResolvedValue(''),
+  json: () =>
+    Promise.resolve({
+      success: false,
+      error: 'Unexpected Id',
+    }),
+} as unknown as nodeFetch.Response;
+
+describe('sendSocialMessage', () => {
```
levity, line 24: can you add a test that sends two, has one fail, and verifies that the other one succeeds?

also, no need to have a `describe` block around all of this if there's only one in the whole test file

levity (dismissed):

api-lib/sendSocialMessage.test.ts
sshmm: as was discussed before we didn't want to send to discord/telegram bad requests just for testing especially if the tests may run multiple times every time someone posts a change

api-lib/sendSocialMessage.ts
sshmm: to log resp.status and resp.status text if the error is thrown when resp.ok = false. that's why I was using .then after catch in the previous commits implementation.

api-lib/sendSocialMessage.test.ts
levity: yes, i agree. but doesn't this line mean "set the mock implementation to the real implementation of fetch"? doesn't that cause a real call to the remote URL?

OHH, i remember now, this is because you actually want the first request to go through, because the first request is generated by the `createCircle` helper, right?

sshmm: > yes, i agree. but doesn't this line mean "set the mock implementation to the real implementation of fetch"? doesn't that cause a real call to the remote URL?
> 
> OHH, i remember now, this is because you actually want the first request to go through, because the first request is generated by the `createCircle` helper, right?

yes, if mocked fetch implementation for all createCircle fetch will be mocked too.

api-lib/sendSocialMessage.ts
levity: maybe you could do
```js
throw New Error(`${update.status} ${update.statusText}`)
```

after all if it's a 404 or 500 then the `.json()` call will probably fail

crabsinger (approved):

PR #2085, opened 2023-04-11
allow org members to see & create reactions

levity:


hasura/metadata/databases/default/tables/public_reactions.yaml
```diff
@@ -26,14 +26,21 @@ insert_permissions:
...
-                  - profile:
-                      id:
+            _or:
+              - members:
+                  _and:
+                    - profile_id:
                         _eq: X-Hasura-User-Id
+                    - deleted_at:
+                        _is_null: true
+              - circles:
```
crabsinger, line 36: once we launch org members, are we going to want to keep this OR or make these only org_member dependent? @levity 

I think just org_member dependent seems simpler... but not until it launches can we make that changes ,right?

crabsinger (approved):

PR #2101, opened 2023-04-18
remove src/lib/zod/index

levity:
Remove src/lib/zod/index.ts

Co-locate input schemas with their handlers instead
Move handler helper methods to api-lib/requests/schema

PR #2102, opened 2023-04-18
improve API input validation helpers

levity:
This PR combines three methods of parsing and validating input to API handlers:
- `composeHasuraActionRequestBodyWithSession`
- `composeHasuraActionRequestBodyWithoutPayload`
- `composeHasuraActionRequestBodyWithApiPermissions`

...into a new method, [`getInput`](https://github.com/coordinape/coordinape/pull/2102/files#diff-6771e63411c8028afe8710457cec5ee257dafca3a1d0a382d4e189e8f720d7d2R100).

(It also combines `getPropsWithUserSession`, an earlier half-finished attempt at the same.)

This method has the same functionality but its output is less nested, it removes an argument that is always the same (HasuraUserSessionVariables) and it uses shorter names (e.g. "session" instead of "sessionVariables", and the method name itself). It is always async, to allow for zod's `parseAsync`.

There is an `allowAdmin` option which adds admin session variables as well as user session variables. This is an option rather than a default because in many cases we know we only want a handler to be used in a user context, so it's nice to be able to typecheck that e.g. `session.hasuraProfileId` exists.

Also, it may seem a bit weird to have API permission checking inside of input value validation (or at least it does to me), but it was like that when I got here 😉 

Please check out `handlerHelpers.test` for some usage examples.


levity:
* main
  * **PR #2101** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/2101" target="_blank">Graphite</a> 
    * **PR #2102** <a href="https://app.graphite.dev/github/pr/coordinape/coordinape/2102" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/coordinape/coordinape/2102?utm_source=stack-comment).

CryptoGraffe (approved):
Great work! Cleans things up a lot. I looked closely for typos and it all looked good to me.

PR #2108, opened 2023-04-20
Visit circle members page directly

crabsinger:
Fix cypress test

THis was broken by ORg Members link being added


wingfeatherwave (approved):

levity:
Ah, cool.

It looks like this also works, if you want to avoid the extra class:
```js
cy.get(`a[href="/circles/${circleId}/members"]`).first().click();
```

crabsinger:
> Ah, cool.
> 
> It looks like this also works, if you want to avoid the extra class:
> ```js
> cy.get(`a[href="/circles/${circleId}/members"]`).first().click();
> ```

I actually did that first, but wanted to click on an element for better testing of a user flow and to ensure that if the "Members" link is broken we would fail this test rather than with the direct visit option.

levity:
oh, i see. could you change it from className to data-testId then to make
the intent clearer?

PR #2109, opened 2023-04-21
View past epoch statements

wingfeatherwave:
Screenshot 2023-04-21 at 11 41 44 AM


src/pages/HistoryPage/getHistoryData.ts
```diff
@@ -111,6 +111,21 @@ export const getHistoryData = async (circleId: number, userId: number) => {
                         },
                       ],
                     },
+                    epochStatements: {
+                      histories: [
+                        {},
+                        {
+                          id: true,
+                          bio: true,
```
crabsinger, line 119: looks like we're not using all these fields, right? can we remove all that arent sued.

crabsinger (changes requested):

crabsinger (approved):

PR #2116, opened 2023-04-26
fix: `null` epoch number

karelianpie:
Return the newly set epoch number in `setNextEpochNumber()`


crabsinger (approved):

crabsinger:
Looks good - but cannot modify your branch to rebase it, so recreating a new PR.

PR #2117, opened 2023-04-26
Bugfix: Filter epoch statements to exclude blank ones

wingfeatherwave:
Here's the bug:
Screenshot 2023-04-26 at 9 50 54 AM
blank ones should't be included

### DIFF
--- a/src/pages/HistoryPage/Notes.tsx
+++ b/src/pages/HistoryPage/Notes.tsx
@@ -24,7 +24,10 @@ export const NotesSection = ({
 }) => {
   const receivedLength = received.filter(g => g.gift_private?.note).length;
   const sentLength = sent?.filter(g => g.gift_private?.note).length;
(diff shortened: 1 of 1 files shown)


levity (approved):

src/pages/HistoryPage/Notes.tsx
```diff
@@ -24,7 +24,10 @@ export const NotesSection = ({
 }) => {
   const receivedLength = received.filter(g => g.gift_private?.note).length;
   const sentLength = sent?.filter(g => g.gift_private?.note).length;
-  const epochStatementsLength = epochStatements?.length;
+  const filteredEpochStatements = epochStatements?.filter(
+    e => e.bio && e.bio.length > 0
```
crabsinger, line 28: I would have actually done this on the DB side in the query. Like this: (untested)

https://github.com/coordinape/coordinape/compare/filter-epoch-statement-list?expand=1

PR #2118, opened 2023-04-26
Minor Bug-fixes: circle row breakpoint & profile nav current

wingfeatherwave:
old / new
Screenshot 2023-04-26 at 11 29 07 AM
simple breakpoint fixes

bug getting fixed for nav:
Screenshot 2023-04-26 at 11 31 26 AM
those links are incorrectly highlighted as they match the `/` path. gave them unique url param.


levity:
> those links are incorrectly highlighted as they match the / path. gave them unique url param.

you can tell me i'm being too picky, but i'd rather not change the URL when these links are clicked. could you try e.g. making NavItem's `to` prop optional, and not rendering as a NavLink when it's not set?

levity (approved):

PR #2128, opened 2023-05-02
Feature flag showing org members

crabsinger:
### <samp>🤖 Generated by Copilot at 810e545</samp>

Add feature flag for organization view feature. Hide `Members` nav item unless the flag is enabled.

### DIFF
--- a/src/features/nav/NavCurrentOrg.tsx
+++ b/src/features/nav/NavCurrentOrg.tsx
@@ -1,12 +1,12 @@
 import { Activity, DollarSign, Member } from '../../icons/__generated';
 import { paths } from '../../routes/paths';
 import { Box } from '../../ui';
+import isFeatureEnabled from 'config/features';
 
 import { NavOrg } from './getNavData';
 import { NavItem } from './NavItem';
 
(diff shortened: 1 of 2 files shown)


src/features/nav/NavCurrentOrg.tsx
```diff
@@ -27,7 +28,7 @@ export const NavCurrentOrg = ({ org }: { org: NavOrg }) => {
         to={paths.vaultsForOrg(org.id)}
         icon={<DollarSign />}
       />
-      {isInOrg && (
+      {isFeatureEnabled('org_view') && isInOrg && (
```
CryptoGraffe, line 31: Does this really work tho? where is the code that says "only show this if there are members" is it the isInOrg check?

CryptoGraffe:
This is the problem tho , it won’t show up unless you already have org
members

wingfeatherwave:
`Members` link in the current org menu is not working.  attempting to fix

wingfeatherwave (changes requested):
`Members` link in the current org menu is not working.  attempting to fix

wingfeatherwave (approved):

PR #2094, opened 2023-04-12
Move rFullCircle-related code into features/map

levity:
## What

the map is now the only page that uses the rFullCircle selector and associated code. so this PR removes references to it from all other parts of the app and adds a fetchCircle query when the map page loads

remove the useFixCircleState hook; it's no longer needed

move the remaining recoil state into recoilState/index

### <samp>🤖 Generated by Copilot at a9d4138</samp>

> _We're refactoring the code, me hearties, yo ho ho_
> _We're using circle id for queries, and recoil for the flow_
> _We're deleting the legacy code, me hearties, yo ho ho_
> _We're heaving on the `paths.home`, on the count of three, let's go_

PR #2087, opened 2023-04-11
move map files into src/features/map

levity:
## Motivation and Context

this makes it a bit easier to see what code is associated with the map

PR #2129, opened 2023-05-02
allow access to map for all circles in org

levity:
- add Map link to side nav for all circles
- move route so org-wide access is permitted

in initial testing, it seems like the map code already works for an org member not in the circle (after the changes in #2087 and #2094)

### DIFF
--- a/src/features/map/AMForceGraph.tsx
+++ b/src/features/map/AMForceGraph.tsx
@@ -5,15 +5,16 @@ import { StitchesTheme } from 'features/theming/ThemeProvider';
 import cloneDeep from 'lodash/cloneDeep';
 import ForceGraph2D, { NodeObject, LinkObject } from 'react-force-graph-2d';
 import AutoSizer from 'react-virtualized-auto-sizer';
+import { useRecoilValue, useSetRecoilState } from 'recoil';
 import { styled } from 'stitches.config';
 
 import { Box } from 'ui';
 
 import {
-  useMapGraphData,
-  useMapContext,
-  useSetAmEgoAddress,
   AmContextDefault,
+  rMapContext,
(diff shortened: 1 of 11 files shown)


sshmm (commented):
I think it would be better to add a button to Map page in non Member circle row 
image

sshmm (commented):
https://github.com/coordinape/coordinape/blob/3e3ab136aa9d1d29f398c5170e348a148782d7e6/src/utils/modelExtenders.ts#L58
May be unrelated to your changes but upon clicking on view profile for deleted users in map page it navigates to /profile/0xFAKE and keeps reloading forever there

src/features/map/AMForceGraph.tsx
```diff
@@ -41,9 +42,9 @@ export const AMForceGraph = ({
   stitchesTheme: StitchesTheme;
 }) => {
   const fgRef = useRef<any>(null);
```
sshmm, line 44: is fgRef still needed after removing
`if (!fgRef.current) {
        return;
      }` ?

levity: ah, good point, it's not in use anymore i see. i'll add a comment about it in case we want to use it in the future

levity:
> https://github.com/coordinape/coordinape/blob/3e3ab136aa9d1d29f398c5170e348a148782d7e6/src/utils/modelExtenders.ts#L58
> 
> 
> May be unrelated to your changes but upon clicking on view profile for deleted users in map page it navigates to /profile/0xFAKE and keeps reloading forever there

good catch. i'll remove that link for deleted users

levity:
> I think it would be better to add a button to Map page in non Member circle row image

yep, just added this to #2131 :)

PR #2132, opened 2023-05-03
Dismissible banner for Org Member adding

wingfeatherwave:
Screenshot 2023-05-03 at 4 27 24 PM

Modified our `HintBanner` to accept a `dismissible` prop.  If present it will give the option for the user to close the banner and never see it again, unless they clear their local storage.  

There is also a promo banner on the home page, but it doesn't have the links or button cta, and doesn't show up if you just have the sample org.


src/components/HintBanner.tsx
```diff
@@ -1,36 +1,64 @@
...
-        {title && (
-          <Text h2 as="h2">
-            {title}
-          </Text>
-        )}
-        {children}
-      </Flex>
-    </Panel>
+    <>
+      {localStorageBannerVisibility && (
```
levity, line 33: since there's just a single boolean test inside these fragment tags, i would probably rewrite this as:

```js
if (!localStorageBannerVisibility) return null;

return (
  <Panel ...
```

Also, i think you mean `bannerVisibility` here? I guess it's a little funny because it works with either one. but you can't just drop the `useState` entirely or the component won't know it should re-render

levity (dismissed):
the thing i pointed out in my other comment is not a big deal so feel free to merge anyway if you want

src/components/HintBanner.tsx
wingfeatherwave: Thanks.  I changed the boolean test based on your comment.  I like that.

src/components/HintBanner.tsx
```diff
@@ -1,25 +1,51 @@
...
+  dismissible,
   type = 'info',
   children,
 }: {
   title?: string;
+  dismissible?: string;
   type?: 'info' | 'alert';
   children: React.ReactNode;
 }) => {
+  const dismissibleAs = `banner:${dismissible}` || '';
```
crabsinger, line 18: if dismissible is undefined, this will be `banner:` || '', so the '' will never be touched, right?

crabsinger (approved):

CryptoGraffe (approved):

PR #2138, opened 2023-05-04
Org name validation

sshmm:
### <samp>🤖 Generated by Copilot at 52e0994</samp>

This pull request adds a new feature to prevent name duplication among coordinape users and a new test case to verify its behavior. It modifies the `createOrgMembers` action handler to check for existing profile names before adding new org members, and returns an error if any name conflict is found. It also updates the `createOrgMembers.test.ts` file to include a scenario where two users have the same name but different addresses.

### <samp>🤖 Generated by Copilot at 52e0994</samp>

> _`createOrgMembers`_
> _checks names before adding them_
> _fall leaves no duplicates_

## Why

To reject duplicate names when adding new profiles for new org members

## Test and Deployment Plan

tested by createOrgMembers.test.ts

### <samp>🤖 Generated by Copilot at 52e0994</samp>

*  Prevent name duplication among coordinape users by checking existing profiles before creating org members ([link](https://github.com/coordinape/coordinape/pull/2138/files?diff=unified&w=0#diff-94334cc1c7f4caa25dc329463606e8404f55b378e1610eae5b7d5d8c20f78863R89-R123))
* Add a test case to verify the name conflict logic in the `createOrgMembers` action handler ([link](https://github.com/coordinape/coordinape/pull/2138/files?diff=unified&w=0#diff-9619bc278f9f361a18f4962a96937ae35c2eff2b3816183b552b6b5387a52893R250-R315))


api/hasura/actions/_handlers/createOrgMembers.ts
```diff
@@ -86,6 +86,41 @@ async function handler(req: VercelRequest, res: VercelResponse) {
...
+  const { profiles: existingNames } = await adminClient.query(
+    {
+      profiles: [
+        {
+          where: {
+            _or: users.map(user => {
+              return {
+                _and: [
+                  { name: { _eq: user.name } },
+                  { address: { _nilike: user.address } },
```
crabsinger, line 99: if the address DOES NOT match, but the name does, we still have an error though right? @sshmm

crabsinger: should this just be to check if any profile name is already in use, regardless of address?

crabsinger: nevermind - this is correct.

crabsinger (approved):
nice work on this @sshmm !

api/hasura/actions/_handlers/createOrgMembers.ts
sshmm: May be that user already has the a profile for this address so if the name and address matches the profile it won't throw an error.
What it does is that it fetches  all profiles with names from the input as long as these names are assigned for profiles with different addresses from the input addresses.

PR #2159, opened 2023-05-24
cosoul public view

crabsinger:
## What

Create public page to view any minted cosoul.

- creates a serverless function that allows unauthenticated access to minted cosoul data, with a 5 min vercel cache
- setup UI on page

### DIFF
--- a/api/cosoul/[address].ts
+++ b/api/cosoul/[address].ts
@@ -1,0 +1,236 @@
+import assert from 'assert';
+
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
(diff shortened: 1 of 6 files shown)


crabsinger:
@CryptoGraffe @wingfeatherwave ready for review. I refactored this to use a simpler, and more common useQuery pattern for fetching the data rather than the useEffect hook pattern. I put some sampler error handling (if you put in an invalid URL) on the page.

@wingfeatherwave not sure if there's any cleanup you still want to do around the "Connect" button and the funkiness there?

levity (approved):

PR #2180, opened 2023-06-07
point at newly deployed contracts that prevent transfer

CryptoGraffe:
## What
Updates the contracts in app to point to one that doesn't allow token transfer


hardhat/dist/hardhat.config.js
```diff
@@ -189,6 +189,7 @@ const config = {
             },
             deploy: ['./scripts/deploy/03-cosoul/'],
             live: true,
+            gasPrice: 35000000000,
```
crabsinger, line 192: hmm, im kinda confused by this...

PR #2182, opened 2023-06-07
update jsx-a11y

crabsinger:


CryptoGraffe (commented):
Why does this only update yarn.lock but not package.json?

CryptoGraffe:
Same concern as before - why package.json not updated? whatsup here

crabsinger:
@CryptoGraffe i fixed it. look now.

CryptoGraffe (approved):

PR #2228, opened 2023-06-23
Profile avatars glow if cosoul minted

wingfeatherwave:
Screenshot 2023-06-23 at 11 41 27 AM

### <samp>🤖 Generated by Copilot at 96958d8</samp>

No summary available (Limit exceeded: required to process 67256 tokens, but only 50000 are allowed per call)

### <samp>🤖 Generated by Copilot at 207dacb</samp>

> _`address` prop flows_
> _through components and queries_
> _coSoul glows in spring_

### <samp>🤖 Generated by Copilot at 96958d8</samp>

No walkthrough available (Limit exceeded: required to process 67256 tokens, but only 50000 are allowed per call)


src/pages/HistoryPage/HistoryPage.test.tsx
```diff
@@ -73,7 +73,13 @@ jest.mock('./getHistoryData', () => {
             {
               id: 4,
               tokens: 11,
-              recipient: { profile: { avatar: 'bob.jpg', name: 'Bob' } },
+              recipient: {
+                profile: {
+                  avatar: 'bob.jpg',
+                  name: 'Bob',
+                  address: '0x9999999999999',
```
crabsinger, line 80: if this is not provided, shouldn't this address be optional and not needed?

hasura/metadata/databases/default/tables/public_cosouls.yaml
```diff
@@ -17,6 +17,4 @@ select_permissions:
         - synced_at
         - token_id
         - updated_at
-      filter:
-        profile_id:
-          _eq: X-Hasura-User-Id
+      filter: {}
```
crabsinger, line 20: @wingfeatherwave this change would make this table visible to everyone logged in, i think thats ok but a bit of a change.

src/ui/Avatar/Avatar.tsx
```diff
@@ -140,13 +147,26 @@ export const Avatar = ({
 }) => {
   const avatarPath = getAvatarPath(path);
 
+  const { data: coSoul } = useQuery(
+    [QUERY_KEY_PROFILE_TOTAL_PGIVE, address],
+    () => queryProfilePgive(address),
```
crabsinger, line 152: hmm, do we make a ton of queries for this now? prob fine but might be nice to list as a tech debt improvement to try and batch queries

crabsinger: @CryptoGraffe what do you think about this?

Screenshot 2023-06-26 at 11 22 19 AM

crabsinger (commented):
some thoughts, and asking @CryptoGraffe for some input as well

hasura/metadata/databases/default/tables/public_cosouls.yaml
CryptoGraffe: this is good

src/ui/Avatar/Avatar.tsx
CryptoGraffe: Can we rather join in the cosoul data whenever we fetch the avatar info instead of doing these extra queries?

CryptoGraffe (approved):

PR #2255, opened 2023-07-05
fix repeating full month epoch end date

sshmm:
Determined the epoch start weekday based on the timezone instead of UTC. Used offset DateTime.offset to determine the offset from UTC. Added the offset in hours plus 1 minute to make sure the day is correct Subtracted in case of negative time zone
Removed the offset after start of the next month calculations

## What

copilot:summary

copilot:poem

## Why

The start day is different based on the timezone.
Timezone information is not sent to back-end in the start_date object but in a separate parameter

## Test and Deployment Plan

used different timezone to create a monthly epoch starting at the fourth of July.

## How

copilot:walkthrough


CryptoGraffe (changes requested):
This change seems good but I think it needs to be covered by a test - one that fails before this fix and demonstrates the problem.

CryptoGraffe (approved):

PR #2270, opened 2023-07-25
update new member name from DB

sshmm:
### <samp>🤖 Generated by Copilot at b0cd5b7</samp>

This pull request adds a feature to the Add Members page that allows fetching and displaying profile names by address. It implements a new API endpoint `/api/profileName/:profileAddress` that queries the database and caches the results. It also updates the `NewMemberEntry` and `NewMemberList` components to use the endpoint and validate the form data.

### <samp>🤖 Generated by Copilot at b0cd5b7</samp>

> _We are the rebels of the web, we defy the central power_
> _We use the blockchain to create our own identity_
> _We fetch the profile names by address, we validate the data_
> _We use the hooks and the cache to optimize our speed_

## Why

When the admin is trying to add a member with existing profile with a name, the name should be fetched from DB and displayed to the user.

## Test and Deployment Plan
1- Add user with existing name in DB by address or by CSV 
2- choose different names from existing names
expected:
the stored name will replace the entered name and the name input will be disabled for editing

### <samp>🤖 Generated by Copilot at b0cd5b7</samp>

*  Add a new API endpoint `/api/profileName/:profileAddress` that returns the name of a user profile given an Ethereum address ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-f2faf8a6659c00b5401fb976d39bd3c001d0587f68cc8984f3884c6f7dbba220R1-R59))
  * Use a GraphQL query to fetch the profile from the database and cache the response for 5 minutes ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-f2faf8a6659c00b5401fb976d39bd3c001d0587f68cc8984f3884c6f7dbba220R1-R59))
  * Validate the address and handle errors using `@ethersproject/address` and `next-connect` libraries ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-f2faf8a6659c00b5401fb976d39bd3c001d0587f68cc8984f3884c6f7dbba220R1-R59))
  * Import the API handler `profileAddress` to `scripts/serve_dev.ts` and add a new route `/api/profileName/:profileAddress` to the Express app ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-dd65c0157396f6335fc1f86164d9ffeb75c19d3c2e7fb5ebcd684e9f1a41f5b7R27), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-dd65c0157396f6335fc1f86164d9ffeb75c19d3c2e7fb5ebcd684e9f1a41f5b7L89-R92))
* Update the `NewMemberEntry` component to fetch the profile name of the address entered by the user and set the value of the name input accordingly ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cL1-R12), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR19-R29), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR36-R37), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cL22-R81), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR91))
  * Add new imports for React hooks, `@ethersproject/address`, `zod`, and `react-hook-form` libraries ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cL1-R12))
  * Define a `FormValuesSchema` and a `FormValues` type to validate the shape of the form data ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR19-R29))
  * Add `control` and `setValue` props to the component and pass them from the parent component `NewMemberList` ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR36-R37), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-1069257ea6b0200ecd79cb31093f737edec0d134fab4b490ae4336b4ead1faaeR98), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-1069257ea6b0200ecd79cb31093f737edec0d134fab4b490ae4336b4ead1faaeR235-R236))
  * Use the `useController` and `useEffect` hooks to access the `address` field state and trigger a side effect when the address value or error changes ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cL22-R81))
  * Call the new API endpoint `/api/profileName/:profileAddress` with the address value and parse the response as JSON ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cL22-R81))
  * Set the value of the `name` field and disable the name input if the response contains a name, or re-enable the name input and set `isFetched` to false otherwise ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cL22-R81), [link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR91))
  * Add a `disabled` prop to the `TextField` component that renders the name input and set it to the value of `isFetched` ([link](https://github.com/coordinape/coordinape/pull/2270/files?diff=unified&w=0#diff-39828da3b1c60dc64d454fa58f16929b9d1e3241e1234edef7a266002af8249cR91))


api/ProfileName/[profileAddress].ts
```diff
@@ -0,0 +1,59 @@
+import assert from 'assert';
```
Comment (crabsinger) [inline api/ProfileName/[profileAddress].ts:1]: we usually dont capitalize directories like ProfileName, and just have it be lower case.

api/ProfileName/[profileAddress].ts
```diff
@@ -0,0 +1,59 @@
...
+
+import { adminClient } from '../../api-lib/gql/adminClient';
+import { errorResponse, NotFoundError } from '../../api-lib/HttpError';
+
+const CACHE_SECONDS = 60 * 5;
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  try {
+    let address: string | undefined;
+    if (typeof req.query.profileAddress == 'string') {
+      address = req.query.profileAddress;
```
Comment (crabsinger) [inline api/ProfileName/[profileAddress].ts:14]: is there a use case for the address to be wrapped in an array? Or should we simplify and expect the profileAddress to always just be in the first `if` case?

api/ProfileName/[profileAddress].ts
```diff
@@ -0,0 +1,59 @@
...
+    } else if (Array.isArray(req.query.address)) {
+      address = req.query.profileAddress.pop();
+    }
+
+    if (!address || !isAddress(address)) {
+      throw new NotFoundError('no valid address provided');
+    }
+
+    address = address.toLowerCase();
```
Comment (crabsinger) [inline api/ProfileName/[profileAddress].ts:23]: we're doing an ilike in the query, so this isn't actually necessary, right? doesn't hurt to do I guess...

api/ProfileName/[profileAddress].ts
```diff
@@ -0,0 +1,59 @@
...
+    {
+      profiles: [
+        {
+          where: {
+            address: { _ilike: address },
+          },
+          limit: 1,
+        },
+        {
+          id: true,
```
Comment (crabsinger) [inline api/ProfileName/[profileAddress].ts:45]: are we using the `id` ? if not, lets not fetch it.

src/pages/AddMembersPage/NewMemberEntry.tsx
```diff
@@ -29,6 +88,7 @@ const NewMemberEntry = ({
             fullWidth
             error={error?.name ? true : undefined}
             autoComplete={'off'}
+            disabled={isFetched}
```
crabsinger, line 91: image

the field has state disabled, but it looks exactly the same as the other Name fields, which is confusing. I think we should make it disabled and have a grey background so its clear that field is not editable @wingfeatherwave

api/ProfileName/[profileAddress].ts
```diff
@@ -0,0 +1,59 @@
+import assert from 'assert';
+
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+import { isAddress } from 'ethers/lib/utils';
+
+import { adminClient } from '../../api-lib/gql/adminClient';
+import { errorResponse, NotFoundError } from '../../api-lib/HttpError';
+
+const CACHE_SECONDS = 60 * 5;
+export default async function handler(req: VercelRequest, res: VercelResponse) {
```
Comment (crabsinger) [inline api/ProfileName/[profileAddress].ts:10]: this handler doesn't have any permissions on it, and so this will expose ALL user names for all addresses in our DB.

I think we should just use a normal graphql query to request this data and use the logged in users auth token to authenticate the request. Is there any reason this should be a serverless function?

crabsinger (changes requested):
@sshmm left a few comments, i have some refactoring that I think needs to be done before shipping this:

also seeing this console error which we should check is not happening after your refactoring:

image

api/ProfileName/[profileAddress].ts
Comment (sshmm) [reply api/ProfileName/[profileAddress].ts:14]: I agree, I will fix that

api/ProfileName/[profileAddress].ts
Comment (sshmm) [reply api/ProfileName/[profileAddress].ts:10]: the name already exposed in getCosoulData from api/cosoul/[address] so it's not a sensitive data anymore and I didn't want to change the permission of the column

Comment (sshmm) [reply api/ProfileName/[profileAddress].ts:10]: I will replace it with hasura action

Comment (sshmm) [reply api/ProfileName/[profileAddress].ts:10]: I will replace it with hasura action

Comment (sshmm) [reply api/ProfileName/[profileAddress].ts:10]: > I will replace it with hasura action
For type-safety and permission

Comment (crabsinger) [reply api/ProfileName/[profileAddress].ts]: I don't think we need an action - just use the normal graphql interface exposed by hasura, right @sshmm

src/pages/AddMembersPage/NewMemberEntry.tsx
wingfeatherwave: fixed

api/ProfileName/[profileAddress].ts
Comment (sshmm) [reply api/ProfileName/[profileAddress].ts]: @crabsinger thank you. I removed the action and used a query instead

scripts/serve_dev.ts
```diff
@@ -86,7 +86,6 @@ app.get('/api/cosoul/art/:artTokenId', (req, res) => {
 app.get('/api/cosoul/metadata/:tokenId', (req, res) => {
   return tf(tokenId)({ ...req, query: req.params }, res);
 });
-
```
crabsinger, line 89: no need for this change

hasura/schema/admin/schema.graphql
```diff
@@ -6250,6 +6286,7 @@ input circles_variance_order_by {
   min_vouches: order_by
   nomination_days_limit: order_by
   organization_id: order_by
+  starting_tokens: order_by
```
crabsinger, line 6289: these are for the other PR, not this one right?

src/pages/AddMembersPage/NewMemberEntry.tsx
```diff
@@ -1,25 +1,99 @@
...
+                { name: true },
+              ],
+            },
+            { operationName: 'NewMemberEntry_getUserName' }
+          );
+          assert(profiles, 'failed to fetch user profiles');
+
+          const name = profiles[0]?.name ?? '';
+          if (name.length > 0) {
+            setValue(`newMembers.${index}.name`, name);
```
crabsinger, line 77: @sshmm when we do `setValue` we also need to clear out any errors on this form field...

crabsinger (changes requested):
@sshmm nice improvements, using the graphql api and simplifying things.

A few issues remain:
- you are not clearing the error state when programatically setting the field when a async request returns successfully, for both nomination and members list.
- additionally, it's a bit weird that we auto complete the name even if the user is already a member of the circle... this seems to indicate the form entry is valid, but it is not because a user with address already exists in the circle.

I'm not sure if we want to expand the scope on this task to include this (cc @ashmbanks ) but I do think this form would have better UX if when we fetch the username, we also make sure the given address is NOT ALREADY a member of the circle, and if it is already a member we communicate that and remove it so the form doesn't fail on submission.

src/pages/AddMembersPage/NewMemberEntry.tsx
sshmm: done

crabsinger (approved):
lgtm - nice work on this @sshmm

PR #2299, opened 2023-08-18
automatic generation of CoSoul static images

CryptoGraffe:
* Update CoSoul screenshot/static image on mint and when pgive changes.
* Reduce batch size of cosoul cron processing because screenshotting takes a lot of time. 
* Add support for lambda/vercel friendly chromium so that we can screenshot on the server side.
* New /api/cosoul/screenshot/{tokenId} api for storing a screenshot on the server side


api/cosoul/metadata/[tokenId].ts
```diff
@@ -118,12 +118,12 @@ async function getCosoulMetaData(tokenId: number) {
...
 
+  const s3bucket = process.env.IMAGES_AWS_BUCKET || 'coordinape';
+
   return {
     description: description,
     external_url: external_url,
-    //TODO: Update this placeholder image
-    image:
-      'https://coordinape-prod.s3.amazonaws.com/assets/static/images/cosoul-thumb.png',
+    image: `https://${s3bucket}.s3.amazonaws.com/cosoul/${tokenId}.png`,
```
Comment (crabsinger) [inline api/cosoul/metadata/[tokenId].ts:126]: lets put these in `/assets/static/cosoul` or something a bit more nested rather than on the root

crabsinger (approved):
@CryptoGraffe approved - couldn't get local working but staging works

PR #2300, opened 2023-08-19
add ability to hide sidenav orgs

sshmm:
## What

copilot:summary

copilot:poem

## Why

add the ability to hide orgs from sidenav

image
image

## How

copilot:walkthrough

### DIFF
--- a/src/features/auth/useSavedAuth.ts
+++ b/src/features/auth/useSavedAuth.ts
@@ -35,6 +35,7 @@ const getAllData = (): IAuth => {
 
 // only exporting for testing purposes
 export const saveAllData = (allData: IAuth) => {
+  localStorage.removeItem('hiddenOrgs');
   localStorage.setItem(AUTH_STORAGE_KEY, JSON.stringify(allData));
 };
 
--- a/src/features/nav/NavOrgs.tsx
+++ b/src/features/nav/NavOrgs.tsx
@@ -1,12 +1,13 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 
 import { NavLink } from 'react-router-dom';
 
-import { ChevronRight, PlusCircle } from '../../icons/__generated';
+import { ChevronRight, EyeOff, PlusCircle } from '../../icons/__generated';
 import { paths } from '../../routes/paths';
 import {
   Avatar,
   Box,
+  Button,
   Collapsible,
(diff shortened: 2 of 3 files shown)


src/features/auth/useSavedAuth.ts
```diff
@@ -35,6 +35,7 @@ const getAllData = (): IAuth => {
 
 // only exporting for testing purposes
 export const saveAllData = (allData: IAuth) => {
+  localStorage.removeItem('hiddenOrgs');
```
crabsinger, line 38: what triggers saveAllData ?

src/features/nav/SideNav.tsx
```diff
@@ -192,6 +195,7 @@ export const SideNav = () => {
               orgs={data.organizations}
               currentOrg={currentOrg}
               currentCircle={currentCircle}
+              onHideCurrentOrg={hideCurrentOrgHandler}
```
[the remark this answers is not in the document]
crabsinger: id just use a inline function for this

crabsinger (commented):
The URL doesn't change when you hide the current org, i think it should for URL consistency.

src/features/nav/NavOrgs.tsx
```diff
@@ -107,18 +123,46 @@ export const NavOrgs = ({
...
+    : [];
+  const [hiddenOrgs, setHiddenOrgs] = useState<number[]>(initialHiddenOrgs);
+
+  const hideOrgsHandler = (orgId: number) => {
+    setHiddenOrgs(prevHiddenOrgs => [...prevHiddenOrgs, orgId]);
+  };
+
+  useEffect(() => {
+    localStorage.setItem('hiddenOrgs', JSON.stringify(hiddenOrgs));
+    if (hiddenOrgs.includes(currentOrg?.id)) onHideCurrentOrg();
```
crabsinger, line 146: if I navigate directly to an org, via URL like http://localhost:3000/organizations/18 then I see the org, but I don't see it in the sidenav. So this whole pattern seems confusing.

Maybe you should not be able to hide current orgs, and only in the menu hide them. Also seems like maybe their should be a confirmation of some sort, its easy to click accidentally.
image

crabsinger (commented):
@sshmm left some feedback

crabsinger:
@sshmm can this be closed now?

PR #2303, opened 2023-08-24
fix add members from org user exists issue

sshmm:
## What

copilot:summary

copilot:poem

## Why
After adding members to a circle using "Add from Org' options if the user tried to add other members without refreshing an error indicating that the user is still trying to add the previously added user is thrown.
solution:
clear the list of members to add after submitting

## How

copilot:walkthrough


crabsinger (approved):
@sshmm nice work!

PR #2413, opened 2023-11-08
get users emails from magic

sshmm:
## What

copilot:summary

copilot:poem

## Why

add a cron to get magic users  emails and set them as primary emails

## How

copilot:walkthrough


CryptoGraffe:
Thanks for doing this sshmm. I want to think about how we want to fetch - maybe not with cron. I'm deep in other stuff now but we can think/discuss next week.

api/hasura/cron/updateMagicEmails.ts
```diff
@@ -0,0 +1,62 @@
+import { Magic } from '@magic-sdk/admin';
+import type { VercelRequest, VercelResponse } from '@vercel/node';
+
+import { adminClient } from '../../../api-lib/gql/adminClient';
+import { errorLog } from '../../../api-lib/HttpError';
+import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
+
+const API_KEY = process.env.SECRET_MAGIC_API_KEY;
```
CryptoGraffe, line 8: can you rename this to MAGIC_SECRET_API_KEY

api/hasura/cron/updateMagicEmails.ts
```diff
@@ -0,0 +1,62 @@
...
+  //check if names are used by other coordinape users
+  const magic = await Magic.init(API_KEY);
+  magic.users.getMetadataByPublicAddress;
+  const { profiles } = await adminClient.query(
+    {
+      profiles: [
+        {
+          where: {
+            _and: [{ connector: { _eq: 'magic' } }, { _not: { emails: {} } }],
+          },
```
CryptoGraffe, line 20: Can you add a limit here so this doesnt time out the 60s vercel. Maybe 20?

hasura/metadata/cron_triggers.yaml
```diff
@@ -110,3 +110,12 @@
     - name: verification_key
       value_from_env: HASURA_EVENT_SECRET
   comment: syncs cosouls in db and sets pgive on chain
+- name: updateMagicEmails
+  webhook: '{{HASURA_API_BASE_URL}}/cron/updateMagicEmails'
+  schedule: 0 * * * *
```
CryptoGraffe, line 115: Since we will batch 20 at a time, just change this to run every minute * * * * *

crabsinger (changes requested):
@sshmm nice work on this so far.

This is great for getting emails for all old users, but we also want to get emails for new users immediately after they join so we DO NOT prompt them to add their email, since we should always get it.

So I think we should do the following:

When new users login with magic, we can see it https://github.com/coordinape/coordinape/blob/main/api/login.ts#L43 here and then in that file we should call a function getMagicEmail or something and this function should be the same shared code as what you wrote already, but it should fetch the email and set it for just for the one user who is logging in.

So then we have cron for all the old emails
and a system for getting the email immediately on login for new users.

Does that make sense?

CryptoGraffe:
@crabsinger @sshmm  before we deploy this we need to set the MAGIC_SECRET_API_KEY for our production Magic app in vercel. Do you have access to this?

CryptoGraffe (approved):

PR #2417, opened 2023-11-09
Mint CoSoul in the Wizard Flow

wingfeatherwave:
Screenshot 2023-11-09 at 2 44 17 PM


crabsinger (approved):
@wingfeatherwave looks like a great start

The CoSoul mint animation and overlay will need some adjusting to make sense within the CoLinks Wizard flow. I also think the "ethcing pgive on blockchain" thing is maybe less relevant now as well. This wizard is looking really nice so far!

PR #2431, opened 2023-11-15
Add rep data to cosoul display

wingfeatherwave:
Screenshot 2023-11-14 at 4 41 33 PM

Will work on better linking later


src/features/cosoul/CoSoulCompositionRep.tsx
```diff
@@ -0,0 +1,506 @@
...
+                  ...nodeLineStyle,
+                  right: '99.5%',
+                  rotate: '8deg',
+                  transformOrigin: '100% 0',
+                },
+              }}
+            >
+              {!coSoulMinted && (
+                <Text
+                  ref={nodeScramble3}
```
CryptoGraffe, line 301: looks like you used node3 twice is that intended? 6 nodes for 8 attributes?

src/features/cosoul/CoSoulCompositionRep.tsx
```diff
@@ -0,0 +1,506 @@
...
+                  ...nodeLineStyle,
+                  left: '99.5%',
+                  rotate: '8deg',
+                  transformOrigin: '0 0',
+                },
+              }}
+            >
+              {!coSoulMinted && (
+                <Text
+                  ref={nodeScramble6}
```
CryptoGraffe, line 424: node6 used twice also, maybe ok?

CryptoGraffe (approved):

PR #2478, opened 2023-11-29
Task cards

wingfeatherwave:
Screenshot 2023-11-29 at 2 04 34 PM

Goal: nudge people to add more rep and buy links

task cards for:
rep below # (currently set to <51)
user hasn't purchased anyone other links

Showing these on the activity page and on the current user profile page.

### DIFF
--- a/src/pages/colinks/CoLinksTaskCards.tsx
+++ b/src/pages/colinks/CoLinksTaskCards.tsx
@@ -1,0 +1,160 @@
+import { QUERY_KEY_COLINKS } from 'features/colinks/wizard/CoLinksWizard';
+import { client } from 'lib/gql/client';
+import { useQuery } from 'react-query';
+import { NavLink } from 'react-router-dom';
+
+import { coLinksPaths } from 'routes/paths';
+import { Button, Flex, Panel, Text } from 'ui';
+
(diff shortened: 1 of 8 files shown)


src/pages/colinks/CoLinksTaskCards.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { coLinksPaths } from 'routes/paths';
+import { Button, Flex, Panel, Text } from 'ui';
+
+export const CoLinksTaskCards = ({
+  currentUserAddress,
+}: {
+  currentUserAddress: string;
+}) => {
+  const { data: myProfile } = useQuery(
+    [QUERY_KEY_COLINKS, currentUserAddress, 'taskRep'],
```
crabsinger, line 15: if i buy a link, does this cache get invalidated automatically?

usually the variable in query key is at the end, but it depends on exactly how we want to invalidate and refresh these i guess: https://tanstack.com/query/v4/docs/react/guides/query-keys @wingfeatherwave

crabsinger: seems to be working fine, so appriving

crabsinger (approved):

PR #2539, opened 2023-12-11
Member item style improvements, profile header improvements

CryptoGraffe:
## What

copilot:summary

copilot:poem

## How

copilot:walkthrough


src/lib/zod/formHelpers.ts
```diff
@@ -60,11 +60,13 @@ export const zUsername = z
 
 export const zDescription = z.string().min(3).max(160);
 
-const urlRegex = /^https?:\/\/([\w\d-]+\.){1,}[\w\d-]+(\/[\w\d-]+)*\/?$/;
+const url = z
```
crabsinger, line 63: ?? is z.url the same?

CryptoGraffe: its bettter, it tolerates query strings and dots and stuff

PR #2592, opened 2023-12-16
Add recent txs

crabsinger:
## What

copilot:summary

copilot:poem

## How

copilot:walkthrough


src/features/colinks/CoLinksNavProfile.tsx
```diff
@@ -51,6 +53,9 @@ export const CoLinksNavProfile = ({
         width: '100%',
       }}
     >
+      {showTxModal && (
```
crabsinger, line 56: Not sure where this should be nested in the dom

CryptoGraffe (approved):

PR #2678, opened 2024-01-04
Dont use public api keys

crabsinger:
## What

- [x] - set in vercel


sshmm (commented):
I think you should do the same for the rest of the environment variables.
They are all used in api-lib/provider.ts @crabsinger

PR #2696, opened 2024-01-14
Node20

CryptoGraffe:


scripts/serve.sh
```diff
@@ -18,7 +18,8 @@ until curl -s -o/dev/null http://localhost:$PROXY_PORT; do
 done
 
 if [ "$COVERAGE" ]; then
-  $BIN/nyc --silent $BIN/ts-node --swc scripts/serve_dev.ts $PORT $PROXY_PORT & API_PID=$!
+#  $BIN/nyc --silent tsx scripts/serve_dev.ts $PORT $PROXY_PORT & API_PID=$!
```
CryptoGraffe, line 21: This is the remaining issue here.....

PR #2699, opened 2024-01-15
Oneclick unsubscribe

crabsinger:
## What
Create one click link to unsubscribe from receiving emails:
1- generate a token based on profile Id, email and email type
2- generate a link with the token and send it to post mark
3- adjust postmark template
4- create a landing page for the unsubscribe link

## Why

Allow users to easily unsubscribe from emails

image
image

## Before Merging
The production templates on postmark needs to be updated as well
1- A new layout needs to be updated "CoLinks with one click unsubscribe"
this layout will have the ubsubscribe link at the footer
image

2- CoLinks Unread Notifications template needs to use this new layout instead.

### DIFF
--- a/api-lib/email/unsubscribe.ts
+++ b/api-lib/email/unsubscribe.ts
@@ -3,34 +3,40 @@ import { createHmac } from 'crypto';
 import { HMAC_SECRET } from '../config';
 import { UnauthorizedError } from '../HttpError';
 
-export function genToken(profileId: string, email: string): string {
-  const token = genHmac(profileId, email);
+export function genToken(
+  profileId: string,
+  email: string,
+  emailType: string
+): string {
+  const token = genHmac(profileId, email, emailType);
 
-  const params = new URLSearchParams({ profileId, email, token });
+  const params = new URLSearchParams({ profileId, email, emailType, token });
   return params.toString();
 }
 
 export function decodeToken(encodedString: string): {
   profileId: string;
   email: string;
+  emailType: string;
 } {
   const params = new URLSearchParams(encodedString);
   const profileId = params.get('profileId');
   const email = params.get('email');
   const token = params.get('token');
+  const emailType = params.get('emailType');
 
-  if (!profileId || !email || !token) {
+  if (!profileId || !email || !token || !emailType) {
     throw new UnauthorizedError('Invalid unsubscribe token');
   }
 
-  const generatedToken = genHmac(profileId, email);
+  const generatedToken = genHmac(profileId, email, emailType);
   if (token !== generatedToken) {
     throw new UnauthorizedError('Invalid unsubscribe token');
   }
-  return { profileId, email };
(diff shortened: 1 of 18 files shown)


api/email/unsubscribe/[unsubscribeToken].ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { decodeToken, isEmailType } from '../../../api-lib/email/unsubscribe';
+import { adminClient } from '../../../api-lib/gql/adminClient';
+import { NotFoundError, errorResponse } from '../../../api-lib/HttpError';
+
+export default async function handler(req: VercelRequest, res: VercelResponse) {
+  let unsubscribeToken: string | undefined;
+  if (typeof req.query.unsubscribeToken == 'string') {
+    unsubscribeToken = req.query.unsubscribeToken;
+  } else if (Array.isArray(req.query.unsubscribeToken)) {
```
Comment (crabsinger) [inline api/email/unsubscribe/[unsubscribeToken].ts:13]: do we need this? should never be an array, right?

api-lib/email/unsubscribe.ts
```diff
@@ -0,0 +1,47 @@
...
+  const profileId = params.get('profileId');
+  const email = params.get('email');
+  const token = params.get('token');
+  const emailType = params.get('emailType');
+  if (!profileId || !email || !token || !emailType || !isEmailType(emailType)) {
+    throw new UnauthorizedError('Invalid unsubscribe token');
+  }
+
+  const generatedToken = genHmac(profileId, email, emailType);
+  if (token !== generatedToken) {
```
crabsinger, line 38: use https://nodejs.org/dist/latest-v6.x/docs/api/crypto.html#crypto_crypto_timingsafeequal_a_b not ==

api/email/unsubscribe/[unsubscribeToken].ts
Comment (sshmm) [reply api/email/unsubscribe/[unsubscribeToken].ts:13]: @crabsinger I think so but typescript is expecting either string or string[] so I handled it the same way api/cosoul/[address] is handled

api-lib/email/unsubscribe.ts
sshmm: @crabsinger  I am currently using timingSafeEqual. I converted the string into a buffer first.
I have tested the functionality after.

sshmm (commented):
@crabsinger I think it's good to go

sshmm:
@wingfeatherwave I think we may need a different background image for UnsubscribeEmailPage. I am using the same image as VerifyEmailPage

sshmm:
@crabsinger  postmark needs to be updated first before merging

## Before Merging
The production templates on postmark needs to be updated as well
1- A new layout needs to be updated "CoLinks with one click unsubscribe"
this layout will have the ubsubscribe link at the footer
image

2- CoLinks Unread Notifications template needs to use this new layout instead.

sshmm (approved):

PR #2704, opened 2024-01-17
reomove object.assign from syncCosoul 

sshmm:
reomove object.assign from syncCosoul


crabsinger (approved):
@sshmm did you test this?

PR #2710, opened 2024-01-23
Allow CoLinks use on other chains

crabsinger:
Allow the app to be used with a readonly provider and disabling any
contract interactions (buy/sell/mint)


CryptoGraffe:
This looks reasonable but it seems like a better implementation would not disable the buttons but rather prompt chain switch

src/features/colinks/CoLinksContext.tsx
```diff
@@ -54,19 +54,6 @@ const CoLinksProvider: React.FC<CoLinksProviderProps> = ({ children }) => {
 
   const { data } = useCoLinksNavQuery();
 
-  useEffect(() => {
```
CryptoGraffe, line 57: I think w/o this the wizard will not act nicely when you are on a different chain, would need to test to see how it behaves

PR #2715, opened 2024-01-25
Walletless life

CryptoGraffe:
## What
TODO:

- [ ] remove the fake login stuff
- [ ] actually let wallet connecting happen when you hit the web3 barrier in `useDoWithCoLinksContract`
- [ ] make sure wizard is good
- [ ] load/save the address on refresh when logged in w/o wallet


src/hooks/useConnectedAddress.ts
```diff
@@ -1,11 +1,11 @@
 import assert from 'assert';
 
-import { useWeb3React } from 'hooks/useWeb3React';
```
crabsinger, line 3: this is breaking tests

crabsinger (approved):

PR #2747, opened 2024-02-22
fix subscription cancellation decoding function

sshmm:
##what
fix unsubscribe decode function for production
removed successful cancelling header on errors

## Why
unsubscribe token is getting extra parameters on production

## Test and Deployment Plan
tested old unsubscribe links to unsubscribe locally and on staging

## How
split token on "?"


sshmm:
replaced by #2749

PR #2759, opened 2024-02-27
profile skill tags

wingfeatherwave:
Screenshot 2024-02-27 at 2 01 53 PM

size: XL WIP
Screenshot 2024-02-27 at 1 44 35 PM


crabsinger:
@wingfeatherwave has some conflicts

src/features/points/GiveReceived.tsx
```diff
@@ -111,16 +120,66 @@ export const GiveReceived = ({
...
+            css={{
+              columnGap: '$sm',
+              rowGap: 0,
+              flexWrap: 'wrap',
+              m: size == 'xl' ? '$sm' : 0,
+              alignItems: 'flex-start',
+            }}
+          >
             {sortedGives.map(g => (
-              <Flex key={`give_${g.skill}`} css={{ gap: '$md' }}>
```
crabsinger, line 120: will this key always be unique? I think maybe g.id would be more unique?

src/features/points/GiveReceived.tsx
```diff
@@ -111,16 +120,66 @@ export const GiveReceived = ({
...
+              <Flex key={`give_${g.skill}`}>
```
crabsinger, line 173: https://legacy.reactjs.org/docs/lists-and-keys.html#keys

> The best way to pick a key is to use a string that uniquely identifies a list item among its siblings. Most often you would use IDs from your data as keys:

I don't think this flex needs a key since it has no siblings

wingfeatherwave: I was seeing bugs when it didn't have a key because of the map

src/features/points/GiveReceived.tsx
wingfeatherwave: added index

crabsinger (approved):

PR #2764, opened 2024-02-29
Show avatar on skillpicker

crabsinger:


crabsinger:
@wingfeatherwave its pretty weird if user doesn't have an avatar - pretty green initials...

wingfeatherwave (approved):

PR #2772, opened 2024-02-29
Colinks decent swap

sshmm:
## What
Add a bridge Component upon out of balance while buying links or minting Cosouls

### Getting Balance
1- Checking the network is removed in colinks wizard steps so that the user can switch network and swap 
2- getBalance from wagmi core is used to get user balance for the target network for example Optimism on production even if the user is using different network at the moment
3- it's used to determine whether we show the swap component or not
4- it's using wagmiConfig in DecentSwap/config.ts to get the network alchemy rpc
5- useUsersBalances from '@decent.xyz/box-hooks' in useBalance.ts to get the balance for the rest of token 'source tokens' 

### Selecting Pair
1- The destination is always OP Eth in production and OP Sepolia Eth in stagin unless feature test_decent is set to true
2- ChainSelectorModal is used to select the source chain
3- The source: TokenSelector from '@decent.xyz/box-ui' is used to select the token with setSrcToken. it provides srcToken with the token info

### Swaping
1- useSwapQuotes us used to get a quote for the user
2- confirmRoute function makes sure that the user is switched to the correct chain.
uses generateDecentAmountInParams and generateDecentAmountOutParams to prepare useBoxAction input 
3- useBoxAction  from '@decent.xyz/box-hooks' is used to get cost information and get transaction data 
4- after confirming transaction executeTransaction is used with the data from step 2 and 3 
5- approveToken is used to get the transaction data and then it gets sent to metamask

## Screenshots
image


sshmm:
@wingfeatherwave  I updated it with the agreed changes except for the fine print. we need to agree on it

wingfeatherwave:
Updating styling:
Screenshot 2024-03-06 at 12 57 23 PM
Screenshot 2024-03-06 at 12 48 47 PM
Screenshot 2024-03-06 at 12 39 13 PM
Screenshot 2024-03-06 at 12 39 20 PM

crabsinger:
@sshmm can you cleanup the PR descirption and add lots of info

package.json
```diff
@@ -59,6 +63,7 @@
     "@spruceid/siwe-parser": "^2.0.0",
     "@stitches/react": "^1.2.8",
     "@storybook/react": "^7.6.0",
+    "@tanstack/react-query": "^5.18.0",
```
crabsinger, line 66: is this required?

src/features/colinks/BuyOrSellCoLinks.tsx
```diff
@@ -53,23 +67,26 @@ export const BuyOrSellCoLinks = ({
...
   const doWithCoLinksContract = useDoWithCoLinksContract();
 
   const { data: opBalance } = useQuery(
     ['balanceOf', account],
     async () => {
       if (account) {
-        return await library?.getBalance(account);
+        return await getBalance(wagmiConfig, {
+          address: account as Address,
+          chainId: wagmiChain.id,
```
crabsinger, line 79: were using wagmi here?

src/features/colinks/wizard/WizardList.tsx
```diff
@@ -37,7 +30,6 @@ export const WizardList = ({ progress }: { progress: WizardProgress }) => {
       }}
     >
       <Step label="Connect Wallet" test={!!address} />
-      <Step label="On Optimism" test={onCorrectChain} />
```
crabsinger, line 40: How do we know if your ever on the right chain?

package.json
sshmm: I don't think it's needed anymore it was used for when I was using wagmi to connect.
I am going to remove it

src/features/colinks/BuyOrSellCoLinks.tsx
sshmm: to get the balance for OP eth regarding for the used network. I  was able to do it with react web3

src/features/colinks/wizard/WizardList.tsx
sshmm: when you try to mint or buy a link you will be asked to switch networks.
if you don't have enough balance it will till you that you only have xx balance of Optimism Eth and so

sshmm: although once you have balance you will be asked to switch to chain to min at CoSoulButton.tsx line 125
`  if (chain && !onCorrectChain) {
    return (
      <Button color="cta" size="large" onClick={safeSwitchToCorrectChain}>
        Switch to {chain.chainName} to Mint
      </Button>
    );
  }
`

CryptoGraffe (approved):

PR #2776, opened 2024-03-01
Show next give at in relative time

crabsinger:
Screenshot 2024-03-01 at 10 58 41 AM

### DIFF
--- a/src/features/points/usePoints.tsx
+++ b/src/features/points/usePoints.tsx
@@ -1,10 +1,15 @@
 import assert from 'assert';
 
+import { DateTime } from 'luxon';
 import { useQuery } from 'react-query';
 
 import { client } from '../../lib/gql/client';
 
-import { POINTS_PER_GIVE, getAvailablePoints } from './getAvailablePoints';
+import {
+  EMISSION_PER_SECOND,
+  POINTS_PER_GIVE,
+  getAvailablePoints,
+} from './getAvailablePoints';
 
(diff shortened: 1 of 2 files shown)


src/features/points/usePoints.tsx
```diff
@@ -52,3 +58,11 @@ const getMyAvailablePoints = async () => {
     profile.points_checkpointed_at
   );
 };
+
+const nextGiveAvailableAt = (currPoints: number) => {
```
CryptoGraffe, line 62: should this handle it being available now?

src/features/points/usePoints.tsx
```diff
@@ -52,3 +58,11 @@ const getMyAvailablePoints = async () => {
...
+  const needed = POINTS_PER_GIVE - (currPoints % POINTS_PER_GIVE);
```
CryptoGraffe, line 63: Modulus seems strange here

crabsinger: for current use, its not necessary since component only shows this if you have < 1 give. But this could be used elsewhere, like "when will I get my 4th Give available" ? In that case, we want to know the time until next give.

Lets say you have 3.2 GIVE available, time to next give should be 1 - .2, not 1 - 3.2. Thus modulus removes the excess, right? @CryptoGraffe

CryptoGraffe (approved):
could be better, but ok

PR #2784, opened 2024-03-06
UI Bug: "first to add" button position

wingfeatherwave:
Not sure why this "be the first to add" is appearing above the list, but we can pin it to the bottom with css.  

buggy prod:
Screenshot 2024-03-06 at 3 04 18 PM

absolute pin:
Screenshot 2024-03-06 at 3 00 40 PM

Hopefully this will work in prod

### DIFF
--- a/src/components/SkillComboBox/SkillComboBox.tsx
+++ b/src/components/SkillComboBox/SkillComboBox.tsx
@@ -258,6 +258,7 @@ const AddItem = ({
 
   return (
     <Command.Item
+      className="firstToAdd"
       color={'cta'}
       key={search}
       value={search}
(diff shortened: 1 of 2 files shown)


src/components/SkillComboBox/SkillComboBox.tsx
```diff
@@ -267,16 +268,7 @@ const AddItem = ({
         css={{ justifyContent: 'space-between', width: '100%', gap: '$lg' }}
       >
         <Text>Be the First to Add</Text>
```
crabsinger, line 270: Does moving this to bottom change ordering? Maybe Graffe can help investigate why ordering isnt working as expected?

wingfeatherwave: it doesn't change ordering as far as I can tell.  i'm not able to reproduce the bug locally

CryptoGraffe (approved):

PR #2811, opened 2024-03-28
Batch pGIVE Syncing

sshmm:
## What
Switch cosoul pgive syncer to use batchSetSlot_UfO  instead of  setSlot:
- batch syncing in cron
- Updated existing unit testing.
- added a new unit test for syncCoSouls cron to make sure all changes work together

## Test and Deployment Plan
The changes are tested in the following files:
hardhat/test/01-cosoul/00-cosoul.ts
src/features/cosoul/api/cosoul.test.ts
api-test/hasura/cron/syncCoSouls.test.ts - new file

## How
1 - batchSetSlot_UfO needs bytes array: first byte is pgive slot which is 0
then 8 bytes for each pgive where 4 bytes for pgives amount and four for token id.
I utilized paddedHex and converted to typescript to freely convert the number to hex and add 0x on choice

### DIFF
--- a/_api/hasura/cron/syncCoSouls.ts
+++ b/_api/hasura/cron/syncCoSouls.ts
@@ -9,7 +9,9 @@ import { getCirclesNoPgiveWithDateFilter } from '../../../api-lib/pgives';
 import { Awaited } from '../../../api-lib/ts4.5shim';
 import { verifyHasuraRequestMiddleware } from '../../../api-lib/validate';
 import {
+  CoSoulArgs,
   getOnChainPGive,
+  setBatchOnChainPGive,
   setOnChainPGive,
 } from '../../../src/features/cosoul/api/cosoul';
 import { getLocalPGIVE } from '../../../src/features/cosoul/api/pgive';
@@ -17,7 +19,7 @@ import { storeCoSoulImage } from '../../../src/features/cosoul/art/screenshot';
 
 Settings.defaultZone = 'utc';
 
-const LIMIT_USERS_TO_SYNC = 10;
+const LIMIT_USERS_TO_SYNC = 11;
 
 type CoSoul = Awaited<ReturnType<typeof getCoSoulsToUpdate>>[number];
 
@@ -30,8 +32,15 @@ async function handler(req: VercelRequest, res: VercelResponse) {
     });
   }
 
-  const success = await syncCoSouls();
-  res.status(200).json(success);
+  let success;
+  try {
+    success = await syncCoSouls();
+  } catch (e: any) {
+    errorLog(`Error in syncCoSouls cron: ${e}`);
+    return res.status(500).json({ success: false, error: e.message });
+  }
+
+  return res.status(200).json(success);
 }
 
 export async function syncCoSouls() {
@@ -53,13 +62,12 @@ export async function syncCoSouls() {
   }
 
   // update each one on-chain if needed, otherwise just update the checked_at column
-  const updated = [];
+  const pending = [];
   const errors = [];
   const ignored = [];
   for (const cosoul of cosouls) {
     const localPGIVE = await getLocalPGIVE(cosoul.address);
     const onChainPGIVE = await getOnChainPGive(cosoul.token_id);
-    let success = true;
     if (localPGIVE !== onChainPGIVE) {
       // update the screenshot
       // this might take a while and might need to be handled in a separate process
@@ -69,13 +77,7 @@ export async function syncCoSouls() {
         console.error('failed to screenshot CoSoul ' + cosoul.token_id, e);
         // proceed with setting on-chain pgive
       }
-      success = await updateCoSoulOnChain(cosoul, localPGIVE);
-
-      if (success) {
-        updated.push(cosoul.id);
-      } else {
-        errors.push(cosoul.id);
-      }
+      pending.push({ cosoul, localPGIVE });
     } else {
       ignored.push(cosoul.id);
       console.log(
@@ -92,8 +94,17 @@ export async function syncCoSouls() {
       await updateCheckedAt(cosoul.id);
     }
   }
-  const message = `${cosouls.length} CoSouls updated`;
-  const status = { message, updated, errors, ignored };
+  let success = false;
+  if (pending.length > 0) {
+    success = await updateCoSoulOnChain(pending);
+
+    if (!success) {
+      errors.push(...pending.map(({ cosoul }) => cosoul.id));
+    }
+  }
+
+  const message = `${pending.length} CoSouls updated`;
+  const status = { message, pending, errors, ignored };
   console.log(status);
   return status;
 }
@@ -196,8 +207,8 @@ const syncCoSoulToken = async (
           },
           _set: {
             pgive: totalPGIVE,
-            checked_at: new Date().toISOString(),
-            synced_at: new Date().toISOString(),
+            checked_at: 'now()',
+            synced_at: 'now()',
           },
         },
         {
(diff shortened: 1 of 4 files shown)


_api/hasura/cron/syncCoSouls.ts
```diff
@@ -148,76 +153,96 @@ const getCoSoulsToUpdate = async () => {
...
-        totalPGIVE
-    );
-  } else {
-    console.log(
-      'skipping setting on-chain PGIVE because it is 0 for tokenId: ' +
-        tokenId +
-        ' address: ' +
-        address
-    );
+  let payload = paddedHex(PGIVE_SLOT, 2, true); //1byte for slot
```
crabsinger, line 158: can we make all paddedHex stuff hidden away inside a helper function in the cosoul api? @sshmm

src/features/cosoul/api/cosoul.ts
```diff
@@ -51,17 +72,10 @@ export const getOnChainPGIVE = async (tokenId: number) => {
 };
 
 // set the on-chain PGIVE balance for a given token
-export const setOnChainPGIVE = async (tokenId: number, amt: number) => {
```
crabsinger, line 54: make two functions, one for multi setting, one for single setting

sshmm: @crabsinger  done. we have now two functions and the tests are updated accordingly

_api/hasura/cron/syncCoSouls.ts
sshmm: @crabsinger done getPayload will take tokenId and pgives and returns 16 characters hex for them

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -206,24 +219,116 @@ const syncCoSoulToken = async (
   );
 };
 
-async function updateCoSoulOnChain(cosoul: CoSoul, totalPGIVE: number) {
+const syncPatchCoSoulToken = async (
```
crabsinger, line 222: patch? you mean batch? @sshmm

hardhat/test/utils/payload.ts
```diff
@@ -0,0 +1,23 @@
+export const paddedHex = (
```
crabsinger, line 1: what else should go in this file? seems this should just be in another file

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -216,24 +229,118 @@ const syncCoSoulToken = async (
...
+          cosoul.address
+      );
+    }
+  }
+  if (payload.length > 4) {
+    // 0x and one byte for slot 0x00
+    //if only one needs to be updated on chain, we can use setOnChainPGIVE
+    if (payload.length === 20) {
+      //0x00 plus 4 bytes for pgive and 4 bytes for tokenId
+      const tx = await setOnChainPGIVE(
```
crabsinger, line 259: do we need to use this individual syncer? i think the code would be cleaner if we just always use the batch rather than all these ifs... but we can change later.

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -216,24 +229,118 @@ const syncCoSoulToken = async (
...
+      console.log(
+        'skipping setting on-chain PGIVE because it is 0 for tokenId: ' +
+          cosoul.token_id +
+          ' address: ' +
+          cosoul.address
+      );
+    }
+  }
+  if (payload.length > 4) {
+    // 0x and one byte for slot 0x00
```
crabsinger, line 255: is this comment referring to the line above or below?

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -216,24 +229,118 @@ const syncCoSoulToken = async (
...
+    //if only one needs to be updated on chain, we can use setOnChainPGIVE
+    if (payload.length === 20) {
+      //0x00 plus 4 bytes for pgive and 4 bytes for tokenId
```
crabsinger, line 258: 0x00 + 4 bytes + 4 bytes = 20? can you make the comments more clear about what code they are referring to @sshmm

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -216,24 +229,118 @@ const syncCoSoulToken = async (
...
+        await adminClient.mutate(
+          {
+            update_cosouls_by_pk: [
+              {
+                pk_columns: {
+                  id: cosoul.id,
+                },
+                _set: {
+                  pgive: localPGIVE,
+                  checked_at: new Date().toISOString(),
```
crabsinger, line 283: for the future, these can just be `now()` like https://github.com/coordinape/coordinape/blob/main/api-lib/poap/poap-api.ts#L208

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -216,24 +229,118 @@ const syncCoSoulToken = async (
...
+    }
     return true;
   } catch (e: any) {
     // don't ruin the whole thing, this might be an on-chain issue or temporary setback
     // TODO: send this to sentry
+    const failedBatch = updatedCosouls.reduce(
+      (accumulator, currentValue, currentIndex) =>
+        accumulator +
+        `[Cosoul ${currentIndex} with id: ${currentValue.cosoul.id}, tokenId: ${currentValue.cosoul.token_id}, address: 
+      ${currentValue.cosoul.address} ,and targetPIVE: ${currentValue.localPGIVE}]`,
```
crabsinger, line 338: the ` ${currentValue.cosoul.address} ,and` seems weird, you mean `${currentValue.cosoul.address}, and` ?

hardhat/test/utils/payload.ts
sshmm: @crabsinger  removed the file

_api/hasura/cron/syncCoSouls.ts
sshmm: if it's only one token to be updated I think we don't need the extra operations in batch pgive like getting the length and dividing it by 8 to decide the number of tokens to update and looping on tokens.
@crabsinger  @CryptoGraffe  if these extra operations won't cost anything I can make all of them use batch syncer

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -85,7 +88,14 @@ export async function syncCoSouls() {
       await updateCheckedAt(cosoul.id);
     }
   }
-  const message = `${cosouls.length} CoSouls updated`;
+  let success = true;
+  if (updated.length > 0) {
+    success = await updateCoSoulOnChain(updated);
```
crabsinger, line 93: can we rename these variables.

using `updated` as the list of cosouls that are gathered prior to be updated is confusing. it seems this should really be named `pending` or `to_be_updated` or `needSync`.

additionally, it seems unconventional to set `success = true` prior to doing the update. I would set success = false, do update, if successful then set to true. or some other pattern that is clearer and less error prone. @sshmm

_api/hasura/cron/syncCoSouls.ts
```diff
@@ -207,24 +217,118 @@ const syncCoSoulToken = async (
   );
 };
 
-async function updateCoSoulOnChain(cosoul: CoSoul, totalPGIVE: number) {
+const syncBatchCoSoulToken = async (
+  updatedCosouls: { cosoul: CoSoul; localPGIVE: number }[]
```
crabsinger, line 221: `updatedCosouls` are the ones we want to update, not the ones that have been updated right? Maybe make variable name no longer the past tense. @sshmm

src/features/cosoul/api/cosoul.ts
```diff
@@ -65,6 +89,13 @@ export const setOnChainPGIVE = async (tokenId: number, amt: number) => {
   return await contract.setSlot(PGIVE_SLOT, amount, tokenId, gasSettings);
 };
```
crabsinger, line 91: create simple method `export const setBatchOnChainPGIVE = async ([tokenId: number, amt: number]) => {`

 that takes same type as setOnChainPGive but array of it, and all details are hidden inside.

PR #2859, opened 2024-04-16
start of give party impl

CryptoGraffe:


_api/frames/router.tsx
```diff
@@ -289,8 +293,15 @@ const addFrame = (frame: Frame) => {
     async (_req, res, params) => {
       const ir = new ImageResponse(await frame.imageNode(params), {
         // debug: true,
-        height: 1000,
-        width: 1000,
+        ...(frame.aspectRatio === '1.91:1'
+          ? {
```
crabsinger, line 297: @CryptoGraffe what's this? why do we need to support two aspect ratios?

CryptoGraffe:
Proposal for give party from zemm and ash was to use the shorter ratio

crabsinger (approved):

PR #2860, opened 2024-04-16
add notifications for others replies

sshmm:
## What
when a user replies to a post you have replied to before you will receive a notification

## Test and Deployment Plan
Create a post by a user and comment on it with two other users and then comment by the post creator, each user will receive a different notification as per by the screenshots

image
image
image

## How
by the reply id I fetched all the users that has replied to this post before.
excluded users that are mentioned to that reply because they are getting different notifications

### DIFF
--- a/hasura/migrations/default/1713188205470_alter_table_public_notifications_add_column_second_actor_profile_id/up.sql
+++ b/hasura/migrations/default/1713188205470_alter_table_public_notifications_add_column_second_actor_profile_id/up.sql
@@ -1,0 +1,2 @@
+alter table "public"."notifications" add column "second_actor_profile_id" int8
+ null;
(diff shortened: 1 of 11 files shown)


hasura/migrations/default/1713188205470_alter_table_public_notifications_add_column_second_actor_profile_id/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."notifications" add column "second_actor_profile_id" int8
```
crabsinger, line 1: seems a bit weird as a column name. i think we should re-rethink the DB schema and see how best to design this. @sshmm

sshmm:
replaced by #2925

sshmm:
replaced by #2925

PR #2865, opened 2024-04-18
make invite code optional

sshmm:
## What
- reverted removing redeemInvitePage from the flow
- made code redemption optional

## Why
instead of showing the friendly /start page, its going right to /wizard which forces log in

## Test and Deployment Plan
Join using an invite code or without it
sign out and you should reach this page
image

image

joining without invite code
image

joining with invite code
image

## How
- instead of setting invite_code_redeemed_at to current time by default, the user will set it to current time upon joining
- invited_by is used to decide the success message


src/features/invites/RedeemInviteCode.tsx
```diff
@@ -113,6 +113,36 @@ export const RedeemInviteCode = ({
...
 
+  const updateRedeemedAt = async () => {
+    try {
+      setLoading(true);
+      const { update_profiles_by_pk } = await client.mutate(
+        {
+          update_profiles_by_pk: [
+            {
+              pk_columns: { id: profileId },
+              _set: { invite_code_redeemed_at: 'now()' },
```
CryptoGraffe, line 124: this is ok for now but this implies anyone can set this, probably not the best pattern

CryptoGraffe (approved):

PR #2914, opened 2024-04-24
WC direct links

crabsinger:


api-lib/gql/__generated__/zeus/const.ts
```diff
@@ -15743,6 +15749,7 @@ export const ReturnTypes: Record<string, any> = {
     skill: 'citext',
     target_profile_id: 'bigint',
     updated_at: 'timestamptz',
+    warpcast_url: 'String',
```
CryptoGraffe, line 15752: should this not just be URL to make it more general purpose

hasura/migrations/default/1714062664278_alter_table_public_colinks_gives_add_column_warpcast_url/down.sql
```diff
@@ -0,0 +1 @@
+alter table "public"."colinks_gives" drop column "warpcast_url";
```
CryptoGraffe, line 1: warpcast_url->source_url or related_url or something?

PR #2923, opened 2024-04-30
Colinks handle inactive emails

sshmm:
## What
set verified_at to null if the user is rejecting colinks emails

## Test and Deployment Plan
I tried it by throwing an error with 'spam complaint' and the notified email got unverified


crabsinger (approved):
looks good, needs to be ushered into production and verify it works in prod

PR #2925, opened 2024-05-01
add notifications for others replies

sshmm:
## What
#replaces #2860 
when a user replies to a post you have replied to before you will receive a notification

## Test and Deployment Plan

Create a post by a user and comment on it with two other users and then comment by the post creator, each user will receive a different notification as per by the screenshots

Commenter 1
image

Thread owner
image

Commenter 2

## How
1- When a user reply to a post, get all other replies to that post and their posters.
2- notify other replies posters.

### DIFF
--- a/api-lib/event_triggers/createNotificationReplies.ts
+++ b/api-lib/event_triggers/createNotificationReplies.ts
@@ -1,6 +1,7 @@
 import type { VercelRequest, VercelResponse } from '@vercel/node';
 import { uniq } from 'lodash-es';
 
+import { replies_select_column } from '../gql/__generated__/zeus';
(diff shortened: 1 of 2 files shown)


crabsinger:
@sshmm we spoke about adding indices to ensure that the nested queries are using indices for the fetches, did you verify all these queries already exist, or do you still need to add them?

crabsinger:
only 2 file changes, seems to small.

sshmm:
> only 2 file changes, seems to small.

@crabsinger  yes because this approach doesn't require schema changes

crabsinger (approved):

PR #2944, opened 2024-05-29
Put Give Onchain

crabsinger:


_api/hasura/cron/giveOnchainSyncer.ts
```diff
@@ -0,0 +1,144 @@
+/* eslint-disable no-console */
```
CryptoGraffe, line 1: hmmm

api-lib/gql/__generated__/zeus/const.ts
```diff
@@ -2204,15 +2204,18 @@ export const AllTypesProps: Record<string, any> = {
     _not: 'colinks_gives_bool_exp',
     _or: 'colinks_gives_bool_exp',
     activity_id: 'bigint_comparison_exp',
+    attestation_uid: 'String_comparison_exp',
     cast_hash: 'String_comparison_exp',
     created_at: 'timestamptz_comparison_exp',
     give_skill: 'skills_bool_exp',
     giver_profile_public: 'profiles_public_bool_exp',
     id: 'Int_comparison_exp',
+    onchain_sync_error: 'String_comparison_exp',
```
CryptoGraffe, line 2213: maybe we want on onchain_synced_at timestamp

_api/hasura/cron/giveOnchainSyncer.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+async function giveToSync() {
+  // get all give that is not on-chain
+  const { colinks_gives } = await adminClient.query(
+    {
+      colinks_gives: [
+        {
+          where: {
+            tx_hash: { _is_null: true },
```
CryptoGraffe, line 117: what about ones with errors? could we get stuck forever?

_api/hasura/cron/giveOnchainSyncer.ts
```diff
@@ -0,0 +1,144 @@
...
+          {
+            operationName: 'cron_giveOnchainSyncer__attestGive',
+          }
+        );
+
+        success++;
+      } catch (e: any) {
+        errors++;
+
+        await adminClient.mutate(
```
CryptoGraffe, line 65: should this just happen in the attest function so the updates are all in one place

api-lib/eas.ts
```diff
@@ -0,0 +1,78 @@
...
+  '0x44ba9714fe382b86f0e126a592116bc2fc45905ee034d0e8b5db218bf366dafe';
+const BASE_SEPOLIA_SCHEMA_UID =
+  '0x55df90070189848b41fc4f2a8e07f82d6434718696286b9188ceb1d737f6cf87';
+
+export async function attestGiveOnchain(give: Give) {
+  const chainId = Number(baseChain.chainId);
+  const provider = getProvider(chainId);
+
+  const syncerWallet = new Wallet(COSOUL_SIGNER_ADDR_PK);
+  const signer = syncerWallet.connect(provider);
```
CryptoGraffe, line 25: do we want a diff addr for this or is this all good

api-lib/eas.ts
```diff
@@ -0,0 +1,78 @@
...
+  );
+  const encodedData = schemaEncoder.encodeData([
+    {
+      name: 'from',
+      value: giverAddr,
+      type: 'address',
+    },
+    { name: 'skill', value: give.skill, type: 'string' },
+    { name: 'amount', value: 1, type: 'uint16' },
+    { name: 'context', value: context, type: 'string' },
```
CryptoGraffe, line 53: context still feels ambiguous

api-lib/gql/__generated__/zeus/const.ts
crabsinger: ok, will add and use as primary search for things to sync

_api/hasura/cron/giveOnchainSyncer.ts
crabsinger: yeah... we could, any good ways around this?

api-lib/eas.ts
crabsinger: i think just this one and then we have onchain.coordinape.eth set to it

hasura/migrations/default/1717008407293_alter_table_public_colinks_gives_add_column_tx_hash/up.sql
```diff
@@ -0,0 +1,2 @@
+alter table "public"."colinks_gives" add column "tx_hash" text
```
sshmm, line 1: is not it better to add all changes of the same table of the same feature to the same migration file?

_api/hasura/cron/giveOnchainSyncer.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    // eslint-disable-next-line no-console
+    console.log('Syncing complete', { success, errors });
+    res.status(200).json({
+      success: true,
+      gives: gives.length,
+      errorsCount: errors,
+      successCount: success,
+    });
+  } catch (e: any) {
```
sshmm, line 33: will this "try ..catch" catch any errors? don't we need to throw again if errors is greater than 0

crabsinger: good catch - moved the give fetching into the block

hasura/migrations/default/1717008407293_alter_table_public_colinks_gives_add_column_tx_hash/up.sql
crabsinger: it is better - but im lazy now

PR #2953, opened 2024-06-11
add script to seed colinks data

sshmm:
## What
bootstrap local colinks users for testing. 
Create 5 users and have them all buy each others links and create posts and gives between them. 

### creating users
src/utils/provider is used to get the users with balance
createProfiles is used to generate profiles for these users

### minting cosouls
insertCosouls is used to mint cosouls 
contract.mint() and api mintCoSoulForAddress are used to mint cosouls only if getTokenId doesn't return a cosoul id

### buying keys
buyLinks is used for buying links
CoLinks.connect and Colink.buyLinks are used for buying links 
linkBalance is used to skip buying if the user already has a link.

### creating posts and replies
addPosts is used for that
first contibutions are created then activities manually then replies


scripts/seed_colinks.sh
```diff
@@ -0,0 +1,32 @@
...
+echo "Seeding..."
+tsx ./scripts/db_seed_colinks.ts
+
+# Re-enable event triggers post-seeding
+# Don't apply metadata, because disable_cron will do it
+echo "Re-enabling triggers and cron"
+APPLY_METADATA=no ./scripts/enable_triggers.sh
+
+# disable cron tasks to minimize unexpected CPU spikes
+./scripts/disable_cron.sh
```
crabsinger, line 30: we leave these disabled?

sshmm: the same script 'disable_cron' re-enable it after
`toggle_cron`

`pnpm hasura metadata apply`

`toggle_cron`

PR #2984, opened 2024-06-24
update rep score to new model, add farcaster rep

CryptoGraffe:


hasura/metadata/cron_triggers.yaml
```diff
@@ -1,21 +1,21 @@
...
-- name: backfillFarcasterConnect
-  webhook: '{{HASURA_API_BASE_URL}}/cron/backfillFarcasterConnect'
+  comment: auto connect farcaster accounts for those with matching address
+- name: backfillReputation
+  webhook: '{{HASURA_API_BASE_URL}}/cron/backfillReputation'
   schedule: '* * * * *'
   include_in_metadata: true
   payload: {}
   headers:
     - name: verification_key
```
crabsinger, line 16: this is jsut diff noise i think. maybe worth committing if we think its going to stay stable... but if not then best to leave out.

hasura/migrations/default/1719206064672_alter_table_public_reputation_scores_add_column_farcaster_score/down.sql
```diff
@@ -0,0 +1,4 @@
+-- Could not auto-generate a down migration.
+-- Please write an appropriate down migration for the SQL below:
+-- alter table "public"."reputation_scores" add column "farcaster_score" integer
```
crabsinger, line 3: down migrations

crabsinger (changes requested):
a few easy fixes

PR #2999, opened 2024-07-03
manage big questions

sshmm:
## How
1- access big question edit page from gear button
image
it's visible and accessible only to selected profiles
2- you can add new questions or edit current/upcomming questions
image

@wingfeatherwave  could you review the image styling?

## Testing
to locally test you need to add your profile id to this list BIG_QUESTION_MANAGERS at src/common-lib/constants.ts 
you need to test also the absence of access when your profile id is not in this list


crabsinger:
@sshmm lets use a hasura permission to allow inserts or edits to the big questions for the following profile ids:

```
| profile_id |  name |
4031,Graffe
194851,Wing
1346,crabsinger.eth
249055,ashmbanks.eth
31,zemm.eth
```

_api/hasura/actions/_handlers/updateBigQuestion.ts
```diff
@@ -0,0 +1,79 @@
...
+    console.log(true);
+  }
+  if (big_question_id) {
+    await adminClient.mutate(
+      {
+        update_big_questions_by_pk: [
+          {
+            pk_columns: { id: big_question_id },
+            _set: {
+              cover_image_url: cover_image_url,
```
crabsinger, line 44: would be ideal if we had a way to upload an image inthe UI, and it would uplaod to cloudflare

like this: generateOneTimeUpload.ts @sshmm

PR #3031, opened 2024-08-07
Refactor give delivered cast delivery to happen for all gives to casts

crabsinger:


api-lib/insertCoLinksGive.ts
```diff
@@ -89,3 +94,25 @@ export const insertCoLinksGive = async (
...
+export const publishCastGiveDelivered = async (
+  hash: string,
+  giveId: number,
+  precache = true
+) => {
+  if (precache) {
+    // PRE-CACHE farme and image by calling the URL
+    await fetch(getFrameUrl('give', giveId), {
+      signal: AbortSignal.timeout(10000),
+    });
```
crabsinger, line 107: will make handler resposne slwoer, but no other easy way to precache the frame image.

PR #3053, opened 2024-09-03
Cleanup and Refactor Alchemy tokens

crabsinger:
Alchemy (finally) supports multiple chains with one API key, so we can vastly simplify our env vars which use Alchemy tokens.

We have two API tokens:

VITE_FE_ALCHEMY_API_KEY - this is visible to clients and users and has domain origin limitations set on its use. Only use this from the FE where requests will have an origin.

BE_ALCHEMY_API_KEY - this is a private API_KEY and should not be exposed to clients, it is used for backend requests without origin.

Change all contract instance code to use appropriate versions of frontend or backend clients.

### DIFF
--- a/_api/hasura/cron/syncCoSouls.ts
+++ b/_api/hasura/cron/syncCoSouls.ts
@@ -2,7 +2,7 @@
 import { VercelRequest, VercelResponse } from '@vercel/node';
 import { DateTime, Settings } from 'luxon';
 
-import { IS_LOCAL_ENV } from '../../../api-lib/config';
+import { BE_ALCHEMY_API_KEY, IS_LOCAL_ENV } from '../../../api-lib/config';
 import { adminClient } from '../../../api-lib/gql/adminClient';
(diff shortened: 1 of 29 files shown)


_api/hasura/cron/syncCoSouls.ts
```diff
@@ -171,7 +174,8 @@ const syncCoSoulToken = async (
     totalPGIVE = Math.floor(totalPGIVE);
     const txHash = await setOnChainPGive({ tokenId, amount: totalPGIVE });
 
-    const publicClient = getReadOnlyClient();
+    const chainId = Number(chain.chainId);
+    const publicClient = getReadOnlyClient(chainId, BE_ALCHEMY_API_KEY);
```
CryptoGraffe, line 178: Maybe this should be wrapped in something in api-lib like getBackendReadOnlyClient

CryptoGraffe (approved):

PR #3066, opened 2024-09-10
use wagmi for decent transactions

sshmm:
used  sendTransaction and  waitForTransactionReceipt from wagmi core
@crabsinger  I have update publicClient.ts with the rest of chains but I am not using it right now

### DIFF
--- a/src/features/cosoul/CoSoulButton.tsx
+++ b/src/features/cosoul/CoSoulButton.tsx
@@ -1,5 +1,3 @@
-// FIXME: reeanble DecentSwap
-/* eslint-disable @typescript-eslint/no-unused-vars */
 import { ChainId } from '@decent.xyz/box-common';
 import { getBalance } from '@wagmi/core';
 import { ethers } from 'ethers';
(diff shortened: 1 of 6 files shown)


src/features/cosoul/CoSoulButton.tsx
```diff
@@ -80,17 +78,18 @@ export const CoSoulButton = ({ onReveal }: { onReveal(): void }) => {
             another L2 (eg. Base, Arbitrum).
           </Text>
           <Flex column css={{ mt: '$sm' }}>
-            {/* {defaultAvailableChains.includes(chainId as ChainId) &&             (!IN_PREVIEW || (IN_PREVIEW && isFeatureEnabled('test_decent'))) ? (
+            {defaultAvailableChains.includes(chainId as ChainId) &&
+            (!IN_PREVIEW || (IN_PREVIEW && isFeatureEnabled('test_decent'))) ? (
```
crabsinger, line 82: Since Decent is live, can we just remove this and all uses of `test_decent` and just show Decent everywhere? @sshmm

crabsinger (approved):
@sshmm one comment, looks good otherwise.

src/features/cosoul/CoSoulButton.tsx
sshmm: @crabsinger this feature flag is there to allow testing the convertion to optimism instead of optimism sepolia on staging and to also to prevent using actual tokens by mistake while testing

crabsinger: ah, makes sense. great.

PR #3102, opened 2024-10-01
logout when useNotificationCount or getNavData gets Auth Hook errors

CryptoGraffe:

### DIFF
--- a/src/features/nav/getNavData.ts
+++ b/src/features/nav/getNavData.ts
@@ -1,7 +1,10 @@
 import { order_by } from 'lib/gql/__generated__/zeus';
 import { client } from 'lib/gql/client';
 import { useQuery } from 'react-query';
+import { useNavigate } from 'react-router';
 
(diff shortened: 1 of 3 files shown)


src/features/nav/getNavData.ts
```diff
@@ -94,6 +99,14 @@ export const useNavQuery = () => {
       return { ...data, profile };
     },
     {
+      onSettled: async (_, error: any) => {
+        if (isErrorWarrantingLogout(error)) {
+          // eslint-disable-next-line no-console
+          console.log('logging out due to auth error in useNavQuery');
+          await logout();
```
crabsinger, line 106: the logout action will fail since we dont have an valid auth token... does it fail silently and still do the other browser side logout stuff? seems maybe we want to use a different hook that won't make a client.mutate which will fail? @CryptoGraffe

crabsinger (dismissed):

crabsinger (approved):

PR #3115, opened 2024-10-09
when creating a new user from farcaster, upload their url to cloudflare

CryptoGraffe:

### DIFF
--- a/api-lib/neynar/findOrCreate.ts
+++ b/api-lib/neynar/findOrCreate.ts
@@ -1,6 +1,7 @@
 /* eslint-disable no-console */
 import assert from 'assert';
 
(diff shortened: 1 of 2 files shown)


api-lib/neynar/findOrCreate.ts
```diff
@@ -154,6 +155,11 @@ const createProfile = async (
     console.log('Creating new profile with name', name);
   }
 
+  // if we have an avatar_url, upload it to cloudflare for our use
+  const avatar = avatar_url
+    ? await uploadURLToCloudflare(avatar_url)
```
crabsinger, line 160: this is gonna error in test/dev if you dont have a cloduflare token or valid keys?

CryptoGraffe: fixed that thanks

PR #3159, opened 2024-11-19
Create api to get top/trending give skills and component

crabsinger:
## 1. Create a query that figures out the top skills. The top skills are the skills that have the most gives.

Create an API to get the top skills.

## 2. Create a component with the top skills over last 7 days

image

- [ ] 3. Create a component that shows a list of "Gives with Casts" for each of these skills.

- All gives should be of same 'skill'.
- Question: All gives should have casts? or just all gives? (current is all gives)

### DIFF
--- a/_api/give/trending.ts
+++ b/_api/give/trending.ts
@@ -6,7 +6,7 @@ import { order_by } from '../../api-lib/gql/__generated__/zeus/index';
 import { adminClient } from '../../api-lib/gql/adminClient';
 import { errorResponse } from '../../api-lib/HttpError';
 
-const LIMIT = 10;
+const LIMIT = 5;
 
 // TODO: refactor to one global cache config?
 // 1 hour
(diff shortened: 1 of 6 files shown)


crabsinger:
getting weird error in typecheck for this file ,but not other _api files...

image

_api/give/trending.ts
```diff
@@ -0,0 +1,57 @@
...
+import { order_by } from '../../api-lib/gql/__generated__/zeus/index';
+import { adminClient } from '../../api-lib/gql/adminClient';
+import { errorResponse } from '../../api-lib/HttpError';
+
+const LIMIT = 5;
+
+// TODO: refactor to one global cache config?
+// 1 hour
+const maxAge = 60 * 60;
+export const CACHE_CONTENT = `public, s-maxage=${maxAge}, max-age=${maxAge}, stale-while-revalidate=${maxAge * 2}`;
```
CryptoGraffe, line 14: yeah there is already one just use that

CryptoGraffe (approved):

PR #3165, opened 2024-12-03
Fix trending give skills query

crabsinger:
Fetch total give skills not give skills per user.


CryptoGraffe (approved):
Nice easy one good catch

PR #3238, opened 2025-02-28
refactor give weight to a function, add give weight to EAS

CryptoGraffe:

### DIFF
--- a/api-lib/give.ts
+++ b/api-lib/give.ts
@@ -1,5 +1,13 @@
-export const UNLIMITED_GIVE_PROFILES = [4064783, 4064757, 4064836, 4065247];
-
+// FIXME: we need to make this dynamic ideally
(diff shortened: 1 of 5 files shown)


api-lib/give.ts
```diff
@@ -0,0 +1,17 @@
...
+  total_score: number | undefined;
+}) => {
+  const isBot = isBotOrAgent(profileId);
+  const botWeight = 69;
+  const give_weight = isBot ? botWeight : total_score || 1;
+  return give_weight;
+};
+
+export const isBotOrAgent = (profileId: number) =>
+  UNLIMITED_GIVE_PROFILES.includes(profileId);
```
crabsinger, line 17: this isnt currently up to date, and wont work in the future... should add a FIXME here. @CryptoGraffe

PR #3239, opened 2025-02-28
export anongql

CryptoGraffe:
## What

Exports the anongql client from the coordinape repo so other projects can use it. Unfortunately this carries all along the coordinape dependencies with it. We could potentially break this out at some point. 

Here is how to use it (example in the fv2 project):

```
# note this is currently branch tracked to the export_anongql branch
pnpm add github:coordinape/coordinape#export_anongql
```

then in code:
```
import { anonClient } from 'coordinape-app/anongql';
```

works!

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,9 @@
   "engines": {
     "node": "20"
   },
+  "exports": {
+    "./anongql": "./src/lib/anongql/index.ts"
+  },
   "dependencies": {
     "@apollo/client": "3.7.15",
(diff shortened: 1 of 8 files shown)


crabsinger (approved):

crabsinger:
@CryptoGraffe 
> Unfortunately this carries all along the coordinape dependencies with it. 

like ALL of them, all of them? Seems like not what we want for frames...

